PR #35, opened 2024-10-02
Bug/settings and exclusions

js3110:
Settings updated so that exclusions and inclusions are correctly implemented in slopes, nca results and exclusions table after upload. Code also updated so that no issues with upload of Dummy data settings.

**Checks**

- [x] Run app for all 4 dummy data sets and ensure settings can be saved and uploaded easily, and that they are implemented in NCA results, slope plots and exclusions table.
- [x] Test app for settings change by manually changing variables in the settings file so that when uploaded, the profiles with modified exclusions/inclusions are reset.
- [ ] Check all other functionalities also running fine

### DIFF
--- a/R/exportCDISC.R
+++ b/R/exportCDISC.R
@@ -97,7 +97,6 @@ exportCDISC <- function(resNCA){
                 "AVALC",
                 "AVALU")
 
-# browser()
 
 pp_info = resNCA$result  %>%
   filter(is.infinite(end) | PPTESTCD=='auclast' )  %>%
--- a/inst/shiny/tabs/data.R
+++ b/inst/shiny/tabs/data.R
@@ -60,8 +60,7 @@ observeEvent(input$local_upload, {
               mutate(TIME = ifelse(DOSNO == 1, AFRLT, ARRLT),
                      NDOSEDUR = as.numeric(NDOSEDUR),
                      ADOSEDUR = as.numeric(ADOSEDUR))
-    
-    # browser()
+
     ADNCA(ADNCA)
 })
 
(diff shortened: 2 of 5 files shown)


Gero1999:
Hello @js3110. I just noticed that saved changes in the table do not translate now in plot changes! Is this branch only involving in the settings input?

Gero1999:
### ToDo
- [ ] 1) Load a dataset including flag settings and AUC intervals (without exclusions/selections) > 2) Change things in the UI > 3) Run NCA > 4) Error: 
```
Warning: Error in filter: ℹ In argument: `&...`.
ℹ In group 1: `TYPE = NA`, `PATIENT = NA`, `PROFILE = NA`.
Caused by error in `NA:NA`:
! NA/NaN argument
```
### Small changes 
- [ ] Method "Linear LinearLogInterpolation" is not part of PKNCA. I don't know when was added, but take it off to elude mistakes in the settings as well.
- [ ] Delete the commented browser() call
- [ ] I have seen that you substituted a previous observeEvent() with a function (handle_nca_excsel). I personally won't recommend to do that, as this function is modifying the App behaviour. In my opinion, is just 3 lines so it is better to keep it simple and explicit in the code. Otherwise, if you are keen to keep it a function just put it as a separate file R/functions. But don't hesitate to let me know what are your thoughts on this.

### Others
- The problem with the table of exclusions and selections not saving is the true issue, I know that you told me about it. Dw I will try to take care of it and include in the review that also the settings are considered

js3110:
@Gero1999 i made most of the changes please check again! :)

PR #37, opened 2024-10-02
Bug: Slope selection save

js3110:
Slope selection did not save correctly when edits were made

**Checks**

- [x] Slope selections added and correctly saved in output table, nca results and slopes table. (with reasons also saved)


Gero1999:
Good catch! One little point easy to solve and I can merge:

### ToDo
- [ ] Delete lines used for debugging (i.e, str() )

js3110:
@Gero1999 thanks! debugging statements removed :)
ready to merge!

PR #44, opened 2024-10-03
Enhancement: Plotly boxplot violinplot

Gero1999:
Boxplots and violinplots have been also incorporated as plotly objects rendered in the UI

### To check
- [x] Correct display of both plotly objects in Outputs
- [x] Hovering reveals ID information of each point 
- [x] Value distribution of each AUC interval is also available and correct


js3110:
@Gero1999 The boxplots look good however for the violin plots I am getting Error: argument 1 is not a vector. Please take a look again and see if you can fix :)

PR #58, opened 2024-10-09
Enhancement/flexible axis boxplots

Gero1999:
## Issue

Closes #

## Description
Boxplots include now more flexibility for the user to choose in the X axis as many variables as they want as well as to filter values or to choose how to color the boxplot 

## Definition of Done
- [x] Create ability to change x axis based on user input selecting Grouping category
- [x] Include options to group by more than one category (eg Dose and Age)- similar to descriptive stats

## How to test
NCA > Choose the analyte > Submit > Settings > Choose the Dose Number > Run NCA > Outputs > Dose Escalation > Boxplot

## Contributor checklist
- [x] X axis changes based on user input with 1 or more categories
- [x] Color changes based on user input with 1 or more categories
- [x] Boxplots are modified accordingly when values are filtered

### DIFF
--- a/R/flexible_violinboxplot.R
+++ b/R/flexible_violinboxplot.R
@@ -58,7 +58,7 @@ flexible_violinboxplot <- function(boxplotdata,
 
   # Make the plot
   p = ggplot(data = box_data %>% arrange(!!!syms(colorvars)),
-             aes(x = interaction(!!!syms(xvars)), 
+             aes(x = interaction(!!!syms(xvars), sep = '\n'), 
                  y = PPORRES, 
                  color = interaction(!!!syms(colorvars)))) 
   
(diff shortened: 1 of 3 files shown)


js3110:
@Gero1999 X axis values not updated if grouped by >1 variable- only shows the values for one variable. Either x axis group needs to be limited to one variable, or needs multiple x axis lines to show the different values. I think one variable is probably better and then other grouping variables can be done by colouring.

js3110 (changes requested):
- [x] X axis values need to match all grouped variables

Gero1999:
@js3110 would you mind taking me a picture? I am not sure of the problem...

This is what I get when I select for example DOSNO, SEX, RACE:

Screenshot 2024-10-10 133216

js3110 (approved):

PR #61, opened 2024-10-10
Bug: Running NCA directly crushes the app

### LINKED ISSUE #56: Bug: Running NCA directly crushes the app
### Description
Error when running NCA without submitting an analyte. It crushes the app without letting the user to solve the issue. 

### To reproduce
Directly run NCA on any dataset ("Run NCA", top right)

### Expected behaviour
The app should not crush and an error message should appear for the user indicating them that they should select and analyte in "NCA" and press "Submit"

js3110:
## Issue

Closes #56

## Description

App does not crash if analyte is not selected and NCA still run.

## Definition of Done

- [x] The app should not crush
- [x] error message should appear for the user indicating them that they should select and analyte in "NCA"

## How to test
NCA -> Settings -> Run NCA

### DIFF
--- a/inst/shiny/tabs/nca.R
+++ b/inst/shiny/tabs/nca.R
@@ -215,8 +215,9 @@ observe({
 # Choose dosenumbers to be analyzed
 
 observeEvent(input$submit_analyte, {
+  req(mydata())
   updateSelectInput(session, inputId = 'cyclenca', label = 'Choose the Dose Number:', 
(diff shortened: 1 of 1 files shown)


Gero1999 (changes requested):
@js3110 it seems that the message now pops up, but the analyte cannot be submitted because the App crushes

#### To Do
- [ ] Keep app able to operate: Submit analyte > Select profile > Run NCA

Gero1999 (approved):
Now it works fine! :)

PR #103, opened 2024-10-25
Enhancement: Mean plots CV%

### LINKED ISSUE #95: Enhancement: Mean plots show CV%
### Description
Mean plots additional option to show ~~CV%~~ **Confidence Intervals (95%)** as well as standard deviation

### Definition of Done
- [ ] ~~option to show CV% bars on mean plots~~
- [ ] Creation of Confidence Intervals as shaded areas in mean plots (user change of mind)

js3110:
## Issue

Closes #95 

## Description

Mean plots additional option to show CV% as well as standard deviation

## Definition of Done

- [x] ~~Mean plots show CV% when selected~~
- [x] Confidence intervals showed as shaded areas (user change of mind)

## How to test
Outputs -> Mean plots -> Show CV%

## Contributor checklist
- [x] New logic is documented


R/general_meanplot.R
```diff
@@ -92,9 +98,15 @@ general_meanplot <- function(data,
...
-
+  # add sd
+  if (plot_cv) {
+    p <- p +
+      geom_errorbar(aes(ymin = (Mean - CV), ymax = (Mean + CV), color = id_variable), width = 0.4)
+  }
+  # Convert ggplot to plotly
+  p_plotly <- ggplotly(p)
+  
+  return(p_plotly)
```
m-kolomanski, line 109: nitpick: [According to the tidyverse guide](https://style.tidyverse.org/functions.html), one should only use `return()` for early returns. I am not a huge fan of this rule, so we can discuss changing it if you prefer - but would need to include those exceptions in our CONTRIBUTING.md file.

```suggestion
   ggplotly(p)
```

inst/shiny/tabs/outputs.R
```diff
@@ -161,7 +161,8 @@ output$meanplot <- renderPlotly({
                    selected_cycles = input$cyclesmean,
                    id_variable = input$selectidvar,
                    plot_ylog = input$logmeanplot,
-                   plot_sd = input$sdmeanplot) %>%
+                   plot_sd = input$sdmeanplot,
+                   plot_cv = input$cvmeanplot) %>%
```
m-kolomanski, line 165: suggestion: I would suggest using `snake_case` for new names to be more readable. I know it will not be consistent with previous naming, but I think we should slowly implement better practices. If changing other related input names is not complicated (not used in many places), I would suggest changing them all within the scope. Also, if the name has multiple parts to it, please try going from the most general to the most specific:

```suggestion
                   plot_cv = input$mean_plot_cv) %>%
```

inst/shiny/ui.R
```diff
@@ -332,6 +332,7 @@ fluidPage(
               uiOutput("selectidvar"),
               checkboxInput("logmeanplot", label = "Scale y Log"),
               checkboxInput("sdmeanplot", label = "Show SD"),
+              checkboxInput("cvmeanplot", label = "Show CV%"),
```
m-kolomanski, line 335:
```suggestion
              checkboxInput("mean_plot_cv", label = "Show CV%"),
```

tests/testthat/test-general_meanplot.R
```diff
@@ -0,0 +1,86 @@
+# Sample data for testing
+sample_data <- data.frame(
+  STUDYID = rep("Study1", 24),
+  USUBJID = rep(c("Subject1", "Subject2", "Subject3", "Subject4"), each = 6),
+  ANALYTE = rep("Analyte1", 24),
+  DOSNO = rep(1, 24),
+  EVID = rep(0, 24),
+  NRRLT = rep(1:6, 4),
+  AVAL = runif(24, 1, 100),
```
m-kolomanski, line 9: issue: Please avoid using dynamically generated data in your tests. It might cause unexpected failures, that are hard to reproduce. Use constant testing data, so that we know what cases we are covering.

m-kolomanski (changes requested):
Hey, good job in general, code works and is quite clean, thanks for including fine-grained the tests, however I have few comments:

issue(blocking): Code does not pass `devtools::lint()` check.

question, suggestion: For datapoints close to zero, CV is so large that the line is completely horizontal - is that expected and correct behaviour? Migth be worth setting some limits to keep the plot readable?

suggestion: When both SD and CV are selected, they will overlap each other. Might be worth thinking about different approach to display them, or allow only one at the time (in the UI, if one is selected, the other is automatically deselected).

js3110:
@m-kolomanski Hi! I implemented the changes, except for adding a limit to CV. 
The feature of using CV was requested for preclinical data, where the CV is within nicer ranges as the y values are much higher. I tested with a preclinical dataset and it looked fine. I think for clinical they mainly use SD and idk if its worth setting limits because the plot then won't be that accurate/useful? lmk what you think!

m-kolomanski (approved):
Hej @js3110, thanks for implementing the changes. The PR looks good to merge. With SD and CV being excluding each other, the plots feels much more responsive and the issue of having large CV values is less of a problem now, I think it is fine to let the user decide what they need based on their data.

Gero1999:
Hello @js3110, thanks for implementing this! Would it be okay to instead of error bars use bands? I think the user was referring to something more like this for CV%:

image

Gero1999 (changes requested):
- [ ] Implement CV ranges as confidence bands
- [ ] Check if CV representation should be as addition/substraction of % (is that ok?)
- [ ] Make sure CV representation / or SD error bars are well adapted when semiLOG plots are displayed

Gero1999 (changes requested):
- [ ] Coefficient of Variation should scale well in the plot, don't use %
- [ ] Coefficient of Variation should be symetrical

I gave it a look, I would suggest just simply a range such as `[Mean - CV,  Mean + CV]`

js3110:
Updated issue following user feedback:

- [ ] Plot has option for standard deviation bars
- [ ] Confidence interval 95% ribbon

Gero1999 (changes requested):
Just to put some context to also @m-kolomanski, we talked with the specific user that requested this feature, and they concluded it would actually be better to report confidence intervals. If it sounds well I will clarify that in the #ToDo of both the issue and the PR

js3110:
@m-kolomanski @Gero1999 
I have updated now with an attempted fix for the CI log scale.
Another issue is that in testing I can't get it to pass the final test that says it is able to plot in log scale. I can't figure out why this fails as in the app the log scale is plotted. If either of you can take a look and figure it out I would appreciate that!

Gotfrid:
@js3110 Hi, I took a look at your problem and I'm afraid there is no easy solution to it. The thing is that `ggplotly()` doesn't not correctly transform information about log-scaled yaxis. To have a "true" log-yaxis one would need to create a visualization from scratch in plotly. Check this example for reference:

```R
sample_data <- data.frame(x = 1:100, y = 1:100)

p_ggplotly <- (\(x) ggplot2::ggplot(data = sample_data, aes(x = x, y = y)) + geom_line())() |>
  ggplotly() |>
  layout(yaxis = list(type = "log"))

p_ggplotly$x$layout$yaxis$type

p_plotly <- plot_ly(
  type = "scatter",
  mode = "lines",
  x = 1:100,
  y = 1:100
) |>
  layout(
    title = "Log Plot",
    xaxis = list(title = "X-axis"),
    yaxis = list(title = "Y-axis", type = "log")
  )

p_plotly$x$layoutAttrs[[1]]$yaxis$type
```

Please display both plots to see the difference: yaxis from ggplotly is a categorical axis with finite options, while yaxis from plotly itself is "special" in a way it is rendered, and interacted with.

js3110:
@Gotfrid ok I see, thats what I also was able to find. Do you know if there is another variable that confirms that the ggplotly has created a 'log' scale (even if its not defined as such)? just something I can use in the testthat instead

Gotfrid:
@js3110 not anything that I'd know about. One thing we could do, is add a custom attribute to the output ggplotly object, i.e.

```R
if (plot_ylog) {
  p <- p + ....
  pl <- ggplotly(p)
  attr(pl, "yaxis_type") <- "log"
  return(pl)
}
```

and then in the test:

```R
has_log_scale <- attr(p, "yaxis_type") == "log"
expect_true(has_log_scale)
```

However, I don't think that this test would be very meaningful. It only tests that the branch of code responsible for adding log transform was executed.

Gotfrid:
@js3110 I would keep a minimal test with `expect_s3_class` - this way you at least run this code with `plot_ylog = TRUE` argument and make sure that it doesn't throw an error.

But I also noticed, that the rendered plot doesn't have ticks on the Y axis. Is this expected? Cuz if not, this is something that could have been tested - with a certain input data, you expect certain yaxis tickvalues that conform to log scale.

Gero1999:
> @m-kolomanski @Gero1999 I have updated now with an attempted fix for the CI log scale. Another issue is that in testing I can't get it to pass the final test that says it is able to plot in log scale. I can't figure out why this fails as in the app the log scale is plotted. If either of you can take a look and figure it out I would appreciate that!

I will give a look into it and propose a change!

Gero1999:
let me know if you think that will make the work! @js3110 @Gotfrid

js3110:
@Gero1999 nice work around! just changed a couple parts so that all the tests pass but I think it looks good now! thanks

Gero1999 (approved):

R/general_meanplot.R
```diff
@@ -81,20 +95,28 @@ general_meanplot <- function(data,
...
+  # add log scale
+  if (plot_ylog) {
+    p <- p +
+      scale_y_log10(breaks = c(0.001, 0.01, 0.1, 1, 10, 100, 1000),
+                    label = c(0.001, 0.01, 0.1, 1, 10, 100, 1000)) +
+      annotation_logticks(sides = "l")
+  }
 
+  # Convert ggplot to plotly
+  return(p)
```
m-kolomanski, line 119: nitpick: please use `return()` only for early returns

```suggestion
  p
```

m-kolomanski (changes requested):
The feature seems to behave very nicely from the UX standpoint, however I would like to point out some developer issues:

- Please address the warnings in the console:
Screenshot from 2024-11-07 12-10-02
Screenshot from 2024-11-07 12-15-17

- Please make sure the PR passes all checks. If spellcheck gives you trouble, you can rebase the branch so that the WORDLIST file is pulled from main.

R/general_meanplot.R
```diff
@@ -40,39 +42,65 @@ general_meanplot <- function(data,
       DOSNO %in% selected_cycles,
       if ("EVID" %in% names(data)) EVID == 0 else TRUE,
       NRRLT > 0
-    ) %>%
-    # rename(id_variable = id_variable) %>%
+    )
+
+  timeunit <- unique(preprocessed_data$RRLTU)
+  concunit <- unique(preprocessed_data$AVALU)
```
m-kolomanski, line 48: issue: please use snake_case for variable names
```suggestion
  time_unit <- unique(preprocessed_data$RRLTU)
  conc_unit <- unique(preprocessed_data$AVALU)
```

R/general_meanplot.R
```diff
@@ -40,39 +42,65 @@ general_meanplot <- function(data,
...
   # plot the preprocess data
-  p <- ggplot(data = preprocessed_data, aes(x = NRRLT, y = Mean), group = id_variable) +
+  p <- ggplot(data = summarised_data, aes(x = NRRLT, y = Mean), group = id_variable) +
     geom_line(aes(colour = id_variable)) +
     geom_point(aes(colour = id_variable)) +
     facet_wrap(~groups,
                strip.position = "top") +
     labs(
-      x = paste0("Nominal Time [", preprocessed_data$RRLTU[1], "]"),
+      x = paste0("Nominal Time [", timeunit, "]"),
```
m-kolomanski, line 98:
```suggestion
      x = paste0("Nominal Time [", time_unit, "]"),
```

R/general_meanplot.R
```diff
@@ -40,39 +42,65 @@ general_meanplot <- function(data,
...
       y = paste0(
-        "Mean concentration", " [", paste(unique(preprocessed_data$AVALU), collapse = ","), "]"
+        "Mean concentration", " [", concunit, "]"
```
m-kolomanski, line 100:
```suggestion
        "Mean concentration", " [", conc_unit, "]"
```

m-kolomanski (approved):
Hey, looks fine, I have just a minor issue regarding the naming of variables.

I can see that spellcheck and lintr CI are failing, but looking at the errors, I think it is due to lacking WORDLIST and some rules from .lintr on this branch. Rebasing to main or copying those files over should solve the issue, but might create some conflicts that might have to be solved before merging.

js3110:
> Hey, looks fine, I have just a minor issue regarding the naming of variables.
> 
> I can see that spellcheck and lintr CI are failing, but looking at the errors, I think it is due to lacking WORDLIST and some rules from .lintr on this branch. Rebasing to main or copying those files over should solve the issue, but might create some conflicts that might have to be solved before merging.

@m-kolomanski i tried to rebase but it doesn't seem to be working as it doesn't show me any conflicts it just changes the file to the main version. Are we not able to merge anyway despite the spellcheck fail?

m-kolomanski:
@js3110 I have rebased the branch, I believe all checks should pass, when they do you are free to merge

PR #106, opened 2024-10-31
Enhancement: implemented relevant CI workflows from admiralci

### LINKED ISSUE #64: Build: Setup CI

m-kolomanski:
## Issue

Closes #64 

## Description

Adds `.github/workflows/main.yaml` file, which will run relevant checks from [admiralci](https://github.com/pharmaverse/admiralci/tree/main) repository:

- spellcheck
- lintr
- man pages, checking if documentation is up to date with docstrings
- R CMD check, with note: since we have some warnings already, I have made it so the check will pass even with warnings. In the future, once we clean those up, it would be beneficial to enable the check to fail on warnings too

In addition, the workflow will also run `devtools::test()`.


Gero1999 (approved):
Hello! I saw you need 2 reviewers, not sure if I need to check anything else, but what I can confirm:

✅Re-run of the workflow produce same annotations and checks 
✅ Functionalities of the App are kept after file changes (nca.R, general_lineplot.R ...)

README.md
```diff
@@ -77,7 +77,7 @@ devtools::load_all()
 aNCA::run_app()
 ```
 
-Have fun running your tailored Non-Compartemental Analysis!
+Have fun running your tailored Non-Compartmental Analysis!
```
Gotfrid, line 80: question: I can see that there are still a few places in the repo that use `Non-Compartment Analysis`. Should they be changes to `Non-Compartmental Analysis` as well in the scope of this PR?

.github/workflows/main.yml
```diff
@@ -0,0 +1,80 @@
+# Source: https://github.com/pharmaverse/admiralci
+# Stripped down version of admiralci checks. Good for
+# the developement process. When package is ready to
+# be published, revisit and add release-related
```
Gotfrid, line 4: suggestion: might be good to list every missing workflow/job that is release-related. This way it will be easier to come back to it later and just copy-paste necessary stuff.

Gotfrid (approved):
Looks good. I've added non-blocking comments/questions.

Gotfrid (approved):

PR #111, opened 2024-11-08
97 bug duplicated input id for submit tlg order

### LINKED ISSUE #97: Bug: duplicated input ID for "submit_tlg_order"
### Description
In file `inst/shiny/modules/tab_tlg.R`, line 9 and 11 have button with the same ID.

### To reproduce
Run `shiny::devmode(T)` when testing the application to see such errors and warnings.

### Expected behaviour
IDs should be unique for the whole app.

Gotfrid:
## Issue

Closes #97 

## Description

Not sure if the second button is needed at all, but I assumed that it is there for a reason. So I simply updated the second button's id to avoid collision, and updated the observer to listen to both buttons.

Another problem that I solved is the one with conditional panels. As explained in the commit message, the problem's source is the id name itself. Conditions in conditionalPanel are evaluated in javascript, and dots in id value (e.g. `rule_adj.r.squared`) are treated the JS way, producing an error when trying to index into `undefined`.

## Definition of Done

- [x] There are no console errors in a browser's devtools console
- [x] `shiny::devmode(T)` works as expected

## How to test

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

Please test the app functionally to make sure that all the functionalities are still working.

### DIFF
--- a/inst/shiny/modules/tab_tlg.R
+++ b/inst/shiny/modules/tab_tlg.R
@@ -8,7 +8,7 @@ tab_tlg_ui <- function(id) {
       actionButton(ns("remove_tlg"), "Remove TLG"),
       actionButton(ns("submit_tlg_order"), "Submit Order Details"),
       DTOutput(ns("selected_tlg_table")),
-      actionButton(ns("submit_tlg_order"), "Submit Order Details")
+      actionButton(ns("submit_tlg_order_alt"), "Submit Order Details")
     ),
     tabPanel(
       "Tables",
@@ -239,7 +239,7 @@ tab_tlg_server <- function(id, data) {
     })
 
     # When the user submits the TLG order...
-    observeEvent(input$submit_tlg_order, {
+    observeEvent(list(input$submit_tlg_order, input$submit_tlg_order_alt), {
       tlg_order_filt <- tlg_order()[tlg_order()$Selection, ]
 
       if (sum(tlg_order_filt$Type == "Table") > 0) {
--- a/inst/shiny/tabs/nca.R
+++ b/inst/shiny/tabs/nca.R
@@ -135,7 +135,7 @@ observeEvent(input$settings_upload, {
 
   # RSADJ
   if (!is.na(setts$adj.r.squared_threshold[1])) {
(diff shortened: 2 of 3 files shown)


Gero1999 (changes requested):
Hello @Gotfrid, good catch and thanks for the changes, the UI now looks better!

However, the server produces this error when using the Rule Sets:

```
Warning: Error in [[<-: Assigned data `{ ... }` must be compatible with existing data.
✖ Existing data has 61 rows.
✖ Assigned data has 0 rows.
ℹ Only vectors of size 1 are recycled.
Caused by error in `vectbl_recycle_rhs_rows()`:
! Can't recycle input of size 0 to size 61.
```

If I had to guess, I would give a look to the changes you made of the inputId for `rule_` objects. The old IDs using a dot (i.e,` rule_adj.r.squared`)  where put like that because the package PKNCA() produces as result this parameters in myres$result (i.e, adj.r.squared). 

The rule sets are used to flag those results based on this parameters, so at some place in nca.R a loop is used over the inputs to mark accordingly each result. Feel free to change either that flagging strategy or the inputId back to the old name that aligns with PKNCA results!

- [x] Recover server functionality (flagging results)

Gotfrid:
@Gero1999 Thank you, exactly the kind of check that I needed. Your guess is on point: I can't change rule_ ids back due to the limitation described in the PR description, however I am changing underscores back to dots in the same place where rule ids are already being processed (see the latest commit).
Hope it makes sense!

Gero1999 (approved):
seems good to me, good job!

m-kolomanski:
Hey @Gero1999, can we get a confirmation if we need the second submit button and if so, then what is the exact purpose?

Gero1999:
> Hey @Gero1999, can we get a confirmation if we need the second submit button and if so, then what is the exact purpose?

Ups you are right sorry @Gotfrid @m-kolomanski , the second button was just in case the user had a very long TLG table order and decided to submit the order from below while checking, but this was completely my call so feel free to keep or change it!

m-kolomanski (approved):
LGTM

Issue #100, opened 2024-10-24
Enhancement: Add labels to ADNCA


Shaakon35:
### Description
When the user loads ADNCA, sometimes there could be some column labels, sometimes not.
What we need to do is to populate all the labels based on the ADNCA guidelines.
If the variable is in the list: populate the label, if not, populate by the name of the variables:
e.g. ADNCA$ASIAN does not have labels, then populate: 
attr(ADNCA[["ASIAN"]], "label") = "ASIAN"

some code to start with:
`
# Set label list 
label_ADPC <- list(
   USUBJID = 'Subject ID',
   TRT01A  = 'Treatment Group',
   AVAL    = paste0('PK Concentration\n (', unique(data$adpc$AVALU), ')'),
   AVALC   = paste0('PK Concentration\n (', unique(data$adpc$AVALU), ')'),
   NRELTM1 = paste0('Nominal Time \nfrom First Dose (hr)'),
   ARELTM1 = paste0('Actual Time \nfrom First Dose (hr)'),
   NRELTM2 = paste0('Nominal Time \nfrom Most Recent Dose (hr)'),
   ARELTM2 = paste0('Actual Time \nfrom Most Recent Dose (hr)'),
   # NOMDOSE = 'Nominal Dose',
   AVISIT  = 'Visit'
)

# Apply manual labels
for (var in names(label_ADPC))
  if (var %in% colnames(data$adpc))
    attr(data$adpc[[var]], "label") <- label_ADPC[[var]]
`

```


as_factor_preserve_label <- function(x) {  
  label <- attr(x, "label")  
  x <- as.factor(x)  
  attr(x, "label") <- label  
  return(x)  
}  

has_label <- function(x) {  
  return(!is.null(attr(x, "label")))  
}  

set_empty_label <- function(x) {  
  if (is.null(attr(x, "label"))) {  
    attr(x, "label") <- ""  
  }  
  return(x)  
}  
```

### Definition of Done
The requirements for the feature to be complete:
- [ ] if you load a csv file, all the variables should have a label (even if the variable is not standard)
- [ ] if you load a rda file, all the variables should have a label (even if the variable is not standard)
- [ ] make sure that if a variable is a factor with levels, the levels are kept!

Gero1999:
I think this implementation will make more sense after #112 and will make things easier for any case!

js3110:
@Shaakon35 @Gero1999 
[...]

Issue #54, opened 2024-10-07
Enhancement: Export the plot as PDF with Citril and Chevron


Shaakon35:
### Prerequisites
https://github.com/pharmaverse/aNCA/issues/53
This issue need to be finished before starting.

### Description
Use Chevron and Citril packages to generate pdf version of the plot created here: https://github.com/pharmaverse/aNCA/issues/53

### Definition of Done
The requirements for the feature to be complete:
- [ ] Be able to download the pdf version of the plot 
- [ ] Make sure to have the footnotes: "program: /xxx/g_pkconc_LIN.R
output: /xxx/g_pkconc_LIN_ANALYTE.pdf" (xxx being getwd(), ANALYTE being the ANALYTE that was selected)
- [ ] Make sure to have the title "Study: XXXX", extract XXX from unique(adpc$STUDYID) - there should be only one studyid
- [ ] Below the footnotes, there should also be the time : 03OCT2024 09:23 (eventually, if it's possible add the timezone)
- [ ] There should be the pagination Page 1/20 automatically created by Citril

m-kolomanski:
Hey, I would like to ask for clarifications on that one. I have never used `Chevron` or `Citril`, so I struggle to see how this fits together.

For starters, for some reason I cannot find any reference to a package or tool called `Citril` - would you mind sharing some links and/or docs?

From what I can see, `Chevron` is a package for generating ready TLGs. We already have semi-custom functions for generating the TLGs that we want (at least the ones we have implemented to date). So one thing is that we are not already using `Chevron` and I do not see currently implemented TLGs in the `Chevron` catalogue - so I struggle to understand how `Chevron` can aid us in solving this task.

I would love to know what I am missing, thanks!

Gero1999:
@Shaakon35 made some modifications, also this will be blocked until solved #428  

@m-kolomanski I have never used Citril, @Shaakon35 let us know if you are keen in using it a good documentation resource

Shaakon35:
@KOBANAK Could you please provide all the documentation you have for Citril and Chevron please?

PR #124, opened 2024-11-15
Enhancement: setup system for TLG (plots)

### LINKED ISSUE #53: Enhancement: Create a first TLG with teal components
### Prerequisites
https://github.com/pharmaverse/aNCA/issues/51 ✅ 
https://github.com/pharmaverse/aNCA/issues/52 ✅ 

### Description
We would like to implement one single TLG with the teal components, let's start by the easiest one (the individual plot only):
 https://insightsengineering.github.io/tlg-catalog/stable/graphs/pharmacokinetic/pkcg01.html

### Definition of Done

- [ ] **Generation of pkcg01 plots.** When "Submit Order Details" is pressed in the TLG tab  including the row [pkcg01](https://insightsengineering.github.io/tlg-catalog/stable/graphs/pharmacokinetic/pkcg01.html), in the sub-tab "Graphs" the user will see all related pkcg01 plots generated with the data (Dataset: ADPC = data()) .

- [ ] **Customization options for  plots.** Implement the customization inputs that will be used to modify all Graphs and that are included as argument functions in [pkcg01.R](https://github.com/user-attachments/files/17573808/pkcg01.txt). Customization options should be in the right panel. Take into consideration that in future implementations most inputs will be the same for all graphs (i.e, xlab, ylab, title, split-by), only some may be different (i.e, options(display="LOG")). 

- [ ] **Interactive ggplotly graphs.** Either by providing an input option for the user to transform plots into ggplotly or by producing them as this by default.

- [ ] **Navigation tool through pkcg01 plots.** Left and right arrows to see different patient profiles. Another potential option will be to filter the split-by variable values, i.e, from subject one to subject 10 easily. One example would be a dropdown menu to select subjects (2x2 for example to see 4 patients)

### How to do
The script [pkcg01.R](https://github.com/user-attachments/files/17573808/pkcg01.txt) is used to generate all plots by default when the function is called. The script [graphs_add-in_functions.txt](https://github.com/user-attachments/files/17573861/graphs_add-in_functions.R) includes some extra functions that are used by pkcg01.R and could probably be recycled for other similar scripts.  

Feel free to make as many changes as needed in these scripts, the App or the metadata www/data/TLG_order_details.csv. The purpose after this implementation is to escalate it to the rest of the graphs. We will take your final implementation as a reference for building the rest of the functions accordingly.

m-kolomanski:
## Issue

Closes #53

## Description

- Introduces new module and specification system allowing for converting simple functions that generate plots into dynamic plot display that allows the user to customize plot attributes.
- `tlg_plot` module responsible for displaying plots and creating widgets for plot option customizations, which are then dynamically passed to rendering functions.
- `tlg.yaml` file which allows specification of TLG metadata (descriptions, links etc) for compact table display, as well as rendering function and TLG options which are then displayed as shiny widgets and editable.
- Introduces `pkcg01` plot as an example of system implementation.
- Documentation of the system is provided in .github/contributing` folder. 

## Definition of Done

- System for defining TLGs (plots):
  - Allowing the user for quick definition of new plots
  - Easy implementation of variables editable by the user
  - In editable widgets, allow for referencing columns and labels using special key characters _($COLNAME, !LABEL)_
  - Plots rendered using `plotly` for interactivity
  - Documentation available
- First template plot:
  - pkcg01 function implemented and adjusted to fit definition system
  - plots with linear and log scales implemented

## How to test

For basic testing, run the app and **Submit mapping** in the _Mapping and filters_ tab. Then you can go to _TLG_ tab and **Submit order** two default plots. When the plots are rendered, you can test changing the values in the widgets.\

For more advanced testing, please do try to add a new plot. The details are described in `.github/contributing/add-tlg.md`. Please check if the documentation is clear and complete. Adding a new plot will require writing a definition in the `tlg.yaml` file and providing a generating function, which could render a very basic plot.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

Please do review not only the code and functionalities, but also the documentation closely. If anything is confusing, missing or could be explained better - please do let me know. We want the system to be usable by contributors without the need of an introduction on a call.


Gero1999:
Hello @m-kolomanski the idea seems amazing, pretty much what we talked and the yaml provides more flexibility. Only thing I am wondering is that the yaml file should also specify for each option if the input should be text, numeric, choice... And in case it is for example a selectInput multiple-choice we will need to connect it to the dataset and specify multiple=TRUE. Are you thinking in:

1) Adding explicitely in the yaml for each input in opts the type of input
2) Guessing the type of input is needed by some conventional naming

Gero1999:
Hello @m-kolomanski , I see that this is starting to look really cool! I will also ask for feedback from Valentin on this, but I will give you some points regarding my perspective that I noticed:

- I love that now we have subtabs instead of radio buttons for selecting the plots. I think it is definitely more aesthetic! Another aesthetic detail could be that when you submit the order details you directly are driven to a populated tab (Tables, Listings or Graphs).

- The general idea seems to be fitted well! Although I would think about the possibility of displaying more than 1 plot more like a scroll down of plots so the user can check multiple at the same time.

- Page navigation control problem. When pressed fast more than two times "Next page" will start acting on its own

- Labels are displaying literal column names instead of their values. I would also think if there are possibilities that when changing the labels one could not only introduce literal text but also column values like $AFRLT so that is not lost, if that makes sense! I would personally put it as default populated like that so everything keeps in "one place", but of course this is just a developer/personal idea. We can discuss if this could also be a separated issue instead so the task is not overloaded!

Let us know if you agree with everything stated and with the feature developed so far @Shaakon35

Shaakon35:
Hello @m-kolomanski @Gero1999 
Thanks for the progress on this. Looks great, I will just add a few points and confirm the one of Gerardo:
- Bug when you click more than 1 time on the Next page button
- Footnote is not working. We would have a check box next to the footnote: Apply to all.
-> This way you can add a footnote for all plots, or only one plot (one subject). Therefore, we would also need a button Save to save the few things that we changed. (so that when we export all the TLGs in pdf, all the things changed are kept)
- By default, the selectInput on the right for xlab and ylab should be populated.
- Indeed, as Gerardo said, the xlab and ylab should be replaced by the labels of the column AVAL or AFRLT, and in parenthesis ( the unique value of RRLTU or AVALU - it needs to be unique otherwise it does not work)
- To make it unique, we need to have two more buttons on the plot: selection of PARAM (Analyte 1, Analyte 2), and selection of ASMED (Plasma, Urine, etc.)
- Can we make the left sidebar smaller? Also, can we make the plot ggplotly?
- Showing multiple plots on one page should be an option.

I think we are close to what we need now already, thanks a lot for all the work!

Best
Valentin

m-kolomanski:
@Gero1999 @Shaakon35 
Thank you for the feedback, the issues you have mentioned will be addressed (or were fixed today).

I have one question regarding the number of plots per page, do we want to keep the pagination at all? Or do we just want to display all plots on single page and scroll through them? Do we want similar pagination abilities to the slope selector? What would be the best solution?

Gero1999:
Good point. I am actually thinking that maybe all plots at once would be better. 

An alternative idea could be to have some selection controls like in slope plots so the user can make edition on the set of plots they choose by selectInputs or similar (i.e, change the footnote of all plots with the X analyte). But maybe the idea is too crazy, let me know what you both think!

@m-kolomanski @Shaakon35

R/g_pkconc_ind.R
```diff
@@ -0,0 +1,350 @@
+#' Wrapper around aNCA::pkcg01() function. Calls the function with `LIN` scale argument.
+#' @param data Data to be passed into the plotting function.
+#' @param ...  Any other parameters to be passed into the ploting function.
+#' @returns ggplot2 object for pckg01.
+#' @export
+g_pkconc_ind_lin <- function(data, ...) {
```
Gotfrid, line 6: FYI: [purrr::partial](https://purrr.tidyverse.org/reference/partial.html)

inst/shiny/tlg.yaml
```diff
@@ -0,0 +1,86 @@
...
+g_pkconc_ind_lin:
+  is_default: true
+  type: Graph
+  dataset: ADPC
+  pkid: pkcg01
+  label: pkcg01 - linear
+  description: "Individual plots of concentrations vs. time (one plot per patient/subject) with linear scale"
+  link: https://insightsengineering.github.io/tlg-catalog/stable/graphs/pharmacokinetic/pkcg01.html
+  fun: g_pkconc_ind_lin
+  opts:
```
Gotfrid, line 25: question: why not "options"? :)

inst/shiny/modules/tab_tlg.R
```diff
@@ -87,10 +67,33 @@ tab_tlg_server <- function(id, data) {
...
+      )
+
+      purrr::iwalk(.TLG_DEFINITIONS, function(x, id) {
+        tlg_data <<- dplyr::add_row(
+          tlg_data,
+          id = id,
+          Selection = x$is_default,
+          Type = x$type,
+          Dataset = x$dataset,
+          PKid = paste0("<a href='", x$link, "'>", x$pkid, "</a>"),
```
Gotfrid, line 91: suggestion: might be a good idea to have a link with target="blank_"

inst/shiny/modules/tab_tlg.R
```diff
@@ -87,10 +67,33 @@ tab_tlg_server <- function(id, data) {
...
+        Dataset = character(),
+        PKid = character(),
+        Description = character(),
+        Footnote = character(),
+        Stratification = character(),
+        Condition = character(),
+        Comment = character()
+      )
+
+      purrr::iwalk(.TLG_DEFINITIONS, function(x, id) {
```
Gotfrid, line 84: suggestion: please use `purrr::reduce` for this type of logic

inst/shiny/modules/tab_tlg.R
```diff
@@ -267,19 +270,26 @@ tab_tlg_server <- function(id, data) {
...
-          label = "Graph to display",
-          choices = tlg_order_filt$Label[tlg_order_filt$Type == "Graph"]
-        )
-      } else {
-        updateRadioButtons(session = session,
-                           inputId = "buttons_Graphs",
-                           label = "",
-                           choices = "")
-      }
+      tlg_order_graphs <- filter(tlg_order_filt, Type == "Graph")$id
```
Gotfrid, line 273: suggestion: please don't mix dplyr and base.
either
```r
tlg_order_filt[tlg_order_filt$Type == "Graph", "id"]
```
or 

```r
tlg_order_filt |> filter() |> pull()
```

Gotfrid (commented):
Overall idea with YAML configuration is a great direction in my opinion, @m-kolomanski. Good work on that. I left a few comments here and there.

However, I think that the current state of the app is crashing - see the recording.

https://github.com/user-attachments/assets/4fc05b67-60c1-4c57-90e2-d9c704659a3c

Shaakon35:
@m-kolomanski 

Regarding the pagination, my idea is first to wrap up this first graph. I don't think we need any pagination in the TLG tab, we can just click next, and that should be enough.

Then, we will think about pagination in a PDF for the plots:
https://github.com/pharmaverse/aNCA/issues/54

The idea is to have 1 plot per page on the pdf.

I have seen that you created two new buttons for X-axis and Y-axis, I think we should not let the user configure the plot too much as it will complicate the bug fixing of errors. I think we should make it simple for now.

Gero1999:
Hello @m-kolomanski, actually the current layout starts to look really really amazing. So to keep it simple I would include here a list of the intended inputs that should be in all plots, and also their default values for pkgc01 (which I guess may be probably specified in the function itself or somewhere you can think of):

#### Global aspects
- [ ] When user presses `Submit TLG order` update Tab to the first populated page
- [ ] By default plots per page = All
- [ ] Change of buttons order, from left to right: `Plots per page`, `Previous page`, `Page X out of N`, `Next page`
- [ ] If easy/possible include a reset button to change all input values to the default settings

#### General
- **X axis:** `AFRLT`
- **Y axis**: `AVAL`
- **Grouping variables**: `USUBJID, PARAM, PCSPEC, ROUTE` 
- [ ] **Splitting variables** (facet_wrap): ` ` (empty by default)
- [ ] **Color variable**: ` `

#### Label
- [ ] Title: `Plot of PK concentration-time profile - Linear Scale, by Cohort, PK population $STUDYID$`
- [ ] Subtitle: `\USUBJID\: $USUBJID$[\n]
     \ PARAM\: $PARAM$[\n]` 
     \PCSPEC\: $PCSPEC$[\n]
      \ROUTE\: $ROUTE$[\n]
 - X label: `\AFRLT\ ($RRLTU$)`
 - Y label: `$PARAM$ ($AVALU$)`
 - [ ] X breaks/ticks: `NFRLT`
 - [ ] Y breaks/ticks: ` `
- [ ] Color label: ` ` 
- Footnote: ` ` 

#### Limits: 
- xmin: ` `
- xmax: ` `
- ymin: ` `
- ymax: ` `

So as you may have noticed for text inputs it would be cool if the user could see how things are done so they can modify them easily themself, the way I did it here is just an example, but feel free to change the encoding system:

- Something between $X$ would be substituted by the corresponding column value:  i.e, `$PARAM$` = "DummyRO"
- Something between \X\ would be substituted by the label: i.e, `\PARAM\`="Parameter"
- If possible would be also cool if new lines can be interpreted so we can add them by default to the inputs: i.e, [\n]=<br/>

Gero1999:
@m-kolomanski . If you have problems covering all these, keep in mind we can just make the inputs and cover them in future issues/implementaitons and of course do not hesitate to ask for help, I can definitely give you a hand to adapt the function or whatever you need!

m-kolomanski:
Questions regarding the settings for the first plot:

For x axis and y axis, I am getting conflicting info, @Shaakon35 noted that:

> I have seen that you created two new buttons for X-axis and Y-axis, I think we should not let the user configure the plot too much

and I agree in general, I think if we have a plot definition it is kinda strange to let the user specify what is plotted on x and y axis? Please let me know if we want to include that option.

In addition, what exactly is `splitting variable`? Does not seem to be implemented in the current plotting function. Is this text or select input? How exactly is it supposed to work?

Gero1999:
You are right @m-kolomanski , let me check with @Shaakon35 that we indeed neither need X/Y axis in any case for figures.  For now leave it, as it is already implemented and I guess to quit it off is easy. 

Regarding the `splitting variables` they refer to the variables that are introduced to the R function `facet_wrap` to split a plot in multiple ones. It is not in the current function, however we can implement it later on, so let's leave it for now if it seems complicated and probably may end up needing subarguments (fix_x axis, fix_y axis..).

m-kolomanski:
Hey @Gero1999,

how do we want the title and subtitle to behave exactly? Currently both are specified in two places in the provided functions, both basically have two parts. What should happen if the user provides their own title/subtitle? Should the default be overwritten completely? Do we want to keep some parts (like Treatment Group information or Cohort) as always rendered, in addition to what the user provides?

Also question about providing dynamic values - we can do that easily, provided it is supported by the plotting function itself. So we can say that in the box the user will provide something like `.DOSEU` and the app will apply `unique(plot_data$DOSEU)`, so for simple example if user provides `Dose unit: .DOSEU` in the text widget, the app will render `Dose unit: mg`. Is that what you mean?

I do not understand what exactly is the label in `\PARAM\="Parameter"`. Where do we get the label from...?

In terms of adding new lines, we can already achieve that by adding `<br>` tag, since the text is interpreted as html by plotly.

Gero1999:
Hello @m-kolomanski!

Right, I would say the best is that the default title is already default selection shown in the input. If the user wants to modify a part of it (or entirely) they should be able to. 

The dynamic component with `.DOSEU` makes perfect sense to me and seems intuitive enough. Perhaps `<br>` can also work, although based on our users I am not 100% sure.

Regarding the third topic, the plan is that the dataset/input will have labels associated to each of the relevant/recognized column names. These labels are a more human-readable way to comprehend the column meaning. This issue will hopefully be solved soon (#100 ), maybe would make more sense to implement it then, but I leave it to your criteria. One example on how it is assigned would be:

```
attr(adpc[["AFRLT"]], "label") <- "Actual time from first dose"
```

It would be really cool if at some point we figure out an easy way to also add those labels dynamically in the text (i.e, `\DOSEU\ (.DOSEU)` = "Dose Unit (mg/L)"). If not, we will simply use them when building the default standard and provide them for the user as regular text. Hope everything seems more clear now!

m-kolomanski:
Hey @Gero1999,

I have implemented working title/subtitle, with the ability to reference column names and the _label_ attribute, though it is not available in the data at the moment.

Please let me know how the feature looks **functionally** now and if there are any critical things missing. There is still some optimization and UI improvements to do, but I want to know if the basics of the system workflow are complete. We can of course extend that further in the future, if there is such a need.

Gero1999:
Hello @m-kolomanski great work indeed!

The dynamic parse will very much do the trick. As an initial setup this seems super good and without nothing relevant missing. If I had to personally provide feedback and be very picky I would only say:

* Maybe could be better if the user needs to define start and end of column values and labels (e.g, `$DOSNO$ or !DOSNO!`). I am just thinking in some potential confusing situations, like when you want to write `!AVAL!: $AVAL$ ($ANALYTE$)`. I know that rmd uses mroe this type of way to do things, and perhaps for that reasons I could think that it would prevent us problems. But please, let me know your opinion on this one!

*  It could be good if the users is capable of seeing the default input, perhaps even if we need to put it as an argument in the function so you can extract it using `formals`. 

The rest of aspects in the system workflow seems pretty good to me. Thanks for developing such a cool thing!

m-kolomanski:
Hey @Gero1999, thank you for the feedback!

> Maybe could be better if the user needs to define start and end of column values and labels (e.g, $DOSNO$ or !DOSNO!). I am just thinking in some potential confusing situations, like when you want to write !AVAL!: $AVAL$ ($ANALYTE$). I know that rmd uses mroe this type of way to do things, and perhaps for that reasons I could think that it would prevent us problems. But please, let me know your opinion on this one!

I have implemented it as `$`, since this is similar to just subsetting columns the base-R way (`plot_data$DOSEU`). I do not think there would be any technical issues, I have tested it and seems to work in complex scenarios. I would ask @Gotfrid to give a commment on that, please, when you are doing the technical review.

>It could be good if the users is capable of seeing the default input, perhaps even if we need to put it as an argument in the function so you can extract it using formals.

TBH that is something I have struggled with. The problem with filling the inputs with default values is that we have some that are based on dynamic values, like the subtitle. The default subtitle references the default grouping variables. So when the user changes the grouping variables, the subtitle will change automatically. If we fill the input widget with default definition, like `!USUBJID: $USUBJID, !PARAM: $PARAM, !PCSPEC: $PCSPEC, ROUTE: $ROUTE`, then this would overwrite the automatic behaviour and upon changing the grouping variables the subtitle would not change.

I do not see any clean way of achieving this functionality without writing a lot of custom, TLG-specific code in the modules, which defeats the purpose of the system a bit. I would like to leave it for now and maybe extend upon it in the future. I would suggest that empty widgets do not necessarily mean **default** behavior, but rather **automatic**, if that makes sense.

m-kolomanski:
Hi @Shaakon35,

1. In terms of lines cutting out, that is because the data consist of points that are equal to 0, which causes `log10` to produce `-Inf`, so the points fall out of the range of the plot. What is the expected behavior what we wish to achieve?
2. In terms of ticks, sadly `plotly` completely ignores the ticks and annotations set by `ggplot2` (I have tried hard to make this work previously, to no avail) and does not offer much customization options. I am afraid if we wish to keep the same level of interactivity that `plotly` offers, we will have to deal with the drawbacks.

Gero1999:
I would also reconsider the need for `plotly` here to save speed @Shaakon35. In the end, this part is not exploratory but has only reporting intentions and once printed in the pdf-report plotly will not be reactable!

Shaakon35:
Hi @m-kolomanski @Gero1999 
I agree with your comments,  checked an the ticks don't appear anymore on the second plot when I switch to ggplotly...
There might be a workaround, but let's focus on it later as it's a display issue.

Thanks a lot, all good for me.

m-kolomanski:
@Shaakon35 
I want to confirm, in relation to comment from @Gero1999, we are committing to `plotly`, correct? This decision is kinda important, since both the module and the plot-generating functions will be suited to `plotly` and changing this in the future will be hassle, since it will involve updating all the graph functions.

I do not mind either way, both approaches have pros and cons, I just wanna confirm we are reasonably happy with the choice.

Gotfrid:
Hi @m-kolomanski , I've looked into the busy indication. Unfortunately, shiny's busyIndicator simply doesn't work with the uiOutput, and it's not clear if it will work in the near future (I went through the [discussions](https://github.com/rstudio/shiny/issues/4115), and apparently only a change for tableOutput is in the making).

I'm afraid that the only way to solve this would be to use `shinycssloaders` - that way seems to work fine. I will push my commit, so that I can move on. If you disagree, feel free to revert it.

Also, while testing, I noticed that the plots are rendered twice, thus doubling the overall loading time. And also, if I "submit order details" second time, there are no plots visible on the screen.

m-kolomanski:
@Gotfrid If inbuild functionalities fail, I see no issue with using `shinycssloaders`, it is a good package. Thank you for taking the time to look into this!

.github/contributing/adding-tlg.md
```diff
@@ -0,0 +1,64 @@
...
+
+## Tables
+Currently not implemented
+
+## Graphs
+### yaml
+The `Graph` entry should have the following format:
+ ```yaml
+# unique identifier for given entry
+id:
```
Gotfrid, line 20: suggestion: here and with `optid` consider indicating more clearly that the text is supposed to be replaced by the developer. for example, `<id>`, or `<enter plot id>`, or something else maybe?

inst/shiny/modules/tab_tlg.R
```diff
@@ -1,96 +1,83 @@
+.TLG_DEFINITIONS <- {
```
Gotfrid, line 1: this is probably the most obscure object (in terms of its names) that i've ever seen to this day :)

inst/shiny/modules/tab_tlg.R
```diff
@@ -1,96 +1,83 @@
...
-          6,  # Middle column for plot output
-          h4("Table to display"),
-          plotOutput(ns("plot_Tables"))
-        ),
-        column(
-          2,  # Right column for plot customization inputs
-          h4("Inputs with selected vals linked to downloadable obj (i.e, tlg_order())"),
-          textInput(ns("footnote_Tables"), label = "Footnote")
+      card(
+        style = "margin-top: 1em;",
```
Gotfrid, line 36: suggestion: i noticed recently that ppl use `htmltools::css()`. probably it makes R folks more comfortable reading this stuff. just FYI, not actionable.

inst/shiny/modules/tab_tlg.R
```diff
@@ -239,9 +226,19 @@ tab_tlg_server <- function(id, data) {
       }
     })
 
+    # Toggle submit button depending on whether the data is available #
+    observeEvent(session$userData$data(), ignoreInit = FALSE, ignoreNULL = FALSE, {
```
Gotfrid, line 230: suggestion: you already "localized" data before, so maybe you can use `data()`

inst/shiny/modules/tab_tlg.R
```diff
@@ -268,19 +265,45 @@ tab_tlg_server <- function(id, data) {
...
+          } else {
+            tags$div("Plot not implemented yet")
+          }
+        }
+
+        nav_panel(g_def$label, plot_ui)
+      })
+
+      panels$"widths" <- c(2, 10)
+      output$graphs <- renderUI({
```
Gotfrid, line 289: suggestion: i'm strongly against having a render* function inside observe* function, because both are essentiall equally-powered observers.

instead, i suggest extracting `tlg_order_filt <- tlg_order()[tlg_order()$Selection, ]` into a separate reactive, and separate choice update and graphs rendering.

inst/shiny/modules/tlg_plot.R
```diff
@@ -0,0 +1,229 @@
...
+    num_pages <- reactive({
+      req(plot_list(), plots_per_page())
+      ceiling(length(plot_list()) / plots_per_page())
+    })
+
+    plots_per_page <- reactive({
+      if (is.null(input$plots_per_page)) return(NULL)
+      if (is.null(plot_list())) return(NULL)
+      if (input$plots_per_page == "All") {
+        isolate(length(plot_list()))
```
Gotfrid, line 101: nitpick: isolate here has no effect, because `plot_list()` is not isolated 2 lines above

inst/shiny/www/style.css
```diff
@@ -136,3 +136,70 @@
...
+  h1.tlg-group-label {
+    font-weight: bold;
+    font-size: 2rem;
+    border-bottom: thin solid #337ab7;
+    border-radius: 0 0 .25em 0;
+    margin-top: 0.75em;
+  }
+
+  .plot-options-container {
+     & > button {
```
Gotfrid, line 190:
```suggestion
    & > button {
```

Gotfrid (approved):
This work is so big that it probably deserves its own package. I was able to follow the instructions and create my own plot without any problems.

I left a few comments, but I wouldn't consider any of them blocking.

There is one problem that I found: following the test instructions, i was able to see all the plots and then i started configuring them. as a user, I select x-axis to be `studyid` and the app crashes.

Also, I'm not sure how x and y axes limits are supposed to work. I entered some values, but the don't seem to have any effect. Probably another ggplot -> plotly problem.

Gero1999:
Indeed @m-kolomanski great work! I can imagine this has taken a lot, but the results are very very impressive. Just to go part by part:

1) I talked with @Shaakon35 and he is keen in keeping the `ggplotly` functionality for the App visualization. However, we should keep in mind that at some point the App will allow the users to download a PDF document that includes all TLGs.

2) Regarding my review, I only noticed two things:

- [ ]  x and y limits not working
- [ ]  suggestion: default selections for numeric inputs (i.e, `x axis`)

3) Regarding @Gotfrid error with `STUDYID` in `x axis` is probably because this is a discrete variable with only 1 unique value. Maybe another future solution non-blocking for this PR could be:

m-kolomanski:
Hi @Gotfrid @Gero1999, thank you for the feedback!

I have implemented several changes:
- tried to be more verbose in the documentation;
- refactored the code to not nest rendering functions inside observers;
- moved complex function to `utils` file, added test;
- other minor refactoring;

To address the bigger issues:
- I have fixed the axis limits (it was a case of plotly ignoring options set in ggplot again).
- With the application crashing, I have added code to handle that error and display a message to the user, so that the app itself does not crash. I believe this _should_ be good enough and we _could_ expect the users to more-or-less know what they are doing and not select invalid options, but later on it would be useful to add a system to just prevent the user from selection options that crash the app.
- @Gero1999, I would not set any default widget values, since the plots have different limits and the options are applied globally. Any value specified in the widget will affect all plots. So I would suggest leaving the behavior on "automatic" by leaving the fields empty and letting the user decide if they want specific range.

Let me know what you think now.

Gero1999 (approved):
Great work, thanks!

PR #138, opened 2024-11-27
Bug: AUC start 0 (which also needs Feature: C0 imputation)

Gero1999:
## Issue
Closes #2 #60

## Description
After running NCA, AUC calculations always consider the discrete start (start at where first concentration sample was collected). This can indeed alter slightly the real values of the AUC. This is a consequence of 2 problems:

1) Doses times are being considered at the first measure, instead of at the real dose time
2) No imputed measure at dose time is consdiered

Change description.

## Definition of Done
C0 imputation option for the user specified in Setup > Data Selection

## How to test
- [x] At "NCA">"NCA Results", when C0 is imputed it is expected that start=AFRLT-ARRLT (being AFRLT-ARRLT= 0 in DOSNO=1)
- [x] At "NCA">"NCA Results", when C0 is not imputed it is expected that start={time of 1st observation} 
- [x] AUC values  (auclast, aucinf...) from start=AFRLT-ARRLT are calculated when C0 is imputed
- [x] All variables that should only be based on observed data (cmax, tmax...) are not considering C0 imputations
- [x] Test that PKNCA also deals well with predoses as duplicates for the NCA
- [x] Check slope selector plots and NCA results remain well after running NCA
- [x] Check if in general C0 imputation logic works well (NCA experts): 
if `C0 exists` ~ `C0=C0`
if `DOSNO=1 & IV bolus (ADOSEDUR=0, ROUTE=intravascular)` ~ `C0=0`
if `DOSNO>1 & not IV bolus` ~ `C0=predose`
if `IV bolus and monoexponential data` ~ `logslope`
if `IV bolus and not monoexponential data` ~ `C0=C1`

## Contributor checklist
- [x] Code passes lintr checks
- [x] Code passes all unit tests
- [x] New logic covered by unit tests
- [x] New logic is documented

## Notes to reviewer

Anything that the reviewer should know before tacking the pull request?

PR #136, opened 2024-11-22
Enhancement: Refactor Visualisation Tab

js3110:
## Issue

Closes #127 

## Description

With the introduction of the TLGs tab, the Outputs section now is no longer required for reporting but rather for enhanced data visualisation. We want to change the layout so it is better for visualisation.

### Definition of Done

- [x] rename outputs to visualisation tab
- [x] Modify layout to make for better exploratory analysis- bigger plots

## How to test
Check visualisation tab layout

## Contributor checklist
- [x] Code passes lintr checks
- [x] Code passes all unit tests
- [x] New logic covered by unit tests
- [x] New logic is documented

## Notes to reviewer

The plan is to reformat this section using bslib, which will then need to be implemented across the rest of the app. Refactoring of the Data section is being covered by #140

Issue #60, opened 2024-10-10
Bug: AUC calculations do not start at time 0


Gero1999:
### Description
After running NCA, AUC calculations always consider the discrete start (start at where first concentration sample was collected). This can indeed alter slightly the real values of the AUC.  

### To reproduce
NCA > Choose the analyte > Submit > Settings > Choose the Dose Number > Run NCA > NCA Results

### Expected behaviour
At "NCA">"NCA Results", it is expected that start=0 for all rows

billdenney:
Related to #125

NCA should be calculated from the time of dose rather than the first concentration (which is often predose, sometimes by a notable amount of time). Imputation of time 0 can be performed using one of the built-in methods from PKNCA described here: https://billdenney.github.io/pknca/articles/v08-data-imputation.html

Gero1999:
Thanks! Same thing is discussed in #2, I will leave the conversation going on there!

PR #140, opened 2024-12-03
Enhancement: Column Mapping Feature

### LINKED ISSUE #112: Enhancement: Customized column definition as inputs
### Description
We would like to offer the user better possibilities to define the column names that correspond with each column used by the app in the data. This will increase substantially the data flexbility and will also provide the user a better understanding of the App functioning.

### Definition of Done
- [ ] Provide select/selectize inputs for the user to define each of the important columns. Preferably if the layout is nice and with easy searchable options. One proposal is divided in these groups:

- Group identifiers. `Subject ID`, `Dose Number`, `Additional grouping`
- Concentration variables. `Analyte`, `Analyte value` 
- Dose variables. `Drug`, `Dose`, `Administration route`
- Time variables. `Time from 1st dose`, `Time from last dose`, `Dose duration`
- Unit variables. `Concentration unit`, `Time unit`

In theory, DOSNO can be guessed automatically with the time variables. But please, feel also free to add it or not depending on the layout :D 

- [ ] Hovering displays information about the column requirements or details. For example for administration route (`administration preferably stating either "intravascular" or "extravascular"`)

- [ ] Provide inputs for the user to define the columns for unit measures (e.g, AVALU) or otherwise that they can introduce directly a character value ("mg/L" also an option in the select apart from column names) **Note: Last thing is not that important**

Some ideas is either to pair these inputs to the select inputs that are numeric or to create an additional section with select inputs called: `Time unit`, `Concentration unit`

- [ ] Formulas for PKNCAdata and PKNCAdose need to include `other_groups`

js3110:
## Issue

Closes #112 

## Description

We would like to offer the user better possibilities to define the column names that correspond with each column used by the app in the data. This will increase substantially the data flexbility and will also provide the user a better understanding of the App functioning

## Definition of Done

- [x] Provide select/selectize inputs for the user to define each of the important columns. Preferably if the layout is nice and with easy searchable options. One proposal is divided in these groups:

- Group identifiers. `Subject ID`, `Dose Number`, `Additional grouping`
- Concentration variables. `Analyte`, `Analyte value` 
- Dose variables. `Drug`, `Dose`, `Administration route`
- Time variables. `Time from 1st dose`, `Time from last dose`, `Dose duration`
- Unit variables. `Concentration unit`, `Time unit`

- [x] Hovering displays information about the column requirements or details. For example for administration route (`administration preferably stating either "intravascular" or "extravascular"`)

- [x] Provide inputs for the user to define the columns for unit measures (e.g, AVALU)

-  Formulas for PKNCAdata and PKNCAdose need to include `grouping_variables`

## How to test

Data Section -> Column mapping
Try using dataset with renamed columns to see that it still works.

## Contributor checklist
- [x] New logic is documented


js3110:
@Gero1999 I switched to bslib tooltips because the shinyBS ones weren't working. Basically the same thing though. I've left the tooltip messages pretty vague for now, only the grouping and adosedur ones have more info. Lmk if you want any of them to change. I figure once we add labels from #100 then some of the tooltips won't be needed or we can adjust them as needed.

Also not sure about grouping_variables as part of pkconc and pkdose? can also be adjusted later, after you've merged #138

Gero1999 (changes requested):
Nice @js3110! I agree, once #138 is merged can be clever to considered, is just about adding the grouping_varaibles to both formulas. 

Some feedback regarding optional details that could be added:

- [x] I would prefer that the input calls are static and created in the UI (i.e, `selectInput(...))`). Otherwise it is very likely to produce issues, I guess that this could be the reason of why `shinyBS` was not working for you. 
- [ ] For our users label ID would be preferred  to be explanatory names rather than CDISC variables.
- [x] Hovering can add information about the column expected format, i.e, class expected such as numeric or categorical.
- [x] When the user press `Submit Mapping` I would expect something to happen. Either a message notification like "Data mapped" or a Tab displacement
- [x] `grouping variables` could be better called something like `other grouping variables` so the user do not get the wrong perception and thinks the rest are not used and puts them two times. it could also be good to update the choices to elude them to put the same column as in `ANALYTE`, `DOSNO`...

js3110:
@Gero1999  have updated. its a lot more bulky now but the ui is static. the label IDs will be added when we do #100 so I have left it how it was. Not entirely sure what else to put in the tooltips but feel free to make suggestions if you think it needs changing. For grouping variables I have added a clearer description so the user knows what it is for, and if the columns are already named correctly the user cannot select the other columns anyway.

Gero1999:
Good job! I just detected some small things. Let me know if I can help and you agree with them:

- Main task incomplete: Add in formulas for PKNCAdata and PKNCAdose need to include grouping_variables
- App crashes when using the same column name for different inputs! We should either disable this option or prevent the crash wiht a notification!
-  ADOSEDUR should be simply set to 0 when a study is not infusion. I am not sure if the NA is actually currently doing anything (the column is not ignored, simply not renamed).

js3110:
> Good job! I just detected some small things. Let me know if I can help and you agree with them:
> 
> * Main task incomplete: Add in formulas for PKNCAdata and PKNCAdose need to include grouping_variables
> * App crashes when using the same column name for different inputs! We should either disable this option or prevent the crash wiht a notification!
> * ADOSEDUR should be simply set to 0 when a study is not infusion. I am not sure if the NA is actually currently doing anything (the column is not ignored, simply not renamed).

@Gero1999  Ok it should be ready now, as agreed I have left grouping_variables for nca to you for after #60 has been completed. And for the input labels, again this will be included in #100 .

Gero1999:
hey @js3110 as discussed in chat we will accelarate one of the issues into this one directly. Just add an option for the user to apart from choosing column names in the unit inputs also standard concentration variables. Here are some helpful instructions:

- [ ] Take the values directly from PKNCA options if possible, don't hardcode unless you don't see other alternative
- [ ] Make sure to separate them by order and definition in the input. If you use a list object for the select input it will understand how to seggregate the results (`dataset columns` and `measure units`)
- [ ] When a `measure units` is selected, the column is automatically created with only that unit

js3110:
@m-kolomanski @Gotfrid Hi, could one of you please check the code on this and review. There is a linter check saying it has too much cyclomatic complexity, so suggestions on how to fix this would be appreciated. Thanks!

Gotfrid:
Hi @js3110 , I will be able to do the review earliest tomorrow. Cyclomatic complexity issue means that you have too many if-else branches in a function. In case of Shiny it pretty much always means shiny modules are required to extract pieces of server.

js3110:
> Hi @js3110 , I will be able to do the review earliest tomorrow. Cyclomatic complexity issue means that you have too many if-else branches in a function. In case of Shiny it pretty much always means shiny modules are required to extract pieces of server.

ok thanks! I will created a nested module today so its ready by tomorrow.

Gero1999 (approved):
lgtm

inst/shiny/modules/tab_data.R
```diff
@@ -1,52 +1,62 @@
...
-        buttonLabel = list(icon("folder"), "Upload File..."),
-        accept = c(".csv", ".rds")
+  navset_pill(
+    id = ns("data_navset"),
+    nav_panel("Raw Data Upload",
+      card(
+        "Upload your PK dataset in .csv format",
+        # Local upload option
+        fileInput(
+                  ns("local_upload"),
```
Gotfrid, line 19: nitpick: indentation

inst/shiny/modules/column_mapping.R
```diff
@@ -0,0 +1,285 @@
...
+      # Reorder columns based on the desired order
+      ordered_data <- dataset[, c(desired_order, setdiff(names(dataset), desired_order))] %>%
+        mutate(TIME = ifelse(DOSNO == 1, AFRLT, ARRLT)) #TODO: Remove this after auc0 merged
+      processed_data(ordered_data)
+
+      # Execute the callback function to change the tab
+      on_submit()
+    })
+
+    return(list(
```
Gotfrid, line 280: nitpick: no explicit return + empty line in the end of the file

inst/shiny/modules/column_mapping.R
```diff
@@ -0,0 +1,285 @@
...
+    }))
+
+    # Define the desired column order
+    desired_order <- c("STUDYID", "USUBJID", "ANALYTE",
+                       "PCSPEC", "AVAL", "AVALU", "AFRLT",
+                       "ARRLT", "NRRLT", "NFRLT", "RRLTU",
+                       "ROUTE", "DOSEA", "DOSEU", "DOSNO")
+
+    # Populate the static inputs with column names
+    observe({
```
Gotfrid, line 199: issue: prefer `observeEvent` over `observe`, i.e.

```r
observeEvent(data(), { ... })
```

inst/shiny/modules/column_mapping.R
```diff
@@ -0,0 +1,285 @@
...
+      }
+      if (input$select_DOSEU %in% manual_units$dose) {
+        dataset$DOSEU <- input$select_DOSEU
+      }
+      if (input$select_RRLTU %in% manual_units$time) {
+        dataset$RRLTU <- input$select_RRLTU
+      }
+
+      # Reorder columns based on the desired order
+      ordered_data <- dataset[, c(desired_order, setdiff(names(dataset), desired_order))] %>%
```
Gotfrid, line 272: suggestion:

```r
      dataset %>%
        relocate(all_of(desired_order)) %>%
        mutate(TIME = ifelse(DOSNO == 1, AFRLT, ARRLT)) %>%
        processed_data()
```

Gotfrid (approved):
I don't want to be a blocker given how urgent it is, and how rare my involvement is, so I'm approving this PR :)

However, I want to stress that I don't think it's a good idea to migrate to `bslib` in the scope of this PR. This must be a very careful transition. The app in its current state doesn't look good, especially the slope selector in NCA section (see a screenshot).

Other than that I left a few non-blocking comments.

As for the testing, I can see that the reactive parts are not being tested at all in the app right now. So I believe that it is a better idea to have a separate pull request with an example, that could be reused for testing other modules.

Nevertheless, this PR lacks unit testing (thank you @js3110 for reaching out!) because the functionality is not split into "units", i.e. there is a lot of logic inside the reactives - that could be extracted into functions with tests. This is what I would suggest instead of introducing session mocks and the likes.

m-kolomanski:
> However, I want to stress that I don't think it's a good idea to migrate to `bslib` in the scope of this PR. This must be a very careful transition. The app in its current state doesn't look good, especially the slope selector in NCA section (see a screenshot) (cc @m-kolomanski).

Hey @Gotfrid, thanks for the cc, just to be clear and put it out there - I have encouraged @js3110 to start migrating to `bslib` bit by bit, seems like migrating bits, modules at a time is much more manageable than going from top to bottom. Are there any technical adversities of doing that, or is it mostly about the interface parts clashing visually?

inst/shiny/modules/tab_data.R
js3110: the indentation was done to pass the lintr- this is what it made me change it to so I'm hesitant to change it

js3110:
> > However, I want to stress that I don't think it's a good idea to migrate to `bslib` in the scope of this PR. This must be a very careful transition. The app in its current state doesn't look good, especially the slope selector in NCA section (see a screenshot) (cc @m-kolomanski).
> 
> Hey @Gotfrid, thanks for the cc, just to be clear and put it out there - I have encouraged @js3110 to start migrating to `bslib` bit by bit, seems like migrating bits, modules at a time is much more manageable than going from top to bottom. Are there any technical adversities of doing that, or is it mostly about the interface parts clashing visually?

just to add to this I also have PR #136 which migrates the outputs tab to bslib as well, so then all thats left is the nca and tlgs sections.

Gotfrid:
Migrating to bslib is a pain, because for bslib to take effect - one needs to change the outermost shiny "page" function to be from bslib. This will now load a different set of CSS/JS assets (related to bootstrap 5), which will likely destroy most of the styling/layout of the app.

I'm glad that this was a team decision. If it's acknowledged, I'm perfectly fine with it.

Issue #144, opened 2024-12-16
Refactor: Migrate the NCA tab to shiny modules and bslib


js3110:
### Description
Since incorporation of bslib ui layout to data and outputs section, nca section formatting is bad. Needs to be updated so it looks better.

### Definition of Done:

- [x] NCA tab refactored to be contained within shiny modules
- [x] NCA tabs reformatted to fit bslib
- [x] All sections correctly formatted in an easy to use and easy to read way.

m-kolomanski:
In addition, we should finally modularize this tab. #146 adds some changes to `nca.R` file, so probably this issue should follow. I think part of the issues with code logic in the above PR stems from the fact that this part of the application still lives free instead of being encapsulated.

js3110:
@m-kolomanski completely agree, and I was already thinking of modularising the tab once #109 and #146 are complete. We can do it as part of this issue, or separately within #63 , what do you think?

m-kolomanski:
Within this issue, at this point #63 is an "epic" issue, existing to remind of a bigger goal rather than a task on its own right.

I am renaming this issue to reflect the actual wanted changes and we can work within that.

PR #109, opened 2024-11-06
Bug: AUC start 0 (which also needs Feature: C0 imputation)

Gero1999:
## Issue
Closes #2 #60

## Description
After running NCA, AUC calculations always consider the discrete start (start at where first concentration sample was collected). This can indeed alter slightly the real values of the AUC. This is a consequence of 2 problems:

1) Doses times are being considered at the first measure, instead of at the real dose time
2) No imputed measure at dose time is consdiered

Change description.

## Definition of Done
C0 imputation option for the user specified in Setup > Data Selection

## How to test
- [x] At "NCA">"NCA Results", when C0 is imputed it is expected that start=AFRLT-ARRLT (being AFRLT-ARRLT= 0 in DOSNO=1)
- [x] At "NCA">"NCA Results", when C0 is not imputed it is expected that start={time of 1st observation} 
- [ ] AUC values  (auclast, aucinf...) from start=AFRLT-ARRLT are calculated when C0 is imputed
- [ ] All variables that should only be based on observed data (cmax, tmax...) are not considering C0 imputations
- [ ] Test that PKNCA also deals well with predoses as duplicates

## Contributor checklist
- [ ] Code passes lintr checks
- [ ] Code passes all unit tests
- [ ] New logic covered by unit tests
- [ ] New logic is documented

## Notes to reviewer

Anything that the reviewer should know before tacking the pull request?

PR #146, opened 2024-12-17
Enhancement: Units edition results

### LINKED ISSUE #113: Enhancement: allow users to modify unit values of parameter results
### Description
Sometimes user might be interested in changing the standard format of units provided by the app. In one that this may happen frequently is for example clearance (cl.obs), where current units always make very small values (~0) that can confuse when the rounding is used in the summary statistics. 

Preferably as a table in the **setup** section similar to this:
Image

with possible additional option of changing units after NCA in **results** section

### Definition of Done
- [ ] allow user to customize in results the units of measure for each parameter
- [ ] scale appropriately parameter values based on new unit measures imputed

Gero1999:
## Issue
Users would like to be able to choose the units in which the results parameters are displayed.
Closes #113 

## Description
Sometimes user might be interested in changing the standard format of units provided by the app. In one that this may happen frequently is for example clearance (cl.obs), where current units always make very small values (~0) that can confuse when the rounding is used in the summary statistics.

## Definition of Done
- [x]  Allow user to customize in results the units of measure for each parameter
- [x]  Scale appropriately parameter values based on new unit measures imputed

## How to test

## Contributor checklist
- [x] System to change units in parameters works well before NCA
- [x] System to change units in parameters works well after NCA
- [x] Outputs post NCA still work fine and adapt well to new units (i.e, boxplots)

- [x] New logic is documented

## Notes to reviewer


Gero1999:
Hello team @m-kolomanski @Gotfrid @js3110 ! 

I have a small technical problem with this `units` package. This is a package that I used for the function `transform_unit` that automatically calculates conversions. This package is also used in PKNCA and @billdenney has also participated in its development. 

When the package is installed through non-Windows OS, the installation through CRAN with `install.packages()` won't work unless  a terminal command is run before to install the udunits2 object, as indicated in their [README](https://github.com/r-quantities/units) . The problem is that the testing on the remote GitHub-workflow might require this, as it is a Linux system. 

There are many potential solutions to this, as the only affected things are the `transform_unit` function and its testthat testing. However I am not sure what is the best approach to solve it and I would like your input if possible. Here are my current ideas:

 1) Add the package `units` to "Suggests" in the namespace, and only run `transform_unit` when possible with `if (!requireNamespace("units", , quietly = TRUE)` (in other cases return NA, which does not conflict with the App, as the user will be asked to manually populate the conversions).
 
 2) Customize somehow the installation for each OS. I know that with R we can request the `Sys.info()` to obtain all info we need (`sysname` + perhaps `version`) to know what action to take before doing the `install.packages`. However this procedure concerns me, as I am not sure if we require additional permits from the user to do this installation method. 
 
 3) Do a function that is independent from the `units` package that is only intended for the physical units we use (concentration, time, volume, mass). I indeed did something of this in the past using regex, I would just need to refine it and create the tests. 
 
 Let me know what you think of them or if you could come up with a better solution and I will try to implement it!

m-kolomanski:
Hi @Gero1999,

use whatever dependencies you need to use - there is no need to re-invent the wheel if ready solutions are available, and we are not the first to deal with system dependencies, so do not worry :)

In terms of installing the package itself, if installing a dependency (in that case `units`) fails due to missing system packages, the user is prompted to install those and this is alright - something that I deal with frequently as Ubuntu user. We only need to make sure that the package is present in the DESCRIPTION file. I do not think there are any other concerns, @Gotfrid?

In terms of github actions / workflows, we can install the dependency on the machine that runs the checks, I will get on that.

billdenney:
I agree with @m-kolomanski, adding the dependency is the best path. Writing your own unit handling will likely be error-prone. (I've done it a few times in the past, and it always has lots of bugs the first time around.) The units package has the side-benefit that all standard unit conversions should be handled for you.

Gotfrid:
Agreed with everything mentioned above regarding the `units` package. It is not uncommon to require system dependencies installed in the environment (think `sf` package, or any DBMS package).
Comment from my end: once `units` is added to the dependencies, we can recommend installation via `pak` - because it handles system requirements as well (99% of the time).

inst/shiny/modules/units_table.R
```diff
@@ -0,0 +1,199 @@
...
+
+        new_res_nca(res_nca)
+      }
+      # Close the module window once all saving actions are done
+      removeModal()
+
+    })
+
+    return(reactive(new_res_nca()))
+  })
```
m-kolomanski, line 198: suggestion: This also could be wrapped into single `reactive()` expression and returned directly. You can use `bindEvent` to bind a reactive expression to a single specific event, instead of recalculating it every time some dependency change (it is basically like wrapping it in `isolate()`.

```R
reactive({
  # code for pepraring the results
  removeModal()
  res_nca
})
```

inst/shiny/modules/units_table.R
m-kolomanski, line 1: issue: There are a lot of side-effect in this module and they seem to not be consistent. You are modifying `mydata()` object directly in this module, but instead of also modifying `res_nca()` directly you return it as a reactive, and then assign the value in `anca.R` file (Line 37). There is nothing inherently wrong with mutating the state of the object directly, but it gets confusing as the complexity grows and might be hard to debug later on. Please make sure all side-effects of the module are desired and documented.

inst/shiny/tabs/nca.R
```diff
@@ -293,6 +301,11 @@ observeEvent(input$submit_analyte, priority = -1, {
   )
 })
 
+# If the user requests it allows to change the units of the parameters using a modal message table
+units_table_server("units_table_preNCA", mydata, reactive(NULL), params_to_calculate)
+res_nca2 <- units_table_server("units_table_postNCA", mydata, res_nca, params_to_calculate)
+observeEvent(res_nca2(), res_nca(res_nca2()))
```
m-kolomanski, line 307: question: So if I understand that correctly, we are using `units_table_preNCA` module to modify only `mydata` object, but use `units_table_postNCA` to modify both `mydata` and `res_nca`? This is a bit confusing and relates to my previous comment about side-effects. Seems like a single instance of the module should be able to handle both cases (and just return NULL if `res_nca` is not available).

inst/shiny/tabs/nca.R
```diff
@@ -371,8 +384,8 @@ observeEvent(input$nca, {
 })
 
 # run the nca upon button click
-
-res_nca <- eventReactive(rv$trigger, {
+res_nca <- reactiveVal(NULL)
```
m-kolomanski, line 387: issue: `res_nca` is referenced on line 307, but only defined on line 387. This does not cause any technical problems because of the even structure of the program, but is bad practice nevertheless - this should be higher. Considering the importance of that data, I would consider putting this definition on top of the module.

R/transform_unit.R
```diff
@@ -0,0 +1,31 @@
...
+      conversion <- units::set_units(
+        units::set_units(1, initial_unit, mode = "standard"),
+        target_unit, mode = "standard"
+      )
+      as.numeric(conversion)
+    }, error = function(e) {
+      if (initial_unit == target_unit) {
+        1
+      } else {
+        NA
```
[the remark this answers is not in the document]
Gero1999: The section in the app where the function is applied will notify the users if any NA is set as a conversion factor. It will ask them to either introduce themselves the appropriate conversion factor or to change the target_unit to one that is accepted by `units` library

inst/shiny/modules/units_table.R
```diff
@@ -0,0 +1,199 @@
...
+            paste0(
+              "function(row, data, index) {",
+              "  var paramsToCalculate = ",
+              paste0("['", paste(params_to_calculate(), collapse = "','"), "']"),
+              ";",
+              "  if (paramsToCalculate.indexOf(data[1]) === -1) {",
+              "    $(row).hide();",
+              "  }",
+              "}"
+            )
```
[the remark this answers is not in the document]
m-kolomanski: After looking at it again, the glue syntax might cause problems with brackets in the code, maybe simply pasting is a better solution:
```R
paste0("
  function() {
    var paramsToCalculate = ", parameters_array_string, ";
    // ...
")
```

Gero1999:
I ended up merging instead of rebasing because in practice seemed less messy to me (I needed way too many commtits when rebasing). Hope is fine! Regarding some aspects of the module simplicity @m-kolomanski, I tried to document better the changes that occur:

1) I did not manage to make them explicit, although now they are implicit for both objects
2) I did not maange to only make 1 server call, because the UI has two input definitions and seems that I cannot repeat the same one two times!

But please let me know any concern with these solutions

R/get_conversion_factor.R
```diff
@@ -0,0 +1,28 @@
+#' Transform Units
+#'
+#' This function transforms a value from an initial unit to a target unit.
+#'
+#' @param initial_unit A character string representing the initial unit.
+#' @param target_unit A character string representing the target unit.
+#' @return A numeric value for the conversion factor from the initial to the target unit,
```
m-kolomanski, line 7: nitpick: 
```suggestion
#' @returns A numeric value for the conversion factor from the initial to the target unit,
```

tests/testthat/test-transform_unit.R
m-kolomanski, line 1: issue: Please utilize `describe` and `it` in the testing suite

inst/shiny/modules/units_table.R
```diff
@@ -0,0 +1,205 @@
...
+    observeEvent(input$open_units_table, {
+
+      # Keep in a variable all analytes available
+      analyte_choices <- unique(mydata()$conc$data[[mydata()$conc$columns$groups$group_analyte]])
+
+      # Show the modal message with the units table and an analyte selector
+      showModal(modalDialog(
+        title = tagList(
+          span("Units of NCA parameter results"),
+          tags$button(
```
m-kolomanski, line 35: issue: this button does not seem to do anything.

inst/shiny/modules/units_table.R
```diff
@@ -0,0 +1,205 @@
...
+    actionButton(
+      ns("open_units_table"),
+      icon = shiny::icon("scale-balanced"),
+      label = "Parameter Units",
+      disabled = TRUE
+    )
+  )
+}
+
+units_table_server <- function(id, mydata, res_nca = reactive(NULL), params_to_calculate) {
```
m-kolomanski, line 14: issue:
```suggestion
units_table_server <- function(id, mydata, res_nca = reactiveVal(NULL), params_to_calculate) {
```
as the module expects `reactiveVal` as opposed to `reactive`

inst/shiny/modules/units_table.R
m-kolomanski, line 1: issue, style: Please be consistent with namespacing the function calls, there is a mix.

m-kolomanski (changes requested):
Hey, it looks quite nice, but there are some minor issues to iron out.

The data flow is still a bit confusing, but I think we should fix that in the scope #144 , when modularizing the whole NCA tab.

m-kolomanski (approved):
Hey, to me, the code looks good, thank you!

Since this PR touches the `nca.R` file directly, and we wish to rework it ASAP, there are some minor issues / suggestions that we should create a separate issue for and address within a new PR:

- Bug: the application crashes when a text value is provided as conversion factor;
- Suggestion: If conversion factor returns NA, might be worth displaying something else in the table, as opposed to an empty cell (like **Error* or something, to explicitly inform user that there is something wrong).
- Suggestion: Disable the ability to select rows in unit table, as it does not provide any functionality.

js3110:
> Hi @Gero1999 It looks really good! I love the layout especially.
> 
> However I have noticed that the nca results are missing a lot of parameters (such as CL clearance) , which is important as this parameter is the main reason we are trying to implement unit changes (for pre clinical)
> 
> I have had a check and the parameters are also missing in main, but not in our rsconnect version so it must have slipped through in the #109 merge. if you can fix it in this issue then great, and if not then lets make a separate issue for this!

Looks to me as if this is coming from the interval creation, as if you look at mydata() it is set as FALSE for most of the parameters except for the ones included in the format_pkncadata_intervals function. We want all the parameters to be available, or the user should be able to choose. Currently neither are possible.

js3110:
I've created issue #168 as it is separate from this issue.

js3110 (approved):
LGTM

Gero1999:
I implemented your requested changes @m-kolomanski

- Bug fixed -> conversion factor when given a text won't crush and gives a notification of the expected format
- Suggestion -> If conversion factor returns NA, an additional notification is sent to the user at the moment of the action
- Suggestion -> Instead of row selection they will have single cell selection. That way they have a better sense that the table is interactable and how it works!

As mentioned @js3110  I will use a different branch to make the input for NCA parameters. As soon as we merge this one to main I will create it and start working on it.

js3110:
@Gero1999  LGTM! I just added cl.obs into the params as a temporary solution, so that the DPLs can use it for testing before #168 is finished.

js3110 (approved):

Gero1999:
Hey @js3110 I added  `cl.obs`. Your commit did not work because it was just for the default function arguments

inst/shiny/modules/units_table.R
```diff
@@ -121,17 +121,36 @@ units_table_server <- function(id, mydata, res_nca = reactiveVal(NULL)) {
 
     # Accept user modifications in the modal units table
     observeEvent(input$modal_units_table_cell_edit, {
+
       info <- input$modal_units_table_cell_edit
       modal_units_table <- modal_units_table()
+      
+      # If the edited cell is in the 'Conversion Factor' only accept numeric values
+      if (names(modal_units_table)[info$col + 1] == "Conversion Factor" && !is.numeric(info$value)) {
```
m-kolomanski: nitpick: might be work assigning `info$col + 1` to a variable, since it is repeated a bunch and then using that throughout.

inst/shiny/modules/units_table.R
```diff
@@ -0,0 +1,234 @@
...
+        # If the modification lead to an unexpected conversion factor notify the user
+        if (is.na(conversion_factor_value)) {
+          showNotification(
+            paste0(
+              "Unrecognised conversion: ",  def_unit, " > ", cust_unit,
+              ". Either make sure both units are defined in the UNIDATA 
+              library or impute a conversion factor yourself"
+            ),
+            type = "error",
+            duration = 5
```
m-kolomanski, line 163: nitpick: I would consider extending the duration and / or instead adding a close button, it might be a bit hard to read the whole error in 5 seconds.

m-kolomanski (approved):
Hey, looks good, thank you for implementing the suggestions!

I am approving the PR, but leaving two minor, non-blocking nitpicks for consideration.

PR #147, opened 2024-12-18
Enhancement: add labels to ADNCA

### LINKED ISSUE #100: Enhancement: Add labels to ADNCA

js3110:
## Issue

Closes #100 

## Description

When the user loads ADNCA, sometimes there could be some column labels, sometimes not (if csv, there are no labels, if rdata there may be labels).
What we need to do is to populate all the labels based on the ADNCA guidelines.
If the variable is in the list: populate the label, if not, populate by the name of the variables:
e.g. ADNCA$ASIAN does not have labels, then populate:
attr(ADNCA[["ASIAN"]], "label") = "ASIAN"

## Definition of Done

- [x] after loading a csv or rdata, make sure all variables do have a label
- [x] make sure that if a variable is a factor with levels, the levels are kept! please use as_factor_preserve_label  function
- [x] Make sure all column headers in the app show the user the CDISC name and also the label. 

## How to test

Data -> Map columns -> Submit -> Hover over column headers

## Contributor checklist
- [x] New logic is documented


js3110:
@Gero1999 @Shaakon35 I have added tooltips now, can you let me know if you think this looks good or not, and if yes I will then add it to the rest of the data tables using these columns.

inst/shiny/modules/column_mapping.R
```diff
@@ -268,10 +268,16 @@ column_mapping_server <- function(id, data, manual_units, on_submit) {
...
       # Reorder columns based on the desired order
-      dataset %>%
+      dataset <- dataset %>%
         relocate(all_of(desired_order)) %>%
-        mutate(TIME = ifelse(DOSNO == 1, AFRLT, ARRLT)) %>% #TODO: Remove this after AUC0 merged
-        processed_data()
+        mutate(TIME = ifelse(DOSNO == 1, AFRLT, ARRLT))#TODO: Remove this after AUC0 merged
+
+      # Apply labels to the dataset
+      message("Calling apply_labels function")
```
m-kolomanski, line 276: issue: Please use `log_trace` instead of `message`. Furthermore, I would argue it is not worth keeping this log anyway.

tests/testthat/test-apply_labels.R
```diff
@@ -0,0 +1,21 @@
+describe("apply labels works correctly", {
```
m-kolomanski, line 1: nitpick:
```suggestion
describe("apply_labels", {
```

tests/testthat/test-generate_coldefs.R
```diff
@@ -0,0 +1,21 @@
+describe("column definitions generated correctly", {
```
m-kolomanski, line 1: nitpick:
```suggestion
describe("generate_col_defs", {
```

inst/shiny/www/data/labels_app.csv
m-kolomanski, line 1: nitpick: I would suggest more verbose name, like `ADNCA_labels.csv` or something similar, so other developers know exactly what is in this file.

R/label_operators.R
```diff
@@ -1,3 +1,48 @@
...
+#'   # Example usage:
+#'   data <- data.frame(USUBJID = c(1, 2, 3), AVAL = c(4, 5, 6))
+#'   data <- apply_labels(data)
+#'   print(attr(data$A, "label"))
+#' }
+#'
+#' @export
+apply_labels <- function(data) {
+
+  labels_app <- read.csv(system.file("shiny/www/data/labels_app.csv", package = "aNCA"))
```
m-kolomanski, line 23: issue: If possible, we should avoid loading files inside of functions - this means the file will be loaded everytime the function is called (which should not be a lot in the app, nevertheless this is bad practice).

I would suggest loading the specification file on app-level, and then taking the data frame as an argument. This will also make the function more versatile - we could load different specifications with it, theoretically.

R/label_operators.R
```diff
@@ -1,3 +1,48 @@
...
+
+  # Create the label_ADNCA named vector from labels_app
+  label_adnca <- setNames(labels_app$Label, labels_app$Variable)
+
+  log_info("Applying labels to data")
```
m-kolomanski, line 28: issue: `message()` would be more appropriate within a function, if this needs to be kept at all. Usually we only want functions to print something when something of note is going on (warning, error or some important information).

R/label_operators.R
```diff
@@ -1,3 +1,48 @@
...
+      attr(data[[col]], "label") <- col
+    }
+
+    # Check if the column is a factor and keep the levels order
+    if (is.factor(data[[col]])) {
+      data[[col]] <- as_factor_preserve_label(data[[col]])
+    }
+  }
+
+  return(data)
```
m-kolomanski, line 43: nitpick:
```suggestion
  data
```

R/generate_coldefs.R
```diff
@@ -0,0 +1,47 @@
...
+          title = label
+        )
+      )
+    } else {
+      colDef(name = col)
+    }
+  })
+  names(col_defs) <- names(data)
+
+  return(col_defs)
```
m-kolomanski, line 46: suggestion: You can pipe the results of `lapply` into `setNames`. Then the explicit return is not needed.
```suggestion
  }) |>
    setNames(names(data))
```

R/generate_coldefs.R
```diff
@@ -0,0 +1,47 @@
...
+#'     AVAL = c(4, 5, 6)
+#'   )
+#'   attr(data$USUBJID, "label") <- "Unique Subject Identifier"
+#'   attr(data$AVAL, "label") <- "Analysis Value"
+#'   col_defs <- generate_col_defs(data)
+#' }
+#'
+#' @export
+generate_col_defs <- function(data) {
+
```
m-kolomanski, line 23: nitpick: empty line not really needed
```suggestion
```

R/generate_coldefs.R
m-kolomanski, line 1: issue: I would argue this file belongs to `/inst/shiny` and not `/R`, since it is responsible solely for generating labels / tooltips for UI elements, and not for any inherent logic behind the data.

Generally,  when I decide where the file goes, I think _if I would use this as a **package**, would I ever want to call this function from the console?_ If yes, `/R`. If not, `/inst/shiny`.

Also, name of the file is slightly different than the function name, which might be confusing.

If automatic sourcing of the function in tests breaks after moving the file, you will need to find the path to the file and source it directly. I would suggest `rprojroot` for that, so for example:
```R
root <- rprojroot::find_root(rprojroot::is_r_package)
source(file.path(root, "/inst/shiny/functions/generate_col_defs.R`))
```

m-kolomanski (changes requested):
Hey,

good job overall, thank you for taking the time to add unit tests. I did find some issues to iron out.

In addition, I would like to share some suggestions in regards to using logs, as some of my comments regard exactly that.
1. `log_*` calls should live in the application itself, while for functions `message()`, `warning()`, `stop()` should be used.
2. For functions, in general we only want to log important/notable evens, warnings, errors etc. to be logged. It would get very annoying very quickly if all functions printed `message("I am called!")` every time they are used, so when everything goes to plan functions should be silent.
3. For application, usually we are interested in tracing events and user actions, so logs at the beginning of reactives, observers and renderers might be useful (but not always are). If an event is complex or branching, some logs informing us which path the application take might be useful. For user actions, some info regarding the exact input the user provided might be useful, like what options were specified for filtering or which TLGs the user picked to generate. I would generally stay away from logs like `Function X is called`, unless it is an exceptionally important part OR has to do with the branching (like function X is only called in specific situations). Otherwise the logs might get polluted really fast.
4. Using `print` / `message` or `log_*` inside functions and application for debugging, to trace what exactly is happening, is perfectly fine, but such code should be removed before merging.

R/generate_coldefs.R
js3110: for the other functions in the shiny folder they have just been sourced in global.R, can I do that for this one too? or is there a more efficient way

m-kolomanski: For the application, yes, absolutely - should just be sourced in `global.R`.

I am talking about the `test-generate_coldefs.R` file. The tests might not import the function automatically when run, hence why the provided snipped might be needed in the test file.

js3110: ah I see, thanks! Yes you were right, it did not import automatically.

js3110:
Hi @m-kolomanski, `devtools::test()` works fine with the generate_col_defs.R function, however `devtools::check()` is coming up with errors `Error ('test-generate_col_defs.R:1:1'): (code run outside of `test_that()`)` 
Seems to be something to do with the root but I can't figure out how to fix it. Do you have any ideas? Thanks!

inst/shiny/www/data/adnca_labels.csv
m-kolomanski, line 1: issue: 
Data files should not be located in the `/www` folder. Files in `/www` are automatically exposed to the browser, which impacts a) performance and b) security. 

When the application is run, you can go to `http://127.0.0.1:PORT/data/adnca_labels.csv` to directly download the file - which is rarely intended behavior. With the package being open-source, the security risks are not a concern, but this might have performance implication and is generally an anti-pattern.

Please use `/shiny/data` folder to store this file.

tests/testthat/test-generate_col_defs.R
```diff
@@ -0,0 +1,23 @@
...
+    expect_equal(col_defs$AVAL$header$attribs$title, "Analysis Value")
+  })
+  it("returns a list of column definitions", {
+    data <- data.frame(
+      USUBJID = c(1, 2, 3),
+      AVAL = c(4, 5, 6)
+    )
+    col_defs <- generate_col_defs(data)
+    expect_type(col_defs, "list")
+  })
```
m-kolomanski, line 22: nitpick: I do not think the separate test for checking return type is needed, I think it can be enclosed in the first test.

```suggestion
  it("generates a list of column definitions with labels", {
    data <- data.frame(
      USUBJID = c(1, 2, 3),
      AVAL = c(4, 5, 6)
    )
    attr(data$USUBJID, "label") <- "Unique Subject Identifier"
    attr(data$AVAL, "label") <- "Analysis Value"
    col_defs <- generate_col_defs(data)
    expect_type(col_defs, "list")
    expect_equal(col_defs$USUBJID$header$attribs$title, "Unique Subject Identifier")
    expect_equal(col_defs$AVAL$header$attribs$title, "Analysis Value")
  })
```

R/label_operators.R
```diff
@@ -1,3 +1,48 @@
+#' Apply Labels to a dataset
+#'
+#' This function adds "label" attributes to all columns in a dataset
+#' Using the labels_app.csv file which contains common labels for ADPP
+#' and ADPC column names
+#'
+#' @param data The dataset to which labels will be applied.
```
m-kolomanski, line 7: issue: docs are missing `labels_file` parameter

m-kolomanski (changes requested):
Hey @js3110,

I have pushed some changes to fix the R CMD check. This is my bad - the `rprojroot::is_r_package` fails, since the R CMD check test the installed package, which changes the general folder/file structure of the package. Sourcing the file using `system.file()` does the trick.

Thanks for addressing the issues I have mentioned before. I performed a re-review and added some more comments to the PR.

js3110:
Thank you @m-kolomanski !
I have made those changes, hopefully it is all good now :)
Happy new year!

m-kolomanski (changes requested):
Thank you and happy New Year to you too!

I am afraid the app crashes though, seems like the label `.csv` file is missing altogether 😅

js3110:
> Thank you and happy New Year to you too!
> 
> I am afraid the app crashes though, seems like the label `.csv` file is missing altogether 😅

Its working fine on my end so that's strange... the .csv file should now be in the inst/data folder

m-kolomanski:
Seems like the file is not committed / pushed to the repo?

js3110:
> Seems like the file is not committed / pushed to the repo?

i found the issue, the data folder is in git ignore so it wasn't updating. Fixed now!

m-kolomanski:
I am afraid the file is still not present in the repository, you probably need to add it outright (specify the direct path in `git add ` call) and then push.

js3110:
> I am afraid the file is still not present in the repository, you probably need to add it outright (specify the direct path in `git add ` call) and then push.

ok.... now I think it is done !

m-kolomanski (approved):
Yep, works nicely, thank you for the changes and all your hard work 👍

Gero1999 (approved):
lgtm, well done! 👌🏽

PR #152, opened 2025-01-03
Bug: Disallow user to check NCA / Outputs tabs without mapping data

### LINKED ISSUE #143: Bug: Disallow user to check NCA / Outputs tabs without mapping data
### Description
When user has not mapped the data (define column names and accepted) it should not be able to go through other tabs (NCA or Outputs) as it can be confusing for them to see errror messages afterwards. Instead, those tabs should remain disable and perhpaps even show a notification

### To reproduce
Intiiate the app and click NCA or Outputs tab. In the second one you won't find results but instead just error messages.

### Expected behaviour
Disability to go to Ouptuts or NCA without mapping.
Notification with instructions on what to do (mapping the data)

js3110:
## Issue

Closes #143 

## Description

When user has not mapped the data (define column names and accepted) it should not be able to go through other tabs (NCA or Outputs) as it can be confusing for them to see errror messages afterwards. Instead, those tabs should remain disable and perhpaps even show a notification

## Definition of Done

- [x] User cannot use other tabs until data is mapped
- [x] Once data is mapped, other tabs are enables

### DIFF
--- a/inst/shiny/modules/column_mapping.R
+++ b/inst/shiny/modules/column_mapping.R
@@ -277,6 +277,11 @@ column_mapping_server <- function(id, data, manual_units, on_submit) {
     observeEvent(input$submit_columns, {
       Sys.sleep(1) # Make this artificially slow to show the loading spinner
 
+      # Enable other tabs
+      shinyjs::enable(selector = "#page li a[data-value=nca]")
+      shinyjs::enable(selector = "#page li a[data-value=visualisation]")
+      shinyjs::enable(selector = "#page li a[data-value=tlgs]")
+
       req(data())
       dataset <- data()
 
@@ -348,6 +353,7 @@ column_mapping_server <- function(id, data, manual_units, on_submit) {
 
       # Execute the callback function to change the tab
       on_submit()
+
     })
 
     list(
--- a/inst/shiny/modules/column_mapping.R
+++ b/inst/shiny/modules/column_mapping.R
@@ -279,7 +279,7 @@ column_mapping_server <- function(id, data, manual_units, on_submit) {
 
       # Enable other tabs
       purrr::walk(c("nca", "visualisation", "tlg"), \(tab) {
-        shinyjs::disable(selector = paste0("#page li a[data-value=", tab, "]"))
+        shinyjs::enable(selector = paste0("#page li a[data-value=", tab, "]"))
       })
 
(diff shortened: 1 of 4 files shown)


inst/shiny/ui.R
```diff
@@ -221,12 +222,12 @@ fluidPage(
...
 
     # VISUALISATION ----
-    nav_panel("Visualisation",
+    nav_panel("Visualisation", value = "visualisation",
       fluid = TRUE,
       tab_visuals_ui("visuals")
     ),
     # New TLG tab
-    nav_panel("TLG",
+    nav_panel("TLG", value = "tlgs",
```
m-kolomanski, line 230: nitpick: Since every other element (title, module name, module id) refers to this tab as `tlg` and not `tlgs`, I would keep that consistent and give it the same value.

inst/shiny/modules/column_mapping.R
```diff
@@ -277,6 +277,11 @@ column_mapping_server <- function(id, data, manual_units, on_submit) {
     observeEvent(input$submit_columns, {
       Sys.sleep(1) # Make this artificially slow to show the loading spinner
 
+      # Enable other tabs
+      shinyjs::enable(selector = "#page li a[data-value=nca]")
+      shinyjs::enable(selector = "#page li a[data-value=visualisation]")
+      shinyjs::enable(selector = "#page li a[data-value=tlgs]")
```
m-kolomanski, line 283: suggestion: In repetitive calls like this you can utilize some functional programming with `purrr`:
```suggestion
  purrr::walk(c("nca", "visualisation", "tlgs"), \(tab) {
    shinyjs::disable(selector = paste0("#page li a[data-value=", tab, "]"))
  })
```
but in this case, considering not much code repetition, it is absolutely fine.

m-kolomanski (approved):
Looks good to me, thank you! I have added two nitpicks/suggestions to consider, but the change is fine as is.

Gero1999 (changes requested):
Hey good job! I think this solution is clean and simple 👌🏽 

In terms of feature I will also add a nitpick request! Could you also include a message in `Review Data` that until the dataframe is uploaded says something like: "Please, remember to submit your variables mapping in `Mapping and Filters`"

Gero1999 (approved):
lgtm good job!

PR #157, opened 2025-01-08
Bug: application glitches when the user changes the pages too fast

### LINKED ISSUE #119: Bug: slope plots glitches when 'next page' button is clicked too quickly
### Description
If the 'next page' button is consecutively clicked too fast, the page starts to glitch

### To reproduce
Run NCA -> double or triple click Next Page button

### Expected behaviour
No glitching

m-kolomanski:
## Issue

Closes #119 

## Description

Adds a JS script that disables the page buttons for 100ms after clicking, preventing the user clicking on them too fast. This can be used with any button within the app, all is needed is to assign `.btn-page` class to the `actionButton()`.

## Definition of Done

- [x] Application does not glitch when pages are changed quickly.

## How to test

Run aNCA analysis as normal, go into slope selector. Try changing the pages quickly. From my experience, the 100ms delay before enabling the button is enough to prevent glitching and is not detrimental to user experience - might depend on the machine it is run on though, please let me know your experience.

## Contributor checklist


inst/shiny/www/index.js
```diff
@@ -0,0 +1,15 @@
+/**
+ * Disables all buttons with class `.btn-page`.
+ * Enables them after 100ms.
+ * This is to prevent the user changing the pages
+ * too fast, which will cause glitches.
+ */
+const togglePageButtons = function() {
```
Gotfrid, line 7: nitpick
```suggestion
function togglePageButtons() {
```

Gotfrid (approved):
Nice solution, lgtm.

js3110 (changes requested):
Looks good! Can you just add the same feature to the previous page button as well 😄

m-kolomanski:
Hey @js3110 the feature is implemented for both buttons, do you mean that it is not working for you with the "previous" button? But the "next" button works as expected?

js3110:
> Hey @js3110 the feature is implemented for both buttons, do you mean that it is not working for you with the "previous" button? But the "next" button works as expected?

ah- I didn't realise. In that case, yeah, for me its not working with the previous button.

m-kolomanski:
> > Hey @js3110 the feature is implemented for both buttons, do you mean that it is not working for you with the "previous" button? But the "next" button works as expected?
> 
> ah- I didn't realise. In that case, yeah, for me its not working with the previous button.

It is super weird that it does not work with only one of the buttons, the system is the same. I am presuming you mean the application glitches as before?

Would you mind going into the `www/index.js` folder on your side and increasing the number on line 12 to 150, 200, 300 and telling me if that helps in your case?

js3110:
Hi, I restarted my R session and then it was the same for both buttons. However, I am still able to make it glitch if I click fast. I increase to 300 and it improved, but it still glitches if I click too fast.

Gotfrid:
Do you by any chance run the app from rstudio viewer? In the past I noticed that javascript might not work correctly this way. Could you please try running the app in google chrome to see if the problem is the same?

js3110:
> Do you by any chance run the app from rstudio viewer? In the past I noticed that javascript might not work correctly this way. Could you please try running the app in google chrome to see if the problem is the same?

I would but unfortunately I have no idea how to do that 😬 if you could let me know how I would appreciate it!

I think this would be something we would want to work for RStudio though as that is how a lot of our users will be running it.

js3110:
ah wait dw I figure out how to run it in chrome
I still get the same issue though, it glitches for me.

m-kolomanski:
@js3110 Would you mind providing a screencap of the bug happening, so I can get a better understanding of how to reproduce it, how exactly you run the application ect? THB I am flabbergasted, I struggle to find a reason why one button would work and the other wouldn't and I have failed to reproduce the issue.

js3110:
> @js3110 Would you mind providing a screencap of the bug happening, so I can get a better understanding of how to reproduce it, how exactly you run the application ect? THB I am flabbergasted, I struggle to find a reason why one button would work and the other wouldn't and I have failed to reproduce the issue.

Hi @m-kolomanski sorry, I should have been more clear. It is glitching with both buttons, the same way as before. There was only one time when I thought it was working for one of the buttons, but maybe I just was clicking slower without realising. Now it still glitches for both buttons on my side. In Rstudio and in Chrome. I'll attach a screencap. I changed the index.js from 100 to 500 in this example, but it doesn't seem to make a difference.

https://github.com/user-attachments/assets/f4b5b7c1-d817-4901-97ab-2cb3aa52a998

js3110:
@m-kolomanski ok, I ran in incognito mode with the button turned up to 5000, and nothing changed. The button did not disable at all and it still glitched. I restarted my R session as well but it didn't make a difference.

js3110:
> @js3110 i am most curious about this. we can jump on a ~pair-programming~ debugging session 😄 i am available for the next ~30 minutes.

Yeah sure! I am also available right now

Gotfrid:
thank you @js3110 for showing the problem.

@m-kolomanski i was able to reproduce the issue on my end: i noticed that it's related to "heavy" plots that take more time to generate. user clicks "next" button, shiny starts rendering the plot, and while that is happening, the user is able to click the button multiple times. which shouldn't be the case, but hey. maybe shiny blocks js thread or something - i'm going to investigate it further now.

Gotfrid:
I also think that it could be a matter of microseconds racing: would JS block the button sooner than another signal is sent to the server - so it might depend on the computer performance.

I suggested to work around this by disabling the buttons on the R side. It's not as elegant as what @m-kolomanski proposed, but maybe it does the job?

@js3110 could you please test it out?

m-kolomanski:
@Gotfrid In my experience JavaScript is always faster than doing UI-related operations through R/Shiny. I have tried more-or-less the same solution that you have proposed at the beginning, but it failed (I could click fast enough for the app to glitch if I tried). I do not understand how is R blocking that JS event? Should be just handled by the browser. 

Also, I believe we can have the two solutions working together?

Gotfrid:
@m-kolomanski that is not what i mean. you have a single button and 2 competing event listeners: one defined by shiny binding (on button click send data from client to server), and another one that you created. once a button is clicked, both events are added to the call stack. it's only a hypothesis, but i think it's possible that multiple consecutive js/shiny events are added to the stack before a single "disable button" event is added.

regarding co-existence of logics - better not imo. let's not scatter the logic and commit to one solution. also, now that i write this - there might be a slightly different solution all together. instead of disabling buttons, maybe we can have a reactive like `should_render <- reactiveVal(TRUE)`, set it to false once plot output is rendered, but set it to true once a button is clicked. and in the plot output have a `req(isTRUE(should_render()))`.

might look like a hallucination tho 😆

js3110:
hi @m-kolomanski  @Gotfrid i tried again and its much better! I can still recreate the glitch- but only if I try to break the world record for speed clicking. With normal and fast clicks it works now. So I'm happy to approve unless you want to keep working on it?

m-kolomanski:
@Gotfrid @js3110 Thank you for looking further and working on this!

@Gotfrid Thank you for the clarification, I get it now! I have tested approach with `should_render` briefly, it seems to me neither better nor worse to what you have already proposed.
@js3110 TBH, I do not know what else we could do in that regard, so if you feel the current solution is enough then I am happy and would suggest moving forward.

js3110 (approved):
Ok, lets leave it there then. If it becomes more of an issue we can always come back to it, but I am happy for now! Thanks

Issue #165, opened 2025-01-20
Bug: Slope Selector Plots incorrect


js3110:
### Description
The slope selector plots do not always match the NCA results, showing the wrong line of best fit or points included in lambda z. This is best noticed when comparing the output of one analyte vs multiple analytes for the same subject profile.

Looks like lamda.z.ix is being calculated incorrectly.
In addition, if user uses the table and selects 1:8 as manual slope, the selection includes 2,3,4,5,6,7 but not 1 and 8. Would be better if it was inclusive of the range the user inputs.

### To reproduce
Run for one analyte
Run again for multiple analytes
Notice the plots have changed (should be the same for each one)
Notice that the points included in the plots don't always correspond to the lambda z points included in the NCA results.

### Expected behaviour

- [ ] Slope plots should correspond exactly to the NCA results, and be the same no matter how many analytes / pcspec have been selected.

- [ ] Exclusions and inclusions should work fine, for both manual table adding or adding via interactive plots
- [ ] Inclusions should include the IX values the user inputs

Issue #120, opened 2024-11-15
Bug: Settings upload does not update dose number


js3110:
### Description
Settings upload does not update dose numbers in Settings Tab of NCA

### To reproduce
Upload settings
### Expected behaviour
Settings upload should find the dose number (profile) for the settings, and update the dose number widget in settings tab so that these dose numbers are selected in the selectInput

Gero1999:
hey @js3110 @m-kolomanski  if you don't mind I will take over this one. I feel it could collapse a little bit with what I am currently changing!

Gero1999:
> @Gero1999 I have also just noticed that settings upload also doesn't work if no exclusions/inclusions were added. Maybe you can include that in this issue? Or should I make a separate one?

Added here!

PR #166, opened 2025-01-20
Enhancement: NCA for multiple analytes

### LINKED ISSUE #3: Feature: Run NCA simultaneously over multiple analytes
**Issue description**
The current used PKNCA package allows the analysis to be run on multiple analytes simultaneously, however our current implementation only provides input selections for one analyte at a time. This can be changed in order to fully satisfy the user's needs. 

- [ ] User should be able to multiple choice in main parameters as now (`ANALYTE`, `DOSNO`) and new ones (`PCSPEC`)
- [ ] Multiple analytes should be able to be run, and slope selector should keep working well
- [ ] Different analytes with different ROUTE, MATRIX...

**Notes**
This feature can be of great interest to enable other future enhancements to the app; i.e, the customization of analyte ratios

js3110:
## Issue

Closes #3 

## Description

The current used PKNCA package allows the analysis to be run on multiple analytes simultaneously, however our current implementation only provides input selections for one analyte at a time. This can be changed in order to fully satisfy the user's needs. 

## Definition of Done

- [x] User should be able to multiple choice in main parameters as now (`ANALYTE`, `DOSNO`) and new ones (`PCSPEC`)
- [x] Multiple analytes should be able to be run, and slope selector should keep working well
- [x] Different analytes with different ROUTE, MATRIX...

How to test:
I have uploaded a dummy dataset with multiple analytes and mutliple matrices in the inst/shiny/data file. (its just the same as the dummyRO but copy and pasted data with ANALYTE or PCSPEC changed. Upload this file and test with different analytes.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

This pull request is limited to the main NCA - setup and results.
I updated the slope selector partially, but noticed a few bugs that are semi-related to the multiple analyte feature, but not completely, so I will be addressing them in separate issues #165 and #120, so that this issue does not become too big.


inst/shiny/tabs/nca.R
```diff
@@ -214,7 +227,7 @@ observeEvent(input$submit_analyte, priority = 2, {
 
   mydose <- PKNCA::PKNCAdose(
     data = df_dose,
-    formula = DOSEA ~ TIME | STUDYID + PCSPEC + DRUG + USUBJID,
+    formula = DOSEA ~ TIME | STUDYID + PCSPEC + DRUG + USUBJID + ANALYTE,
```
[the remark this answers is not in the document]
js3110: Oops you are right, its because the dummy dataset I was using to test contained duplicates so this was the only way to get it to work! I will fix

inst/shiny/tabs/nca.R
```diff
@@ -183,17 +194,19 @@ observeEvent(input$submit_analyte, priority = 2, {
   dosno_column <- "DOSNO"
   route_column <- "ROUTE"
   analyte_column <- "ANALYTE"
+  matrix_column <- "PCSPEC"
```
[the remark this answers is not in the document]
js3110: its needed for filtering which is why I created a seperate one. for the most part it is in group columns so its not hardcoded everywhere else, only in line 206 is this used

inst/shiny/modules/tab_visuals.R
```diff
@@ -201,7 +215,20 @@ tab_visuals_server <- function(id, data, grouping_vars, res_nca) {
       updatePickerInput(
         session,
         "generalplot_analyte",
-        choices = param_choices_analyte
+        choices = param_choices_analyte,
+        selected = param_choices_analyte[1]
```
m-kolomanski, line 219: nitpick(non-blocking): No need to specify selected argument, upon choices change this will automatically change to the first element.

```suggestion
        choices = param_choices_analyte
```

inst/shiny/modules/tab_visuals.R
```diff
@@ -201,7 +215,20 @@ tab_visuals_server <- function(id, data, grouping_vars, res_nca) {
...
+      # Update pcspec picker input
+      param_choices_pcspec <- data() %>%
+        pull(PCSPEC) %>%
+        unique()
+
+      updatePickerInput(
+        session,
+        "generalplot_pcspec",
+        choices = param_choices_pcspec,
+        selected = param_choices_pcspec[1]
```
m-kolomanski, line 231: nitpick(non-blocking): No need to specify selected argument, upon choices change this will automatically change to the first element.
```suggestion
        choices = param_choices_pcspec
```

inst/shiny/modules/tab_visuals.R
```diff
@@ -236,7 +263,8 @@ tab_visuals_server <- function(id, data, grouping_vars, res_nca) {
       updateSelectInput(
         session,
         "analyte_mean",
-        choices = sort(analyte_choices)
+        choices = sort(analyte_choices),
+        selected = analyte_choices[1]
```
m-kolomanski, line 267: nitpick(non-blocking) / question: The same as before, but here we are sorting the choices, but picking the selected from an unsorted vector. Is that intended? It might happen that the selected option will not be the first option of the sorted choices.
```suggestion
        choices = sort(analyte_choices)
```

inst/shiny/modules/tab_visuals.R
```diff
@@ -250,8 +278,20 @@ tab_visuals_server <- function(id, data, grouping_vars, res_nca) {
...
+      # Update pcspec mean choices
+      pcspec_choices <- data() %>%
+        pull(PCSPEC) %>%
+        unique()
+
+      updateSelectInput(
+        session,
+        "pcspec_mean",
+        choices = sort(pcspec_choices),
+        selected = pcspec_choices[1]
```
m-kolomanski, line 290: nitpick(non-blocking) / question: The same as before, but here we are sorting the choices, but picking the selected from an unsorted vector. Is that intended? It might happen that the selected option will not be the first option of the sorted choices.
```suggestion
        choices = sort(pcspec_choices)
```

inst/shiny/tabs/nca.R
```diff
@@ -18,8 +18,19 @@ observeEvent(data(), {
   updateSelectInput(
     session,
     inputId = "select_analyte",
-    label = "Choose the analyte :",
-    choices = unique(data()$ANALYTE)
+    label = "Choose the Analyte(s) :",
+    choices = unique(data()$ANALYTE),
+    selected = unique(data()$ANALYTE)[1]
```
m-kolomanski, line 23: nitpick(non-blocking): No need to specify selected argument, upon choices change this will automatically change to the first element.
```suggestion
    choices = unique(data()$ANALYTE)
```

inst/shiny/tabs/nca.R
```diff
@@ -18,8 +18,19 @@ observeEvent(data(), {
...
+  )
+})
+
+observeEvent(data(), {
+  updateSelectInput(
+    session,
+    inputId = "select_pcspec",
+    label = "Choose the Specimen Type(s) :",
+    choices = unique(data()$PCSPEC),
+    selected = unique(data()$PCSPEC)[1]
```
m-kolomanski, line 33: nitpick(non-blocking): No need to specify selected argument, upon choices change this will automatically change to the first element.
```suggestion
    choices = unique(data()$PCSPEC)
```

inst/shiny/tabs/nca.R
```diff
@@ -275,7 +289,12 @@ output$datatable <- renderReactable({
 # Define the profiles (dosno) associated with each patient (usubjid) for the selected analyte
 profiles_per_patient <- reactiveVal(NULL)
 observeEvent(mydata(), {
-  profiles_per_patient(tapply(mydata()$conc$data$DOSNO, mydata()$conc$data$USUBJID, unique))
+  profiles_per_patient(
+    mydata()$conc$data %>%
+      mutate(USUBJID = as.character(USUBJID)) %>%
+      group_by(USUBJID, ANALYTE, PCSPEC) %>%
+      summarise(DOSNO = list(unique(DOSNO)), .groups = "drop")
+  )
```
m-kolomanski, line 297: issue: Since `profiles_per_patient` is only ever updated by the same formula, at this point it should simply be a `reactive` instead of `reactiveVal`. This would be a nice small step toward refactoring the data flow.

m-kolomanski (changes requested):
Hey, in general looks good, nice job!

I have left a bunch of non-blocking nitpicks, act on them at your own discretion.

In addition, I think there is one opportunity to take a small step towards refactoring the data flow (this is changing the `profiles_per_patient` from `reactiveVal` to a simple `reactive`), so I am requesting a change specifically for that.

inst/shiny/tabs/nca.R
js3110: @m-kolomanski its updated 3 times- twice using mydata() as the dataset, and once using resnca() as the dataset (in slope selector module). Does this still mean it can be changed to reactive?

js3110:
@m-kolomanski thank you for your comments! Regarding the selectinputs, if I don't select an option then no default is selected. I think this is due to the fact that multiple = TRUE. So if its ok I will keep it as being selected, however I have updated so that the sorted ones select the first sorted variable instead.

inst/shiny/tabs/nca.R
m-kolomanski: Thank you for pointing that out!

Yes, I strongly believe so. Especially that the `nca.R` tab/file/future module has the access to `res_nca()` and there is no reason why `slope_selector` specifically should perform that update. I do not see anything in `slope_selector` module that impacts this, it is a simple event that reacts to changes in `res_nca()`, so `nca.R` can do this, and as a result we will _centralize_ calculation of this reactive, instead of having multiple obscure updates split across files/modules. So I would go with a `reactive()` with some `if else` statement to determine whether we wish to use `mydata()` or `res_nca()` at that particular invalidation.

If I am missing something else, please let me know!

m-kolomanski:
> @m-kolomanski thank you for your comments! Regarding the selectinputs, if I don't select an option then no default is selected. I think this is due to the fact that multiple = TRUE. So if its ok I will keep it as being selected, however I have updated so that the sorted ones select the first sorted variable instead.

Ah, you are correct! I am sorry, that is an oversight on my part. Thank you for pointing that out!

inst/shiny/tabs/nca.R
js3110: Ok great thank you! All done :) hopefully it looks good now! @m-kolomanski

m-kolomanski (approved):
Looks good, thank you for the changes!

Gero1999:
Hey great job @js3110! I included the testing dataset (I remake it based on DummyRO). The problem you were finding with it was a 1 line error where `analyte_column` was used for `format_pkncadose_data`. Feel free to modify it further more if needed, but in principle that should be like that!

On the other hand, most of these changes keep the structure hardcoded. In the future, once our current urge finish we might need to reopen the issue to eliminate these dependencies so `additional_group_columns` can also easily work within the workflow.

Gero1999 (approved):

PR #169, opened 2025-01-20
add docs

### LINKED ISSUE #164: Documentation: Add instruction manual
Add pdf guide to documentation and update README

js3110:
## Issue

Closes #164


m-kolomanski (approved):
LGTM, thank you!

Gero1999 (changes requested):
Hey @js3110  thanks for this, looks good! 

Only thing is that it would be good if the source document is also here, even if the link in our README only leads to the pdf. That way whenever there are relevant modifications in the workflow we all can easily change it!

m-kolomanski:
@Gero1999 It depends. Do we want to commit to maintaining this document for a longer period of time? The alternative would be to mark this solution as just a temporary measure, and when we have more important stuff out of the way, we could use this doc to write some complete documentation, on github pages and whatnot.

Depends if we consider this a **temporary** solution or a _temporary_ solution ¯\_(ツ)_/¯ Both are fine, I am just adding my two cents to consider.

Gero1999 (approved):
@m-kolomanski this is just a temporary solution until the github pages is well deployed. then we can put a good tutorial/manual instructions in there 

So I will approve directly @js3110 , good work!

Issue #162, opened 2025-01-16
Enhancement: Delete Data Selection and move to Settings


Gero1999:
### Description

1) All currently done in the tab `Data Selection` can be moved into `Settings` and adapted simplifying both code and UI

2) It will make more sense that the filterings of `Choose the analyte` and `Choose the Dose Number` are applied to the `mydata$intervals` instead of to `mydata$data`

3) It will make more sense that instead of a view of `mydata$data` dataframe, the user after pressing `Submit (settings)` obtains a dataframe of `mydata$intervals`

Gero1999:
I need it to do it anyway for #168 to refactor, so is being already developed there

PR #174, opened 2025-01-23
Enhancement: NCA parameter input

### LINKED ISSUE #168: Bug: NCA Parameters missing
### Description
Many of the NCA parameters are missing from the intervals set, so are not being calculated. All parameters should be calcuated, and the user can then select in the results which they want to see.

### To reproduce
Run NCA- parameters such as Cl and V are missing.

### Expected behaviour
All NCA parameters calculated.

Gero1999:
## Issue

Closes #168  and perhaps #162 

## Description
There are many possible parameters available with the PKNCA package. We want an input table in settings for the user to select which parameters they would like to have calculated.

Change description.

## Definition of Done

- [X] User can select parameters they want to calculate

- [x] If parameter is not possible to calculate (I tested with: f, mrt.md.obs... ) won't crash the app!

- [X] All previously done in tab `Data Selection` has been simplified for the user in `Settings` tab. This also makes the code internally more logical and fluent. 

- [x]  Instead of a view of mydata$conc$data dataframe, the user has a view of their customized NCA details with mydata$intervals

## How to test

Perform data mapping > NCA settings:

- [x] NCA data selection features are in NCA settings
- [x] Parameter selection input can be used to run NCA and obtain the specified parameter results
- [x] Parameter selection input does not crash with all parameters selected
- [x] In NCA settings a `PKNCAdata$intervals` object is displayed that dynamically changes with the user settings
- [x] select_pcspec works as expected and as it used to work! 

## Contributor checklist
- [x] New logic is documented


Gero1999:
This closes #168 and perhaps #162 due to what I needed to change. I included reviewing/testing instructions for the things I consider more sensitive, but feel free to explore/review on whatever way feels more comfortable to you!

js3110 (changes requested):
Hi @Gero1999 , thank you for the work, it looks really good so far!
I have some requests:

**UI Changes**

- I don't think it should automatically select all dose numbers, instead just select the first one, as all is too much and difficult to remove them all easily.
- The parameter selection is not user friendly. There are way too many parameters and in such a small box its hard to know what to select. My suggestion would be a table, similar to the unit parameters. Where the params are grouped into categories (like auc vars etc) so the user has a clearer view of what is available and what they can select.

**Bugs**

- If I select the rule sets, the app crashes
- Partail AUC selection doesn't crash, but the partial auc column is not in the results. I assume this would need to be added as another parameter?
- PPTESTCD column in the results needs to be removed
- Lambda.z.ix column in the results is mapped incorrectly

**Coding**
I would suggest that if possible, you create a 'settings/ setup' module for the code for this area, as it is all quite contained and we need to start modularising the nca tab.

Let me know if you have any questions :)

Gero1999:
@js3110 everything requested solved with 2 exceptions:

- there can still be some problems with lamda.z.ix, will be better as a new issue. Easy to fix once we have the `duplicate_predoses` function merged
- Layout as discussed with a categorized list. I am not sure if I understood the problem with the UI, let me know if not exactly the widhts or position desired for it!

inst/shiny/tabs/nca.R
```diff
@@ -112,6 +93,15 @@ observeEvent(input$settings_upload, {
     selected = doses_selected
   )
 
+  # Specimen or Matrix
+  updateSelectInput(
+    session,
+    inputId = "select_pcspec",
+    label = "Choose the Specimen/Matrix:",
+    choices = sort(unique(data() %>% filter(ANALYTE == setts$ANALYTE[1]) %>% pull(PCSPEC))),
+    selected = unique(data() %>% filter(ANALYTE == setts$ANALYTE[1]) %>% pull(PCSPEC))
```
m-kolomanski, line 102: issue: We do the same operation twice here, would be nicer to save this to a variable and use that in `updateSelectInput`.

inst/shiny/tabs/nca.R
```diff
@@ -387,11 +387,33 @@ observeEvent(input$removeAUC, {
...
+  if (input$rule_adj_r_squared) nca_params <- c(nca_params, "adj.r.squared")
+  if (input$rule_aucpext_obs) nca_params <- c(nca_params, "aucpext.obs")
+  if (input$rule_aucpext_pred) nca_params <- c(nca_params, "aucpext.pred")
+  if (input$AUCoptions) nca_params <- c(nca_params, "aucint.last", "aucint.inf.obs",
+                                        "aucint.inf.pred", "aucint.all")
+  
+  updatePickerInput(session = session,
+                    inputId = "nca_params",
+                    selected = nca_params)
+  print("session updated for nca_params")
```
m-kolomanski, line 404: issue: use `log_*` functions if you wish to log.

inst/shiny/tabs/nca.R
```diff
@@ -434,70 +454,65 @@ observeEvent(input$nca, {
...
+
+      # Increment progress to 100% after NCA calculations are complete
+      incProgress(0.5, detail = "NCA calculations complete!")
+
+      # Make the starts and ends of results relative to last dose using the dose data
+      myres$result <- myres$result %>%
+        inner_join(select(mydata()$dose$data, -exclude)) %>%
+        dplyr::mutate(start = start - !!sym(mydata()$dose$columns$time),
+                      end = end - !!sym(mydata()$dose$columns$time)) %>%
+        dplyr::select(names(myres$result))
```
m-kolomanski, line 502: nitpick: namespacing

```suggestion
        mutate(start = start - !!sym(mydata()$dose$columns$time),
                      end = end - !!sym(mydata()$dose$columns$time)) %>%
        select(names(myres$result))
```

m-kolomanski (changes requested):
Hey,

the code quality looks good, I have only found some minor issues.

In addition, I would like to ask you to bump the package version with this PR. I know your PR template did not include the checkmark, but since #178 was merged, we should start versioning.

Also, I strongly agree with @js3110. If we can extract a part of this into a module that holds the UI for specifying options and returns some kind of list that is then used up the line to prepare the data - that would be ideal!

js3110:
> @js3110 everything requested solved with 2 exceptions:
> 
> * there can still be some problems with lamda.z.ix, will be better as a new issue. Easy to fix once we have the `duplicate_predoses` function merged
> * Layout as discussed with a categorized list. I am not sure if I understood the problem with the UI, let me know if not exactly the widhts or position desired for it!

Hi @Gero1999 , thanks for the updates! 
I still have some bugs:

- The app is still crashing for me when I select all the rule sets.
-  Partial AUC is also not working and instead gives an error saying that there are duplicates.
- The partial aucs sometimes show up in the setup table but sometimes they don't.
- Multiple analyte selection not possible/ buggy

In regards to UI:

-  I like this version of the parameter selection, but it would be good if there were some select all options (I think this was included in the photo you sent me). 
-  Thank you for changing dose number so only one is selected, it would be good to do the same for analytes and matrix, as at the moment they are all selected.
- The error message given when a parameter is selected that cannot be calculated- can we change this slightly? It says 'Error- please report bug', followed by more info. It would be better if instead it has the info of why there is an error, and then 'If this error is unexpected, please report bug'. Otherwise users might report a bug that doesn't exist, if the problem is just that they have selected the wrong parameter.

Gero1999:
I made some changes to satisfy the requirements. Some notes regarding:

#### Modularization
Created a module for "NCA settings" and refactored code. Further submodules are needed due to complexity (I added a `#nolint` line for now in the module function). I will address this partially in issue #120.

#### Parameter Selection with action buttons
I personally think they are rarely needed, as users will never want to select all parameters (most cases not even a quarter of them). I would instead consider auto-selecting key parameters based on data, which can be an independent and less priority issue.

#### Analyte Selection Bug
Issue should be fixed. Report any specific bugs that are not in main via chat for review. If possible indicate for them the dataset used and action taken @js3110!

js3110 (changes requested):
Hi @Gero1999 thanks! The error message is perefect now, and the other changes look good too!
Sorry but I have noticed more bugs 😢 
1. In the selectInputs you are able to delete the selected vars from the boxes, but then it gives an error message (I guess since the table is being generated instantly). This happened to me as I used a dataset where the first PCSPEC was not the one I wanted so I deleted it, but then the app crashed. It would be good if either the user is unable to delete all options, or if they do the app doesn't crash and instead just gives a warning or smth.
2. The partial areas are good! However now we are getting all the vars (AUClast, AUCinf, AUCinfpred etc) for the partial area. The partial area should only give one extra parameter: AUClast(start-end). Since all the other aucs will be the same value, they aren't needed.

Thanks! (and sorry)

Gero1999:
> 1. In the selectInputs you are able to delete the selected vars from the boxes, but then it gives an error message (I guess since the table is being generated instantly). This happened to me as I used a dataset where the first PCSPEC was not the one I wanted so I deleted it, but then the app crashed. It would be good if either the user is unable to delete all options, or if they do the app doesn't crash and instead just gives a warning or smth.
> 2. The partial areas are good! However now we are getting all the vars (AUClast, AUCinf, AUCinfpred etc) for the partial area. The partial area should only give one extra parameter: AUClast(start-end). Since all the other aucs will be the same value, they aren't needed.

hello @js3110!
1. Crush does not occur (function fix + test) and a notification message will pop up to the user
2. AUClast is now the only parameter calculated for AUC intervals

js3110 (approved):
LGTM !!

inst/shiny/modules/nca_settings.R
```diff
@@ -0,0 +1,609 @@
...
+    input_limit <- function(input_id, max = Inf, min = -Inf, update_fun = updateNumericInput) {
+      observeEvent(input[[input_id]], {
+        if (input[[input_id]] < min & !is.na(input[[input_id]])) {
+          update_fun(session, input_id, "", value = min)
+        }
+        if (input[[input_id]] > max & !is.na(input[[input_id]])) {
+          update_fun(session, input_id, "", value = max)
+        }
+      })
+    }
```
m-kolomanski, line 395: suggestion: In another note to this, putting this function outside of the `server` function would reduce the issues with complexity, but will require passing the `input` object as an argument.

inst/shiny/modules/nca_settings.R
```diff
@@ -0,0 +1,609 @@
...
+          mutate(across(c("aucint.last"), ~TRUE)) %>%
+          # Identify the intervals as the manual ones created by the user
+          mutate(type_interval = "manual")
+      })
+
+      # Make sure NAs were not left by the user
+      intervals_list <- intervals_list[!is.na(starts) & !is.na(ends)]
+
+      intervals_userinput(intervals_list)
+    })
```
m-kolomanski, line 492: Two suggestions here:

- from what I can see this should be a `reactive` instead of `observe` with `reactiveVal` - I do not see `intervals_userinput` being modified anywhere else?
- This also could be a separate module on its own that generates the appropriate widgets and returns a reactive list with intervals, BUT I do not want to slow down the development of this feature much longer, since the changes are already quite big. So you could modularize it, but I would be happy with just leaving a `#TODO` for the future, if you do not feel like this is the right call right now.

m-kolomanski (changes requested):

inst/shiny/modules/nca_settings.R
```diff
@@ -0,0 +1,609 @@
...
+            my_input_id <- input_id
+            observeEvent(input[[my_input_id]], {
+              if (is.numeric(input[[my_input_id]]) && input[[my_input_id]] < 0) {
+                input_limit(my_input_id, min = 0)
+              }
+            })
+          })
+        }
+      }
+    })
```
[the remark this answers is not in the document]
Gero1999: I like the notification idea and may add it to input_limit. However, I’d like to keep the current behavior since it works well. Would it be ok for now if I make it more readable and efficient like this? 

```
    # Keyboard limits for the dynamically created partial AUC ranges
    observeEvent(auc_counter(), {
      for (i in auc_counter()) {
        input_limit(paste0("timeInputMin_AUC_", i), min = 0)
        input_limit(paste0("timeInputMax_AUC_", i), min = 0)
      }
    })
```

We can always consider in the future doing JS as you mentioned

inst/shiny/functions/partial_auc_input.R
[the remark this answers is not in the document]
Gero1999: I will include it next to the `# TODO: Modularise all 'Handling AUC Intervals' and its related actions`

js3110 (approved):
LGTM :)

inst/shiny/functions/limit_input_value.R
```diff
@@ -0,0 +1,32 @@
...
+
+    if (!is.na(input_value)) {
+      if (input_value < min_value) {
+        update_fun(session, id, "", value = min_value)
+        showNotification(paste0(label, " input min value is ", min_value))
+      }
+      if (input_value > max_value) {
+        update_fun(session, id, "", value = max_value)
+        showNotification(paste0(label, " input max value is ", max_value))
+      }
```
m-kolomanski, line 29: issue: The update_fun also updates the label to empty string, meaning that **Max** / **Min** labels disappear completely.
issue: Please make the notification a **warning** to work semantically.
```suggestion
    if (!is.na(input_value)) {
      if (input_value < min_value) {
        update_fun(session, id, value = min_value)
        showNotification(paste0(label, " input min value is ", min_value), type = "warning")
      }
      if (input_value > max_value) {
        update_fun(session, id, value = max_value)
        showNotification(paste0(label, " input max value is ", max_value), type = "warning")
      }
```

inst/shiny/modules/nca_settings.R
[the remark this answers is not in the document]
m-kolomanski: This is much cleaner code-wise, thank you!

m-kolomanski (approved):
The code looks good, thank you for the changes.

I am accepting the PR, but I have found two very minor issues, would be great to have those fixed before merging. I also have some nitpicks, but those are non-critical.

Issue #121, opened 2024-11-15
Bug: manual slope selection plot does not update if selected point is before current slope


js3110:
### Description
When a manual slope selection is added, the plot only updates if the point selected is > time than the previous selection, otherwise it doesn't update visually. In the NCA results and in the table everything is fine, it is just the visuals of the plot.

### To reproduce
Run NCA -> Slope selector -> select a point that is before the current slope (black) and then click the end point. The black point selected does not turn green.

### Expected behaviour
The black point selected should turn green and be included in the line of best fit

m-kolomanski:
Hi @js3110 , @Gero1999 ,

I have investigated the issue briefly, the dot on the plot is turned green based on the following:
```R
IX %in% lambda_z_ix_rows$IX ~ "hl.included",
```
and in order to pass that, it needs to go through the following checks:
```R
lambda_z_ix_rows <- conc_pknca_df %>%
    filter(
      DOSNO == dosno,
      USUBJID == usubjid,
      !exclude_half.life,
      TIME >= sum(
        subset(
          lambda_res,
          lambda_res$PPTESTCD == "lambda.z.time.first",
          select = c("start", "PPORRES")
        )
      )
    ) %>%
    arrange(IX) %>%
    slice(0:lambda_z_n_points)
```
and the cases we are talking about do not pass the `TIME >= lamba.z.time.first` check, as @Gero1999 suggested. However, even if we remove this check, the results are still wrong because of the slicing at the end, which always takes the same number of points based on:
```R
lambda_z_n_points <- as.numeric(lambda_res$PPORRES[lambda_res$PPTESTCD == "lambda.z.n.points"])
```
The `lambda.z.n.points` in the results for a given subject does not seem to change based on slope selector. I am afraid my knowledge of PKNCA package and NCA analysis in general is too low to allow me to comment on why that my be and if that is the correct behavior - do you have any thoughts on that?

Removing both time check and slicing seems to make the plots work as expected, however all dots are green at the beginning. I do not know if that is what we would want and whether is has any implications to the data / analysis?

js3110:
@m-kolomanski  removing the checks and having all dots green is definitely not an option.
I think that if the previous point is added to the selection then we would want `lambda.z.n.points` and time first to update to include the new point, same as how n.points should update if for example IX 5 is selected on a plot that used to include IX 6.

[...]

PR #177, opened 2025-01-24
Bug: Slope Selector Plots Incorrect

### LINKED ISSUE #165: Bug: Slope Selector Plots incorrect

js3110:
## Issue

Closes #165 

## Description

The slope selector plots do not always match the NCA results, showing the wrong line of best fit or points included in lambda z. This is best noticed when comparing the output of one analyte vs multiple analytes for the same subject profile.
For some multiple analyte studies, the slope plots crash with error message: `Warning in max(lambda_z_ix_rows$AVAL) :
  no non-missing arguments to max; returning -Inf
Warning: Error in if: missing value where TRUE/FALSE needed
  87: lambda_slope_plot [/aNCA/R/lambda_slope_plot.R#134]`

Looks like lamda.z.ix is being calculated incorrectly.
In addition, if user uses the table and selects 1:8 as manual slope, the selection includes 2,3,4,5,6,7 but not 1 and 8. Would be better if it was inclusive of the range the user inputs.

## Definition of Done

- [x] Slope plots should correspond exactly to the NCA results, and be the same no matter how many analytes / pcspec have been selected.
- [x] Exclusions and inclusions should work fine, for both manual table adding or adding via interactive plots
- [x] Inclusions should include the IX values the user inputs
- [x] Should work for multiple analytes without crashing
- [x] Include entire range of points used to calculate lambda z

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer
As well as fixing the bug issues, I have also refactored the code to avoid hardcoding the variables and instead created `slopes_groups()` reactive which takes the column names from `mydata()$conc$columns$groups`


Gero1999:
I will provide comments to the code as you requested @js3110!

R/create_duplicates.R
```diff
@@ -0,0 +1,95 @@
...
+  
+  # Step 2: Calculate dosing intervals
+  dosing_intervals <- dose_times %>%
+    group_keys(groups) %>%
+    mutate(
+      interval_next = lead(dose_time) - dose_time,
+      interval_prev = dose_time - lag(dose_time),
+      interval_next = replace_na(interval_next, 0),
+      interval_prev = replace_na(interval_prev, 0),
+      next_dose = interval_next + dose_time
```
Gero1999, line 53: question: why not `dose_next` and also `dose_prev`?

R/lambda_slope_plot.R
```diff
@@ -250,9 +247,9 @@ lambda_slope_plot <- function(
         plot_data$is.excluded.hl ~ "red",
         plot_data$IX %in% lambda_z_ix_rows$IX ~ "green",
         TRUE ~ "black"
-      ), size = 15, opacity = 0),  # Make points semi-transparent
```
Gero1999, line 253: suggestion: I would keep this comment (at least for me was not so obvious) 😅

inst/shiny/tabs/nca.R
```diff
@@ -712,21 +713,24 @@ output$settings_save <- downloadHandler(
 )
 
 # Keep the UI table constantly actively updated
-observe({
+observeEvent(input, {
+  req(mydata())
+  dynamic_columns <- c(setdiff(unname(unlist(mydata()$conc$columns$groups)), "DRUG"), "DOSNO")
```
Gero1999, line 718: suggestion: just a suggestion but if you want to make this even more top it would be great if you disconsider all columns that in the dataset has only 1 unique() val. That way the table won't be so wide!

R/create_duplicates.R
```diff
@@ -0,0 +1,95 @@
...
+    ) %>%
+    ungroup()
+  
+  # Step 3: Duplicate predose values for the previous dose
+  predose_duplicates <- conc_data %>%
+    filter(.data[[arrlt]] <= 0, .data[[dosno]] > 1) %>%
+    left_join(dosing_intervals, by = c(dosno, groups)) %>%
+    mutate(
+      !!dosno := .data[[dosno]] - 1,
+      !!arrlt := .data[[arrlt]] + interval_prev,
```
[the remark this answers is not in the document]
js3110: this is correct, as it is the previous interval and arrlt < 0 it is calculating the time from the previous dose.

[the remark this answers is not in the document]
js3110: if we - interval_prev then it would be a very negative number

R/lambda_slope_plot.R
```diff
@@ -52,85 +49,69 @@
...
-  conc_pknca_df = mydata$conc$data,
-  dosno = profile,
-  usubjid = patient,
-  analyte = analyte,
-  pcspec = pcspec,
-  R2ADJTHRESHOL = 0.7
+    res_pknca_df = myres$result,
+    conc_pknca_df = mydata$conc$data,
+    column_names = slopes_groups(),
+    row_values,
```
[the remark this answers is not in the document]
js3110: call what? row_values?

inst/shiny/tabs/nca.R
js3110: thats a good idea!!!

R/create_duplicates.R
js3110: dose prev isn't needed so I didn't create it. If you mean that interval should be renamed dose, then no because its the interval and not necessarily the dosing time. So i think its clearer this way.
Not sure what you mean by interval next- its used to calculate next_dose and then it is removed from the data

inst/shiny/tabs/nca.R
js3110: > please also consider eliminating profile_per_patient() from the code if possible... You use it in the subsequent lines, but you don't actually need it anymore and can use the concentration group columns and the dataset to guess it in a much more better way!

@Gero1999 I am confused what you mean? profiles_per_patient is the data for the plots grouped by the slope_groups. It does use the group columns and the dataset and just means we don't have to keep regrouping again. What do you suggest I use instead?

Gero1999: Maybe I recall bad but I think `profiles_per_group()` is a (reactive) named vector indicating the DOSNO for each USUBJID

This can produce confusion and problems in the code, as we want instead of a named vector a dataframe that indicates USUBJID, ANALYTE, DOSNO, DRUG... all (minimal) information that is necessary to define each plot . This Iguess might be the same object that you want to create within these lines!

js3110: hi, this is what profiles_per_patient has already been modified as ! its a reactive dataframe with the unique information required for each plot

inst/shiny/modules/slope_selector.R
```diff
@@ -102,6 +131,24 @@ slope_selector_server <- function(
...
 
+    #Get grouping columns for plots and tables
+    slopes_groups <- reactive({
+      req(mydata())
+      groups <- c(setdiff(unname(unlist(mydata()$conc$columns$groups)), "DRUG"), "DOSNO")
+
+      # Filter columns with more than one unique value
+      filtered_groups <- groups[sapply(groups, function(col) {
+        length(unique(mydata()$conc$data[[col]])) > 1
+      })]
```
m-kolomanski, line 142: suggestion: This is a bit hard to read, maybe consider utilizing some functions from `purrr`? Something like:
```R
mydata()$conc$columns$groups %>%
  purrr::list_c() %>%
  purrr::keep(c(., "DOSNO"), \(col) {
    col != "DRUG" && length(unique(mydata()$conc$data[[col]])) > 1
  })
```
(note: I did not test this properly, might require some tweaking)

inst/shiny/modules/slope_selector.R
```diff
@@ -102,6 +131,24 @@ slope_selector_server <- function(
...
+      filtered_groups
+    })
+
+    # Reactive for .SLOPE_SELECTOR_COLUMNS
+    .SLOPE_SELECTOR_COLUMNS <- reactive({
```
m-kolomanski, line 147: nitpick: With this setup there is no need for the `.`, anything defined within `moduleServer` is local by definition.

m-kolomanski (changes requested):

R/utils-slope_selector.R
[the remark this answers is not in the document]
js3110: hi @m-kolomanski which ones should be exported?
I created the additional three (apply_slope_rules, handle_plotly_click, and plotly_empty_plot) because the slopes selector module had too high cyclomatic complexity and was failing the lint tests. I would say the empty plot function could be exported but slope rules and plotly click only need to be used within the package? But I'm not sure entirely how you can tell the difference! 😕

R/create_duplicates.R
```diff
@@ -0,0 +1,121 @@
...
+#'  ARRLT = c(-1, 0, 1, 1, 2, 3),
+#'  AFRLT = c(0, 0.5, 1, 3, 4, 5),
+#'  NRRLT = c(0, 1, 2, 1, 2, 3),
+#'  NFRLT = c(0, 0.5, 1, 3, 4, 5),
+#' result <- create_duplicates(conc_data,
+#'                             groups = c("USUBJID"), dosno = "DOSNO")
+#' }
+#'
+#' @export
+create_duplicates <- function(conc_data,
```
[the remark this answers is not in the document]
js3110: @m-kolomanski I'm not sure how to define that in a way that makes sense and is less than 5 words... open to suggestions !

js3110:
Hi @m-kolomanski I have pushed some changes based on your comments. Regarding some of the functions, I am not sure if I have put them in the write place or if they need further documentation added, so please let me know!

inst/shiny/functions/handle_plotly_click.R
```diff
@@ -0,0 +1,74 @@
...
+#' `last_click_data` accordingly. If an update is needed, the function exits early.
+#'
+#' @param last_click_data A reactive Values object storing the last clicked data.
+#' @param dynamic_values A named list containing values from the current click event.
+#' @param idx_pnt The index of the clicked data point.
+#' @param slopes_groups A character vector of slope grouping column names.
+#'
+#' @returns Returns `NULL` and exits early if data was updated. Otherwise, continues execution.
+#'
+handle_plotly_click <- function(id, last_click_data, manual_slopes, slopes_groups) {
```
Gero1999, line 13: issue: `id` & `manual_slopes` should be defined in the documentation. Remove those that are no longer arguments

inst/shiny/functions/handle_plotly_click.R
```diff
@@ -0,0 +1,74 @@
...
+  click_data <- event_data("plotly_click")
+  req(click_data, click_data$customdata)
+
+  log_trace("{id}: plotly click detected")
+
+  identifiers <- jsonlite::fromJSON(click_data$customdata)
```
Gero1999, line 19: question: do we really need this dependency? `jsonlite::fromJSON` /  jsonlite::toJSON
Maybe is easier to just make identifiers a regular dataframe with `dplyr` or `base`?

inst/shiny/functions/handle_plotly_click.R
Gero1999: question: is `id` only used for `log_trace`? I would suggest to derive the id of the plot from `last_click_data` directly if so

R/lambda_slope_plot.R
```diff
@@ -54,84 +51,68 @@
...
-    lambda_z_ix_rows$IX[1],
-    ")/2 = ", time_span / 2,
-    "h"
-  )
-
   subtitle_text <- paste0(
     "    R<sup>2</sup><sub>adj</sub>: ", r2adj_value,
-    "    ln(2)/\u03BB = ", half_life_value, "h",
-    "    (T<sub>", lambda_z_ix_rows$IX[2], "</sub> - T<sub>",
+    "    HL \u03BB<sub>z</sub> = ", half_life_value, "h",
```
Gero1999, line 107: issue: Please, make sure that from now on all units such as "h" come from the relevant column (I think it is `PPSTRESU`) to prevent problems in differences between analytes. Same with span ratio!

Gero1999 (changes requested):
### Bugs detected
- [ ] `Plots per page` input failed when changing it. I tested with `DummyRO_with_metab`, using multiple analyte. The error says:
```
TRACE [2025-02-04 11:39:30] slope_selector: Updating displayed plots
Warning: Error in if: argument is of length zero
```

- [ ] With all default settings run, doing a selection in the first plot of `Point4 > Point7` did add a row to the slopes table but did not change the plot


### UI suggestions
- [ ] It is super good that is good the plots have a standard size. But I feel they are way to big, which makes difficult comparison. If we are keen on keeping this size I would suggest increasing then the axis size

R/lambda_slope_plot.R
```diff
@@ -54,84 +51,68 @@
...
-      USUBJID == usubjid,
+      if_all(all_of(column_names), ~ .x == row_values[[deparse(substitute(.x))]]),
       !exclude_half.life,
-      TIME >= sum(
-        subset(
-          lambda_res,
-          lambda_res$PPTESTCD == "lambda.z.time.first",
-          select = c("start", "PPORRES")
-        )
+      ARRLT >= round(
```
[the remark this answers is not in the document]
js3110: hi, I noticed that the PPTESTCD result was automatically doing some rounding, and so in order to make sure that the wrong times weren't filtered out I had to add rounding too. I will try with columns$time and see if that makes a difference!

inst/shiny/functions/handle_plotly_click.R
js3110: I don't mind, I did this because I had an issue with "_" in the value names causing the identifiers to be split wrong, but I'm sure it could be manually coded. What do you think is best @m-kolomanski @Gotfrid ?

R/lambda_slope_plot.R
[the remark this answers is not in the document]
Gero1999: @js3110  but here you are also rounding the result right? Other thing that I would like to point out is to use `PPSTRES` instead of `PPORRES` and `PPSTRESU` instead of `PPORRESU`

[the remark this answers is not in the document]
js3110: @Gero1999 I tried but I cant use the con$data$time column because that is also AFRLT and we need to use ARRLT. The reason we need ARRLT is because the lambda.z.time.first uses ARRLT so otherwise the caclulation of lamda.z.ix.rows is completely wrong. Hope that makes sense!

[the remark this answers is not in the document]
Gero1999: The thing is that if I am correct `start = dose.time.afrlt` so `start + lambda.z.time.first.(arrlt)` = `lambda.z.time.first.afrlt`

js3110:
> ### Bugs detected
> * [ ]  `Plots per page` input failed when changing it. I tested with `DummyRO_with_metab`, using multiple analyte. The error says:
As discussed, we have a new issue #192 to solve this, as its not related to this issue.

> * [ ]  With all default settings run, doing a selection in the first plot of `Point4 > Point7` did add a row to the slopes table but did not change the plot
 This is covered in issue #121 so I will leave it for there

> ### UI suggestions
> * [ ]  It is super good that is good the plots have a standard size. But I feel they are way to big, which makes difficult comparison. If we are keen on keeping this size I would suggest increasing then the axis size
I agree, the problem is when I make them smaller, they are too small on the small screen.  At the moment if you have more than 2 plots per page they just appear underneath in the same size, and you have to scroll down to view, which makes them too big on the big screen. It would be good if there was a way to set the container size and then even if you add more plots per page they stay within the container. Do you know how to do this @m-kolomanski ? My knowledge of css and styling is limited 😨

js3110:
@Gero1999 @m-kolomanski @Gotfrid I have had a go at changing the styling for the plots. On a big screen, it works fairly nicely and as intended, however on the small screen for some reason it disregards everything that is coded in css and just does whatever it wants. I can't seem to solve that 😅 .
The other issue I have noticed is that the "previous_page"and "next_page" buttons are not disabled when they are supposed to be, which means the user can click when they shouldn't and cause a crash. I was also unable to solve this... 
So I would appreciate some help ! Thanks

Gero1999:
> > ### Bugs detected
> > 
> > * [ ]   `Plots per page` input failed when changing it. I tested with `DummyRO_with_metab`, using multiple analyte. The error says:
> 
> As discussed, we have a new issue #192 to solve this, as its not related to this issue.
> 

As we talked I rebased the branch to solve the multiple analytes metabolites case, so we can use `DummyRO_with_metab.csv` for the testing

R/create_duplicates.R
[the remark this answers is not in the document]
m-kolomanski: Maybe something like `adjust_dose_cycles`? Feels like it gives a bigger picture, what do you think?

js3110:
@m-kolomanski I went with dose_profile_duplicates, a bit of combination between both our suggestions. I hope that makes things clearer 😄

inst/shiny/functions/handle_plotly_click.R
m-kolomanski: `plotly` imports `jsonlite` anyway, so I do not think there is any penalty to using it - we depend on it anyway, just by extension.

That said, I have experimented with it for a second and seems like the custom data for a plotly plot can take any json-serializable object and handle the serialization/deserialization itself. So you could drop the `jsonlite` function calls and go:

`lambda_slope_plot.R`
```R
  customdata <- apply(
    plot_data[, c(column_names, "IX"), drop = FALSE],
    1,
    function(row) as.list(setNames(row, c(column_names, "IX")))
  )

  class(customdata) # "list"
```

`handle_plotly_click.R`
```R
click_data <- event_data("plotly_click")
class(click_data$customdata) # "data.frame"
```

Doing this quickly, seems to do the job, but I have not tested it thoroughly.

Gero1999:
Hey @js3110 thanks for the changes!

Found some issues:

### Potential problems / behavior changes (?)
- [ ] The IX of the points seem to now not consider each DOSNO. That means that DOSNO = 2 do not start with IX=1 in the plot

### Bugs
- [ ] Cannot choose points before the first one automatically selected.
- [ ] Selection and exclusions do not affect the points. Could be related with the IX issue.  
- [ ] When using the multiple analytes (file DummyRO_with_metab.csv) pressing `Next` crashes the App

js3110:
> * [ ]  Cannot choose points before the first one automatically selected.

as previously mentioned, this is in a separate issue #121

Gero1999:
Hey @js3110 thanks for the efforts, now it looks better! Only small things:

- [ ] Issue - Potential conflicts in IX duplicity due to predoses: Check `USUBJID: 11101, DOSNO: 3` for an example. I would even suggest not plotting predose samples, as they must not be used in any case in the HL calculation
- [ ] Nitpick - Statistics in plots for HL & span: HL units have no space with the HL value
- [ ] Nitpick - I love the graphical interface for page controlling, but is covering completely the X label

Gero1999 (changes requested):
Hey @js3110 thanks for the efforts, now it looks better! Only small things:

- [x] Issue - Potential conflicts in IX duplicity due to predoses: Check `USUBJID: 11101, DOSNO: 3` for an example. I would even suggest not plotting predose samples, as they must not be used in any case in the HL calculation
- [x] Nitpick - Statistics in plots for HL & span: HL units have no space with the HL value
- [x] Nitpick - I love the graphical interface for page controlling, but is covering completely the X label

R/utils-slope_selector.R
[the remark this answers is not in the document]
m-kolomanski: With our architecture, package (`R/`) != application (`inst/shiny/`).

The functions you have created (apply_slope_rules, handle_plotly_click, and plotly_empty_plot) currently are fine and defined correctly. However, this PR changes the names of:
- filter_slopes -> .filter_slopes
- check_slope_rule_overlap -> .check_slope_rule_overlap
and removes the `@export` statements for those functions.

Both of those functions are part of the package (`R/`), but also used in the application (`slope_selector.R`) module. Therefore, they should not have the dots and be exported. This is the case in `main` currently, however the PR changes that - and it shouldn't.

inst/shiny/functions/handle_plotly_click.R
```diff
@@ -0,0 +1,76 @@
...
+#' @param last_click_data A reactive Values object storing the last clicked data.
+#' @param dynamic_values A named list containing values from the current click event.
+#' @param idx_pnt The index of the clicked data point.
+#' @param manual_slopes A reactive Values object storing the manually added slope rules.
+#' @param slopes_groups A character vector of slope grouping column names.
+#'
+#' @returns Returns `NULL` and exits early if data was updated. Otherwise, continues execution.
+#'
+handle_plotly_click <- function(last_click_data, manual_slopes, slopes_groups) {
+  click_data <- event_data("plotly_click")
```
m-kolomanski, line 16: issue: This object should be passed as an argument instead of being fetched from the global state, in the spirit of functional programming.

inst/shiny/functions/handle_plotly_click.R
```diff
@@ -0,0 +1,76 @@
...
+  req(click_data, click_data$customdata)
+
+  log_trace("slope_selector: plotly click detected")
```
m-kolomanski, line 19: issue: This log should be directly within the `observeEvent` that is triggered upon plotly click, not within the handler.

inst/shiny/functions/handle_plotly_click.R
m-kolomanski, line 1: issue: Extracting this logic into a function is a great idea. However, I would suggest that the function should not directly modify any global object. Instead, passed data should be localized and that copy should be modified. Then a list with results can be returned, and the changes applied directly to the reactive objects within the `observeEvent` that was triggered.

inst/shiny/modules/slope_selector.R
```diff
@@ -102,6 +132,25 @@ slope_selector_server <- function(
...
+      mydata()$conc$columns$groups %>%
+        purrr::list_c() %>%
+        append("DOSNO") %>%
+        purrr::keep(\(col) {
+          !is.null(col) && col != "DRUG" && length(unique(mydata()$conc$data[[col]])) > 1
+        })
+    })
+
+    # Reactive for SLOPE_SELECTOR_COLUMNS
+    SLOPE_SELECTOR_COLUMNS <- reactive({
```
m-kolomanski, line 149: issue: The convention is that **constants** should be named with capitalized letters. Since `SLOPE_SELECTOR_COLUMNS` is now a reactive, therefore not a constant, then its name should be `slope_selector_columns`.

inst/shiny/modules/slope_selector.R
```diff
@@ -275,73 +343,68 @@ slope_selector_server <- function(
...
+      # Isolate to prevent unnecessary re-renders on every edit
       isolate({
         data <- manual_slopes()
       })
 
+      # Fixed columns (TYPE, RANGE, REASON)
+      fixed_columns <- list(
+        TYPE = colDef(
+          cell = dropdown_extra(
+            id = session$ns("edit_TYPE"),
```
m-kolomanski, line 357: nitpick: since `ns` is extracted from `session` object at the top of the `moduleServer`:
```suggestion
            id = ns("edit_TYPE"),
```

inst/shiny/modules/slope_selector.R
```diff
@@ -275,73 +343,68 @@ slope_selector_server <- function(
...
+            choices = c("Selection", "Exclusion"),
+            class = "dropdown-extra"
+          ),
+          width = 200
+        ),
+        RANGE = colDef(
+          cell = text_extra(
+            id = session$ns("edit_RANGE")
```
m-kolomanski, line 365: nitpick: since `ns` is extracted from `session` object at the top of the `moduleServer`:
```suggestion
            id = ns("edit_RANGE")
```

inst/shiny/modules/slope_selector.R
```diff
@@ -275,73 +343,68 @@ slope_selector_server <- function(
...
+          )
+        ),
+        REASON = colDef(
+          cell = text_extra(
+            id = session$ns("edit_REASON")
```
m-kolomanski, line 370: nitpick: since `ns` is extracted from `session` object at the top of the `moduleServer`:
```suggestion
            id = ns("edit_REASON")
```

inst/shiny/modules/slope_selector.R
```diff
@@ -275,73 +343,68 @@ slope_selector_server <- function(
...
+          ),
+          width = 400
+        )
+      )
+
+      # Dynamic group column definitions
+      dynamic_columns <- lapply(slopes_groups(), function(col) {
+        colDef(
+          cell = dropdown_extra(
+            id = session$ns(paste0("edit_", col)),
```
m-kolomanski, line 380: nitpick: since `ns` is extracted from `session` object at the top of the `moduleServer`:
```suggestion
            id = ns(paste0("edit_", col)),
```

m-kolomanski (changes requested):

js3110:
@Gero1999 if you could rereview this as well since I merged your parameters branch and had to make quite a few changes, so would be good if you can double check them :)

Gero1999 (approved):
Hey hey @js3110 everything looks in place. However I have detected a small bug. However, I would be happy to open a different issue for it, so I will approve. Just let me know and I can create it :) 

- [ ] **page control**: when a `search patient` is set and the `next page` is pressed after getting to the "last plot", then the counter seems to keep summing, so to go back to the previous plot you need to redo your way back pressing an equal number of times `previous page`

js3110:
> * [ ]  **page control**: when a `search patient` is set and the `next page` is pressed after getting to the "last plot", then the counter seems to keep summing, so to go back to the previous plot you need to redo your way back pressing an equal number of times `previous page`

@Gero1999  i mentioned this earlier in the thread (although there so much here it defo got lost!), but I had already tried to fix it with no luck. Its not urgent so I would say lets make a new issue! Thanks :)

js3110:
@m-kolomanski if you could take another look, it would be really good if we can merge this branch before next week, as its the last essential one for this sprint

m-kolomanski (approved):
The code looks good to me. Thank you for all the changes and the hard work! I hope your soul can rest now :)

PR #172, opened 2025-01-23
Enhancement/run app via package installation

m-kolomanski:
## Issue

Closes #122 

## Description

This PR supersedes #123. The old PR was done on a very outdated branch, I have decided to implement changes to the most up-to-date version of the app instead of dealing with conflicts.

This changes brings ability to run the app after installing the package - without cloning the repository.

Changes:
- update `aNCA::run_app()
  - add `...` to the signature to be able to pass arguments to `shiny::runApp()` such as host, port, etc
  - remove all import comments due to their redundant nature: the app should require all imports that are listed in the packages NAMESPACE file
  - call `require()` insider the function to load all required libraries
- fixes:
  - export some utility functions that are used by the app
  - fix function namespacing in test files

## Definition of Done

- [x] user can install the package from github via `pak::pak("pharmaverse/aNCA")`
- [x] user can run the app via `aNCA::run_app()`
- [x] README contains information on how to install & run the app
- [x] Package version is bumped

## How to test

Single command to test the app via package installation (requires docker):

```sh
docker run --rm -it -p 3838:3838 rocker/r-ver:latest bash -c "apt update && apt install libcurl4-openssl-dev && Rscript -e \"install.packages('pak'); pak::pak('pharmaverse/aNCA@enhancement/run-app-via-package-installation2')\" && Rscript -e \"aNCA::run_app(host='0.0.0.0',port=3838)\""
```

Alternatively, install the package in your environment via pak from this branch:

```r
pak::pak('pharmaverse/aNCA@enhancement/run-app-via-package-installation2')
```

If you are installing the package without docker, please use some sort of virtual environment management system (like `renv`) OR make sure to start with a clean slate and remove the current installation that is present in your global library by running `remove.packages('aNCA')`.

## Contributor checklist
- [x] Code passes lintr checks
- [x] Code passes all unit tests

## Notes to reviewer

Please perform the functional testing rigorously. I have performed a simple analysis and clicked through all the options, but there is a high probability that I have missed something. There might be a chance that some packages were not properly put into the NAMESPACE, and due to that are not loaded in the app.

In addition, if you notice any packages that are loaded **but should not** due to being outdated / deprecated / superseded / changed for something else and mistakenly not removed from the `NAMESPACE` file, please do let me know.

PR #178, opened 2025-01-24
Introduce instructions on package versioning

m-kolomanski:
Hi everyone,

with rapid development of the package, we are just few critical features and fixes away from having a stable version of the package, which could be published on CRAN. With that, I believe it would be beneficial to introduce **instructions** and **requirements** on package versioning. Proper, semantic versioning is not an important issue at this point, but since #172 added the possibility of installing the package with `pak`, tracking changes in the development version will be beneficial.

Therefore, I would like to suggest adding the **requirement** of bumping the package version in `DESCRIPTION` file, in all PRs which introduce code changes, before merging. We could go with just incrementing the last, .9000 part of the version, as long as we wish, and then when we believe the package is ready for publication, start off with version 0.1.0.

**This pull request** introduces two changes:
- Adds information on package version bumping to the `CONTRIBUTING.md` file
- Adds *Package version is incremented* to the checklist in the pull request template.

Please let me know what you think of introducing this at this point in time, feedback welcome. I think this will get everyone in the team in the healthy habit of incrementing the package version, in preparation of maintaining the _officialy released_ application.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -19,6 +19,7 @@ How to test features not covered by unit tests.
 - [ ] Code passes all unit tests
 - [ ] New logic covered by unit tests
 - [ ] New logic is documented
+- [ ] Package version is incremented
 
 ## Notes to reviewer
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -60,14 +60,24 @@ This part is here as a reminder to perform basic tasks and checks before the cod
 - Code passes all unit tests
 - New logic covered is by unit tests
 - New logic is documented
+- Package version is bumped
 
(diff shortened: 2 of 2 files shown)


Gero1999:
Sounds a good plan! I would also suggest giving a look to the GitHub workflow. Do you think is possible to make it update the DESCRIPTION version on each successful merge?

Gero1999 (approved):

m-kolomanski:
@Gero1999, this is exactly what I have tried doing [with this commit](https://github.com/pharmaverse/aNCA/commit/56bb3a5f1299497b72113f19211a083e09fa99d3) (basically a script to bump package version automatically). I have tested it, in general it works, but not with our setup. We require pull requests that have 2 approvals in order to add changes, so when the bot/script tries to change the package number it gets an error.

Gotfrid (approved):
Great idea, @m-kolomanski. I think for now pull request template is a good enough solution, let's not block this PR over some merge problems (imo).

PR #187, opened 2025-01-31
Refactor: simplify global imports

m-kolomanski:
## Description

Simplifies import statements in `global.R` file. With this change, there is no need to explicitly import every module and function files. All `.R` files that are placed in `modules/` and `functions/` will be automatically sourced to the application.

## How to test
Run the app, check if it works correctly.

## Contributor checklist


Gotfrid (approved):
LGTM, please add a blank line at the end of global.R

js3110 (approved):
LGTM, Thanks!

Issue #116, opened 2024-11-12
Enhancement: AUC intervals table allow ratios


Gero1999:
### Description
AUC intervals tables should at some point allow the user to compute their own metabolite AUC ratios or profile AUC ratios. These values should be included either in the final table as additional columns or in a new table

Bloqued until multiple analytes avaialble #3

### Definition of Done
- [ ] allow user to customize intervals to compute profile AUCs ratios (i.e, AUC_dosno2/AUC_dosno1)
- [ ] allow user to customize intervals to compute metabolite ratios AUCs (considering Molecular Weights to normalize)

Gero1999:
Hmm something really worthy to talk about. We still need to investigate the theory in detail, but the main idea is to allow the user to fill a dataframe with columns for `analyte1`, `analyte2`, `dosnos1`, `dosnos2`,` mw1`, `mw2`. Then for each row the app will just compute the relations in the post-processing, after pk.nca.

billdenney:
Yes, there are several subtleties, and it can be automated. We should talk about it.

js3110:
removed matrix ratios from this issue and created a new issue #191

PR #198, opened 2025-02-07
Enhancement: non nca ratios

### LINKED ISSUE #191: Enhancement: Non NCA Analysis - Ratios
### Description
Some analysis is needed that does not require NCA, especially for preclinical. Including:
- BPP analysis (Blood plasma partitioning)
- Excretion analysis
- Matrix-Ratio analysis

For now, we want to focus on Matrix ratios and BPP. There should be a section where the user can select the different tissues and calculate matrix ratios for each subject for these tissues- formula =  Cmatrix / Cplasma. Results should then be displayed in a table.
For BPP- the user can select which samples to use (select the ANALYTES) assign which is Blood and which is plasma, and select the time point samples to use. Then formula = Blood Conc / PLasma Conc = BP ratio. Results should be shown in  table.

Results tables should include Grouping columns, Subject, Time, Samples used, Ratio, and Mean? (and N)

### Definition of Done
- [ ] User selection of matrix ratio calculation
- [ ] Matrix-Plasma ratios in results table
- [ ] User selection for BPP ratio calculation
- [ ] BPP results table with summary statistics

js3110:
## Issue

Closes #191 

## Description
Some analysis is needed that does not require NCA, especially for preclinical. Including:

BPP analysis (Blood plasma partitioning)
Excretion analysis
Matrix-Ratio analysis
For now, we want to focus on Matrix ratios and BPP. There should be a section where the user can select the different tissues and calculate matrix ratios for each subject for these tissues- formula = Cmatrix / Cplasma. Results should then be displayed in a table.
For BPP- the user can select which samples to use (select the ANALYTES) assign which is Blood and which is plasma, and select the time point samples to use. Then formula = Blood Conc / PLasma Conc = BP ratio. Results should be shown in table.

## Definition of Done

- [x] User selection of matrix ratio calculation
- [x] Matrix-Plasma ratios in results table
- [x] User selection for BPP ratio calculation
- [x] BPP results table

## How to test

## Contributor checklist
- [x] New logic is documented

## Notes for the reviewer:
In order to help better determine the location of the UI, I created some additional blank sections that will be implemented in #116 and #203 , so that the setup would make sense. These sections can stay as blank for now, or they can also be removed and the code copied to the other issues, I am open to suggestions on what is the best option!


inst/shiny/modules/non_nca.R
```diff
@@ -0,0 +1,254 @@
...
+#' @param data A PKNCA data object that returns the data list containing the concentration and
+#' dose data.
+#' @param grouping_vars A character vector of grouping variables to use for the analysis.
+
+# UI function for the non-nca analysis
+non_nca_ui <- function(id) {
+  ns <- NS(id)
+
+  tagList(
+    tabsetPanel(
```
m-kolomanski, line 14: issue: Please use `bslib` components, I think in the app we use `navset_card_pill` and `nav_panel` generally.

inst/shiny/modules/non_nca.R
```diff
@@ -0,0 +1,254 @@
...
+  moduleServer(id, function(input, output, session) {
+    ns <- session$ns
+
+    #TODO: Remove hardcoding of columns ?? Ensure only non-distinct grouping columns are shown ?
+    # BPP Analysis #########################################
+    output$blood_selector <- renderUI({
+      req(data())
+      blood_options <- unique(data()$conc$data$PCSPEC)
+      selectInput(ns("selected_blood"), "Choose Blood column for BPP",
+                  choices = blood_options)
```
m-kolomanski, line 106: question: Is there any reason to render that dynamically? Defining `selectInput` in the ui function and using `updateSelectInput` to set the choices would be preferable.

If we want to _not display_ the widgets before data is available, please use `shinyjs`: `hidden/hide` and `show`

inst/shiny/modules/non_nca.R
```diff
@@ -0,0 +1,254 @@
...
+    filtered_samples <- reactive({
+      req(data(), input$selected_tissues)
+
+      plasma <- input$selected_plasma
+      tissue <- input$selected_tissues
+
+      df_filtered <- data()$conc$data %>%
+        filter(PCSPEC %in% c(input$selected_plasma, input$selected_tissues))
+
+      df_filtered
```
m-kolomanski, line 203: nitpick: No need to return the variable, if it is never assigned

```suggestion
      data()$conc$data %>%
        filter(PCSPEC %in% c(input$selected_plasma, input$selected_tissues))
```

inst/shiny/modules/non_nca.R
```diff
@@ -0,0 +1,254 @@
...
+      selectInput(ns("selected_plasma"), "Choose Plasma",
+                  choices = tissue_options)
+    })
+
+    # Filter & prepare data for tissue-plasma ratio calculation
+    filtered_samples <- reactive({
+      req(data(), input$selected_tissues)
+
+      plasma <- input$selected_plasma
+      tissue <- input$selected_tissues
```
m-kolomanski, line 198: issue: Looks like those variables are unused.

```suggestion
```

m-kolomanski (changes requested):
Hey, generally clean and very good implementation, nice work! I have some minor comments to the code, and some suggestions to consider:

- Leaving an empty space _for future_ is perfectly fine. One thing that I would maybe consider is adding some kind of placeholder message, like **To be added**, so that the card body is not empty. This informs the user that the feature is not implemented, rather than not showing because of some bugs.
- Consider maybe creating the Additional analysis as a separate tab completely, alongside NCA, instead of within in? That makes sense in my head, dunno if it is generally intuitive. 
- The currently implemented analyses look quite similar. Might be worth extracting that code into a module and re-using it. The module can accept some parameters, like the column names and labels and adjust the analysis for that.
- And/or some of that code might be worth extracting into functions and moved to `R/` folder. We could have functions like `bbp()` and `matrix_ratios()` available, so that if a user wants to build their own custom pipeline with our package they can, instead of relying on our application / interface.

Let me know what you think!

js3110:
> * Consider maybe creating the Additional analysis as a separate tab completely, alongside NCA, instead of within in? That makes sense in my head, dunno if it is generally intuitive.

I have been constantly rearranging this to try to figure out where it should go! On one hand I was thinking of making a 'ratios' section, but the excretion analysis won't be ratios. Then it was a 'non nca' section, but the bioavailability and auc ratios will require the NCA, so not sure if it should be outside of the NCA section. I will try to move it to alongside NCA and see how that looks. Since its modularised its easy to move so we can always change it later!

> * The currently implemented analyses look quite similar. Might be worth extracting that code into a module and re-using it. The module can accept some parameters, like the column names and labels and adjust the analysis for that.
> * And/or some of that code might be worth extracting into functions and moved to `R/` folder. We could have functions like `bbp()` and `matrix_ratios()` available, so that if a user wants to build their own custom pipeline with our package they can, instead of relying on our application / interface.

I agree, sounds like a good idea! I will try to work on it :) thanks!

Gero1999 (changes requested):
I really really like the interface layout, good job! I would only share some little aspects:

- [x] Suggestion: Consider simplifying the Blood-Plasma Partitioning and Matrix Ratios sections. It might be beneficial to retain only the Matrix Ratios and allow users to dynamically add more than one matrix ratio if needed. You could also use a textInput to let users label each dataframe, which could enhance clarity.

- [x] Feature Request: It would be helpful to add a download option for the results. The simplest approach might be to implement the download functionality we commonly use for DT-tables.

- [x] Minor Issue: Currently, users can select the same column for ratios, which leads to an unnotified error. This issue occurs in the Blood-Plasma Partitioning section but not in Matrix Ratios. If you decide to keep only Matrix Ratios, this problem should be resolved.

- [x] Future Consideration: Once #174 is merged, we should dynamically obtain categorical columns from mydata (STUDYID, USUBJID, DRUG, ANALYTE. AFRLT) at the left side of the ratios. Other columns might be displayed separately or omitted altogether.

Gero1999 (changes requested):
Hello @js3110, this looks really good! I added a dataset example to simplify the open source testing
 
 ```
 adnca <- read.csv("inst/shiny/data/DummyRO_Generated_with_metab.csv")
new.rows <- adnca %>%
  mutate(PCSPEC = "CSF",
         AVAL = AVAL - rnorm(nrow(adnca), 0.5, 0.1)
  )  %>%
  mutate(AVAL = ifelse(AVAL < 0, 0, AVAL),
         PCSTRESC = AVAL,
         AVALU = "ng/mL"
  )
adnca <- bind_rows(adnca, new.rows)
write.csv(adnca, "inst/shiny/data/Dummy_complex_data.csv", row.names = FALSE)
 ```

Gero1999 (approved):
lgtm

inst/shiny/modules/additional_analysis.R
```diff
@@ -0,0 +1,72 @@
...
+#' @param id A character string used to uniquely identify the module.
+#' @param data A PKNCA data object that returns the data list containing the concentration and
+#' dose data.
+#' @param grouping_vars A character vector of grouping variables to use for the analysis.
+
+# UI function for the non-nca analysis
+additional_analysis_ui <- function(id) {
+  ns <- NS(id)
+
+  tagList(
```
m-kolomanski, line 13: nitpick: `tagList` is not needed, as everything is wrapped in a single `navset_pill` anyway.

R/ratio_calculations.R
```diff
@@ -0,0 +1,74 @@
...
+                  as_units(Spec2_Units),
+                  mode = "character"),
+      Unit = as.character(units(Ratio)),
+      Unit = ifelse(Unit == "1", "unitless", Unit),
+      Ratio = signif(Ratio, 3)
+    ) %>%
+    filter(Spec1_Label != Spec2_Label) %>%
+    select(all_of(groups), Ratio_Type, Spec1_Value, Spec2_Value, Ratio, Unit)
+
+  df_ratio
```
m-kolomanski, line 73: nitpick: No need to return a variable, the function will return the result of the pipe automatically.
```suggestion
  left_join(df_spec1, df_spec2, by = groups, relationship = "many-to-many") %>%
    filter(!is.na(Spec1_Value) & !is.na(Spec2_Value)) %>%
    rowwise() %>%
    mutate(
      Ratio_Type = paste0(Spec1_Label, "/", Spec2_Label),
      Ratio =
        set_units(Spec1_Value,
                  as_units(Spec1_Units),
                  mode = "character") /
        set_units(Spec2_Value,
                  as_units(Spec2_Units),
                  mode = "character"),
      Unit = as.character(units(Ratio)),
      Unit = ifelse(Unit == "1", "unitless", Unit),
      Ratio = signif(Ratio, 3)
    ) %>%
    filter(Spec1_Label != Spec2_Label) %>%
    select(all_of(groups), Ratio_Type, Spec1_Value, Spec2_Value, Ratio, Unit)
```

m-kolomanski (approved):
Looks very good, I have very very minor nitpicks. Thanks for all the hard work!

PR #202, opened 2025-02-10
fix: descriptive statistics bug

### LINKED ISSUE #196: Bug: Descriptive statistics missing data
### Description
The descriptive statistics section, and also then boxplots are missing data points. For example, using the dummy data, there should be > 10 subjects but the n for Cmax and other parameters is only 3.

Results contain all subjects, so the issue must lie in the filtering and calculation of descriptive statistics.

### To reproduce
Run NCA and go to descriptive statistics. Notice the N values are very small.

### Expected behaviour
Descriptive statistics should be calculated including all relevant subjects.

Definition of Done:

- [ ] Descriptive statistics calculated correctly
- [ ] Stats and boxplots contain all relevant data points

js3110:
## Issue

Closes #196 

## Description

The descriptive statistics section, and also then boxplots are missing data points. For example, using the dummy data, there should be > 10 subjects but the n for Cmax and other parameters is only 3.

Results contain all subjects, so the issue must lie in the filtering and calculation of descriptive statistics.
## Definition of Done
- [x] Descriptive statistics calculated correctly
- [x] Stats and boxplots contain all relevant data points

## How to test

Run NCA -> Visualisation -> Descriptive statistics

## Contributor checklist
- [x] New logic is documented


Gero1999 (changes requested):
Hey! I will comment here the user product issues I think we need to improve:

- [x] Bug: Descriptive statistics summary not showing. `Error: [object Object]`
- [x] Issue: Eliminate one of the download buttons
- [ ] Suggestion: Instead of a bslib/default button use a DT button (similar to `Parameter Datasets`). I think can be more aesthetic!

m-kolomanski (approved):
Hey, the code looks good to me. I have two non-blocking comments.

Please do remember to increment the version before merging.

inst/shiny/modules/tab_visuals.R
```diff
@@ -435,45 +425,61 @@ tab_visuals_server <- function(id, data, grouping_vars, res_nca) {
...
     })
 
+    observeEvent(summary_stats(), {
+      req(summary_stats())
+      # Update the select display parameters picker input
+      updatePickerInput(
+        session,
+        "select_display_parameters",
+        choices = setdiff(colnames(summary_stats()), "Statistic"),
+        selected = setdiff(colnames(summary_stats()), "Statistic")
```
Gero1999, line 458: issue: groupby columns are already in another input selected, should not be in select_display_parameters

```suggestion
        choices = setdiff(colnames(summary_stats()), c("Statistic", input$summary_groupby)),
        selected = setdiff(colnames(summary_stats()), c("Statistic", input$summary_groupby))
```

inst/shiny/modules/tab_visuals.R
```diff
@@ -435,45 +425,61 @@ tab_visuals_server <- function(id, data, grouping_vars, res_nca) {
...
+    })
+
+
     # render the reactive summary table in a data table
     output$descriptive_stats <- renderReactable({
       req(summary_stats())
       log_info("Rendering descriptive statistics table")
 
+      data <- summary_stats() %>%
+        select(input$select_display_parameters)
```
Gero1999, line 469: issue: missing "Statistic" column display (and groupby vars before)

```suggestion
        select(any_of(c(input$summary_groupby, "Statistic")), input$select_display_parameters)
```

inst/shiny/modules/tab_visuals.R
```diff
@@ -435,45 +425,61 @@ tab_visuals_server <- function(id, data, grouping_vars, res_nca) {
 
     # Reactive expression for summary table based on selected group and parameters
     summary_stats <- reactive({
-      req(input$summary_groupby, input$select_display_parameters)
+      req(input$summary_groupby)
```
Gero1999, line 428: suggestion: if there are no groupby variables selected, we can assume the user wants the raw statistics without grouping. I think we should let them have that possibility

```suggestion
```

Gero1999 (changes requested):
Hey I just found some easy to fix issues in the display of the table. After solving those I reviewed the results and they seem good to me, good job! 😉

Gero1999 (approved):

PR #215, opened 2025-02-24
bug: start and end time inconsistent and duplicates missing

### LINKED ISSUE #214: Bug: start and end time calculation inconsistencies
### Description
During development, the derivation of start and end time has swapped occasionally between being the time from first dose, or the time from most recent dose. Due to this, there are some sections later on in the code that are inconsistent- ie some parts assume that it is time from first dose, and others that it is time from mrd. This then causes inaccuracies in the data, for example lambda.z.ix calculations for the slopes selectors.

We need to go through every instance in the code where 'start' or 'end' have been used, and try to ensure that the code matches the time variable. At the moment, start and end are from first dose before the nca, and from recent dose afterwards.

For example, in pivot_wider_pknca_results:

`infinite_aucs_with_lambda <- inner_join(myres$data$conc$data, infinite_aucs) %>%
    group_by(STUDYID, PCSPEC, ANALYTE, USUBJID, DOSNO) %>%
    arrange(STUDYID, PCSPEC, ANALYTE, USUBJID, DOSNO, IX) %>%
    mutate(lambda.z.method = ifelse(
      any(is.excluded.hl) | any(is.included.hl), "Manual", "Best slope"
    )) %>%
    filter(!exclude_half.life | is.na(lambda.z.time.first) | is.na(lambda.z.n.points)) %>%
    filter(TIME >= (lambda.z.time.first + start) | is.na(lambda.z.time.first)) %>%
    filter(row_number() <= lambda.z.n.points | is.na(lambda.z.n.points)) %>%
    mutate(lambda.z.ix = paste0(IX, collapse = ",")) %>%
    mutate(lambda.z.ix = ifelse(is.na(lambda.z), NA, lambda.z.ix)) %>%
    slice(1) %>%
    select(any_of(c(names(infinite_aucs), "lambda.z.method", "lambda.z.ix")))`

Here, in myres, start and end are from most recent dose, however we are filtering using TIME (from first dose). This means that the calculation of lambda.z.ix will only work for the first dose, and not subsequent doses.

### Expected behaviour
Before nca, start and end are time from first dose. After NCA, they are time from recent dose.
Ensure that any calculations involving start and end time reflect this.

js3110:
## Issue

Closes #214 

## Description

During development, the derivation of start and end time has swapped occasionally between being the time from first dose, or the time from most recent dose. Due to this, there are some sections later on in the code that are inconsistent- ie some parts assume that it is time from first dose, and others that it is time from mrd. 
Lambda.z.ix is also not correct due to the start times and also the duplicates are missing.

## Definition of Done

- [x] start and end time used as correct time for calculations
- [x] Plots show the correct time, and lambda.z.ix correctly calculated

## Contributor checklist
- [x] New logic is documented

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.0.0.9010
+Version: 0.0.0.9011
 Authors@R: c(
     person("Ercan", "Suekuer", , "ercan.suekuer@roche.com", role = c("aut", "cre"),
            comment = c(ORCID = "0009-0001-1626-1526")),
--- a/R/lambda_slope_plot.R
+++ b/R/lambda_slope_plot.R
@@ -72,11 +72,11 @@ lambda_slope_plot <- function(
       if_all(all_of(column_names), ~ .x == row_values[[deparse(substitute(.x))]]),
       !exclude_half.life,
       TIME >= sum(
-          subset(
-            lambda_res,
-            lambda_res$PPTESTCD == "lambda.z.time.first",
-            select = c("start", "PPSTRES")
-          )
+        subset(
+          lambda_res,
+          lambda_res$PPTESTCD == "lambda.z.time.first",
+          select = c("start", "PPSTRES")
+        )
       )
     ) %>%
     arrange(IX) %>%
--- a/R/lambda_slope_plot.R
+++ b/R/lambda_slope_plot.R
@@ -58,7 +58,7 @@ lambda_slope_plot <- function(
   #Obtain values for slopes selection
   lambda_res <- myres$result %>%
     filter(if_all(all_of(column_names), ~ .x == row_values[[deparse(substitute(.x))]])) %>%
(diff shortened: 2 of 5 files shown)


m-kolomanski (approved):
LGTM.

Yes, the package version should be bumped with every change to the code.

This might be skipped when PR only involves some work outside of the package and app itself (CI updates, vignettes/articles, README or CONTRIBUTING updates etc).

Gero1999 (changes requested):
hey @js3110 thanks for checking this, just a comment on it

js3110:
@Gero1999 as discussed I changed it so that the results show start and end time from first dose.
I created temporary variables start_dose and end_dose for the parts of the calculations that require it to be from recent dose.

I also modified the slope plots so they show the time from recent dose, as for the exploration and manual slopes selection this is better as they can directly see tmax and lambda.z.time.first without having to do any mental maths.

I hope this makes sense to you now !

@m-kolomanski if you could quickly re review as well, just for safety 😄

m-kolomanski (approved):
TMLG!

Gero1999 (approved):

PR #220, opened 2025-02-25
Bug: Cmax in NCA Calculations

### LINKED ISSUE #148: Bug: Manual slope selection and automatic PKNCA behaviour conflicts
### Description

We would like that PKNCA's automatic behaviour does not include never automatically `tmax` for the half life calculation and that it takes 3 points always... This would be:

```
    # Use the user inputs to determine the NCA settings to apply
    PKNCA::PKNCA.options(
      auc.method = input$method,
      allow.tmax.in.half.life = FALSE,

      # Make sure the standard options do not prohibit results
      min.hl.r.squared = 0.001,
      min.span.ratio = Inf,
      min.hl.points = 3
    )
```

> **Note:** Other potential additional interests could be as well: using the min.hl.r.adj (still not possible in PKNCA) and the min.span.ratio defined by the users. 

However, using this PLNCA.options can also affect the behaviour of the manual configuration done in the `slope_selector` module. For that reason we would like a solution that makes both fit together

An initial proposed easy-fix would be to only compute these PKNCA.options for "best_slopes" while on "manual_slopes" that are selected by the user perform other more permissive PKNCA.options:

```
    # Use the user inputs to determine the NCA settings to apply
    PKNCA::PKNCA.options(
      auc.method = input$method,
      allow.tmax.in.half.life = TRUE,

      # Make sure the standard options do not prohibit results
      min.hl.r.squared = 0.001,
      min.span.ratio = Inf,
      min.hl.points = 2
    )
```

### To reproduce

1) Being in code min.hl.points = 2  
Map data > Submit analyte/dosno > Run NCA > In slope selector try to choose only 2 points for your slope

2) Being in code allow.tmax.in.half.life = FALSE
Map data > Submit analyte/dosno > Run NCA > In slope selector try to choose only 2 points for your slope

### Expected behaviour
That the user can choose any point in the slope selector while the automatic selection in the app does exclude tmax and chooses always 3 points.

js3110:
## Issue

Closes #148 

## Description

We would like that PKNCA's automatic behaviour does not include never automatically tmax for the half life calculation and that it takes 3 points always.

## Definition of Done

- [x] Tmax automatically excluded from half life calculation. and 3 points minimum
- [x] Manual slope selection overrides this setting

## How to test

Check that Tmax not included in automatic slopes
Add manual slopes- include Tmax, and only select 2 points.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer
This was actually an easy fix, as the PKNCA options default settings are to exclude tmax and to have min 3 points in hl calculation. And if you don't explicitly write it (so remove from PKNCAoptions), then any manual slopes will override this for those subjects.

Important to note that if you are testing the Tmax, it wont show up in the plot (because of issue #121 ), so best to check the results and see if labda.z.ix has been updated, and lambda.z.time.first.

Also, the change to pivot_wider_pknca is just temporary, to ensure that lambda.z.ix was correct. This will be updated when #214 has been merged!

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.0.0.9010
+Version: 0.0.0.9011
 Authors@R: c(
     person("Ercan", "Suekuer", , "ercan.suekuer@roche.com", role = c("aut", "cre"),
            comment = c(ORCID = "0009-0001-1626-1526")),
--- a/R/pivot_wider_pknca_results.R
+++ b/R/pivot_wider_pknca_results.R
@@ -52,7 +52,7 @@ pivot_wider_pknca_results <- function(myres) {
     # filter out the rows that do not have relation with lambda calculation (when calculated)
     # and derive the IX
     filter(!exclude_half.life | is.na(lambda.z.time.first) | is.na(lambda.z.n.points)) %>%
-    filter(TIME >= (lambda.z.time.first + start) | is.na(lambda.z.time.first)) %>%
+    filter(ARRLT >= (lambda.z.time.first + start) | is.na(lambda.z.time.first)) %>%
     filter(row_number() <= lambda.z.n.points | is.na(lambda.z.n.points)) %>%
     mutate(lambda.z.ix = paste0(IX, collapse = ",")) %>%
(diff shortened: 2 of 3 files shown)


m-kolomanski (approved):
LGTM

Gero1999 (approved):
lgtm

PR #222, opened 2025-02-25
Bug: units table multiple analyte display & results only working for all analytes with same units

### LINKED ISSUE #218: Bug: Units table for multiple analyte is duplicated
### Description
When running a multiple analyte study the table displayed to indicate the units is unnecessarily repeated, which can be confusing for the user.  

### To reproduce
Dummy_data_complex (multiple analytes) >> Map data >> NCA settings >> Units table

### Expected behaviour
- [ ] Units table should only have one row per parameter

Gero1999:
## Issue

Closes #218 

## Description

When running a multiple analyte study the table displayed to indicate the units is unnecessarily repeated, which can be confusing for the user.

## Definition of Done
- [x] Units table is not replicated per analyte when selecting both in the module message "Units Table"
- [x] Using different units per analyte will show up in results accordingly

## How to test
Map data > NCA setup > Select all analytes in input > Units Table button
Then play with it to see it works and does not crush easily. After that
Run NCA > Check results

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

### DIFF
--- a/R/pivot_wider_pknca_results.R
+++ b/R/pivot_wider_pknca_results.R
@@ -62,7 +62,7 @@ pivot_wider_pknca_results <- function(myres) {
     select(-PPSTRES, -PPSTRESU, -PPORRES, -PPORRESU, -type_interval)  %>%
     pivot_wider(names_from = PPTESTCD, values_from = exclude, names_prefix = "exclude.")
 
-  infinite_aucs <- inner_join(infinite_aucs_vals, infinite_aucs_exclude)
+  infinite_aucs <- left_join(infinite_aucs_vals, infinite_aucs_exclude)
 
   # If there were intervals defined, make independent columns for each
   if (any(myres$result$type_interval == "manual")) {
--- a/R/pivot_wider_pknca_results.R
+++ b/R/pivot_wider_pknca_results.R
@@ -19,6 +19,34 @@
 #'
 pivot_wider_pknca_results <- function(myres) {
 
+  ################################################################################################################## 
+  # Derive lambda.z.n.points & lambda.z.method
+  # ToDo: At some point this will be integrated in PKNCA and will need to be removed//modified
+  added_params <- NULL
+  if (all(c("lambda.z", "lambda.z.n.points", "lambda.z.time.first") %in% unique(myres$result$PPTESTCD))) {
+    added_params <- myres$result %>%
+      filter(PPTESTCD %in% c("lambda.z.n.points", "lambda.z.time.first", "lambda.z"),
+             type_interval == "main") %>%
+      select(unname(unlist(myres$data$conc$columns$groups)), PPTESTCD, PPSTRES, DOSNO, start, end) %>%
+      unique() %>%
+      pivot_wider(names_from = PPTESTCD, values_from = PPSTRES) %>%
+      left_join(myres$data$conc$data) %>%
+      # Derive lambda.z.method: was lambda.z manually customized?
+      mutate(lambda.z.method = ifelse(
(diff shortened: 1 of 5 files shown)


R/pivot_wider_pknca_results.R
```diff
@@ -19,50 +19,59 @@
...
+  # Derive lambda.z.n.points & lambda.z.method
+  # ToDo: At some point this will be integrated in PKNCA and will need to be removed//modified
+  added_params <- NULL
+  if (all(c("lambda.z",
+            "lambda.z.n.points",
+            "lambda.z.time.first") %in% unique(myres$result$PPTESTCD))) {
+    added_params <- myres$result %>%
+      filter(PPTESTCD %in% c("lambda.z.n.points", "lambda.z.time.first", "lambda.z"),
+             type_interval == "main") %>%
+      select(unname(unlist(myres$data$conc$columns$groups)),
```
m-kolomanski, line 32: nitpick: might be worth assigning this to a variable, since it is repeated a lot of times
```R
col_groups <- unname(unlist(myres$data$conc$columns$groups)
```
or something like this.

m-kolomanski (commented):
The units table still seems duplicated for me:
Screenshot from 2025-02-26 10-15-37

inst/shiny/modules/units_table.R
```diff
@@ -77,11 +77,22 @@ units_table_server <- function(id, mydata, res_nca = reactiveVal(NULL)) {
...
+    # Create a function to provide a clean display of the modal units table
+    clean_display_units_table <- function(data, selected_analytes) {
+      data %>%
+        mutate(`Conversion Factor` = signif(`Conversion Factor`, 3)) %>%
+        filter(`Analytes` %in% selected_analytes) %>%
+        group_by(Parameter, `Default unit`, `Conversion Factor`, `Custom unit`) %>%
+        mutate(Analytes = paste(Analytes, collapse = ", ")) %>%
+        ungroup() %>%
+        unique()
+    }
```
m-kolomanski, line 89: issue: This function should be outside of the `moduleServer`. Since it is used only within this module, it could be within the same file, just have a `.` in the name to signify it is a local function. Or,  it could be moved to `functions/` folder.

m-kolomanski (changes requested):
issue: Even if only one of the analytes is selected in the settings, the units table will still propose both. This might be a bit confusing.
[Screencast from 2025-02-27 08-19-39.webm](https://github.com/user-attachments/assets/f639fe34-141f-400a-8960-999378740609)

m-kolomanski (approved):
Thank you for the changes, looks good!

js3110 (approved):
LGTM!

R/pivot_wider_pknca_results.R
```diff
@@ -19,57 +19,63 @@
...
+  # ToDo: At some point this will be integrated in PKNCA and will need to be removed//modified
+  conc_groups <- unname(unlist(myres$data$conc$columns$groups))
 
-  # Create duplicates for accurate lambda.z.ix calculation
   data_with_duplicates <- dose_profile_duplicates(
     myres$data$conc$data,
     c(unlist(unname(myres$data$conc$columns$groups)),
       "DOSNO")
   )
+  added_params <- NULL
```
js3110, line 32: @Gero1999 what are these changes for? how are they related to units?
and lambda.z.ix is incorrect in this branch, but correct in main so maybe its better to restore the changes, except the ones related to units. 
as lambda.z.ix was fixed in a different issue that has since been merged.

js3110: actually nvm i checked main and the issues are also there. I will create a separate issue to address this!

js3110 (approved):

Issue #226, opened 2025-02-26
Bug: TLG plots fail with an error: `All unnamed arguments must be length 1`


m-kolomanski:
### Description
Graph rendering fails with error:
```
ERROR [2025-02-26 14:33:47] Error in plot rendering:
<simpleError: All unnamed arguments must be length 1>
```

This is unrelated to TLG listings or the refactoring of the TLG system - it exist at the moment (2025-02-26) on `main` branch, before any of the above mentioned changes are introduced and merged

### To reproduce
Run the app, map the data, submit a default TLG order and go to Graphs. The plot rendering will fail.

### Expected behaviour
The plots should render corretly

PR #216, opened 2025-02-24
Enhancement: setup system for tlg listings

m-kolomanski:
## Issue
Closes #186 

## Description
Introduces separate module for rendering and customizing TLG listings. Adds minor refactoring of the TLG modules and relevant functions. Adds definition of the first listing in the `yaml` file.

#### Global aspects
- [x] When user presses `Submit TLG order` listings are displayed and classified in the tab `Listings`. Representing the `rlisting` object that will (in the future) be included in the PDF report with the function `export_as_pdf`

#### General
- [x] Grouping tables: by default = c(“PARAM”, “PCSPEC”, “ROUTE”)
- [x] Grouping variables: SelectInput multiple choice (index columns) by default(c("TRT01A" or “TRTA”, "USUBJID", "AVISIT"))
- [x] Reported variables: Select input with relevant column names (default: NFRLT AFRLT, AVALC) (by default NFRLT AFRLT and AVALC)

## How to test
Run the app, map the data. You do not need to run NCA analysis, you can go straight to TLG tab. Submit the default TLG order. Make sure the listing renders. Check if customizing widgets work and the listing is updated.

## Contributor checklist
- [x] Code passes lintr checks
- [x] Code passes all unit tests
- [x] New logic covered by unit tests
- [x] New logic is documented
- [x] Package version is incremented

## Notes to reviewer
One of the features from the task description is not included. The listing function requires passing a table with formatting data for each relevant variable. This is not currently doable within the current system. It will require some refactoring to support such custom inputs. Hence the listing support will be added in a basic form and expanded later once the TLG system is updated.

The progress for the feature is explained in a [comment in the original issue](https://github.com/pharmaverse/aNCA/issues/186#issuecomment-2682662089)

Issue #186, opened 2025-01-31
Enhancement: Setup system for TLG (Listings)


Gero1999:
### Description
In the TLG tab section we already have incorporated a functional subtab Graph with a set of inputs that allow the user to visualize and customize their plots. Now we need a similar setup for the listings. Now we want a similar setup using the function in #179. The inputs expected in the TLG Listings section are the ones defined below:

#### Global aspects
- [ ] When user presses `Submit TLG order` listings are displayed and classified in the tab `Listings`. Representing the `rlisting` object that will (in the future) be included in the PDF report with the function `export_as_pdf`

#### General
- [ ] Grouping tables: by default = c(“PARAM”, “PCSPEC”, “ROUTE”)
- [ ] Grouping variables: SelectInput multiple choice (index columns) by default(c("TRT01A" or “TRTA”, "USUBJID", "AVISIT"))
- [ ] Grouping variables: SelectInput multiple choice (index columns) by default(c("TRT01A" or “TRTA”, "USUBJID", "AVISIT"))
- [ ] Reported variables: Select input with relevant column names (default: NFRLT AFRLT, AVALC) (by default NFRLT AFRLT and AVALC)

#### Formatting table input
A dataframe (or many) used as an input. It may have one row per variable, and the next columns:
- [ ] Variable name: Defined with Reported variables select input. No modifiable. 
- [ ] Label: Text input default populated with $Variable `+ ($unit) if relevant`
- [ ] cero_str: Text input for 0 values (by default NA)
- [ ] na_str: Text input for NA values (by default "NA") 💡 
- [ ] format_fun: character name of the function to use to format the digits ("round", "signif"). Always NA in non-numeric variables 💡 
- [ ] digits: numeric to use in the argument of the previous function 💡 
- [ ] align: how the extension of the variable value will be aligned on the listing displayed (left, center, ) 💡💡 >> `This could also even be in General`
- [X] ~unit: character name of the column to use, It will be added to the label in parenthesis~ >> `Already defined in label, so no needed`

m-kolomanski:
Hi @Gero1999 ! I have a question in regards to the task: Do all of the mentioned variables need to be customizable by the user, including providing the formatting table? If yes, do we have any specific design we would like to go for in terms of UI/UX? Seems like a quite dense feature and I wanna make sure what would be the vision here before committing.

You can see what the current implementation is like in #216, but it is barebones at the moment.

Gero1999:
[...]

PR #227, opened 2025-02-27
Refactor/tlg system update

### LINKED ISSUE #223: Refactor: TLG system
### Description
The current usage of basic shiny inputs does not provide enough flexibility to facilitate the customization options that we wish to achieve.

Furthermore, module for each TLG will end up very similar and those could be abstracted into a single `tlg_module`.

### Definition of Done
- [ ] refactor `tlg_plot` and `tlg_list` into a single module `tlg_module`
- [ ] wrap default shiny inputs into separate modules that are able to handle more advanced logic
- [ ] refactor the options system to work with `reactive`s returned by the modules, as opposed to fetching the customized values from the `input` object.

m-kolomanski:
## Prerequisites
Requires #216 to be merged.

## Issue
Closes #223 

## Description
Refactors the TLG modules in order to enable creation of more customizable editing widgets. The reason being that vanilla shiny inputs were not enough to pass required amount of data to the implementation functions for graphs, listings and tables.

#### Single module to rule them all
Instead of creating separate module for each TLG type (graph, listing, table) there is a single `tlg_module.R`, which takes `type` as an argument and handles rendering appropriately.

#### Input modules
Instead of creating shiny inputs for entry customization, each input type is now a separate module, currently: `tlg_option_text.R`, `tlg_option_numeric.R` and `tlg_option_select.R`.

This might seem like an overkill for the current implementation, but will be required for planned features, in example:
- We plan on implementing a input table, that will allow the user to specify various formatting characteristics across multiple variables (see the implementation of `l_pkconc` function).
- This change will allow us to shift some certain tasks to the input modules, as opposed to the implementation functions. Currently, the user can use special syntax that is parsed against the data (like e.g. providing `$DOSEU` in a label field will display the actual unit that is specified within the data). This is done in functions that return plots/listings, hence every future implementation would need to parse these strings. Instead, the input modules could do that and return already parsed label string - thus making implementing new TLGs a lot less hassle.

## Definition of Done
- [x] refactor `tlg_plot` and `tlg_list` into a single module `tlg_module`
- [x] wrap default shiny inputs into separate modules that are able to handle more advanced logic
- [x] refactor the options system to work with `reactive`s returned by the modules, as opposed to fetching the customized values from the `input` object.

## How to test
Run the app, map the data. Submit TLG order. Go into **Listings** and check if all features on the page work. Please check the **Notes to reviewer** for further info.

## Contributor checklist

## Notes to reviewer
Changes applied are pure refactoring, they should not have any effect on user experience.

The changes should be tested on the **Listings** tab. There is a [known bug with plots rendering](https://github.com/pharmaverse/aNCA/issues/226), which was detected before any of the changes to the TLG system were applied. This bug should be fixed separately, ideally after this PR is merged, to account for any unforeseen consequences the refactoring might have.


inst/shiny/modules/tlg/tlg_option_numeric.R
```diff
@@ -0,0 +1,19 @@
...
+  label <- if (is.null(opt_def$label)) id else opt_def$label
+
+  numericInput(
+    ns("numeric"),
+    label = label,
+    value = opt_def$default
+  )
+}
+
+tlg_option_numeric_server <- function(id, opt_def, data) {
```
Gotfrid, line 13: discussion: here and in other functions - unused arguments. is this for compatibility with existing code?

inst/shiny/modules/tlg/tlg_option_numeric.R
```diff
@@ -0,0 +1,19 @@
+tlg_option_numeric_ui <- function(id, opt_def, data) {
```
Gotfrid, line 1: suggestion: here and in other similar functions - do you think it would be useful to document the parameters?

Gotfrid (approved):
Looks good, left a couple of comments that are non blocking

inst/shiny/modules/tlg/tlg_option_numeric.R
m-kolomanski: Kinda, I wanted the boilerplate for each `tlg_option_*` module to be the same. Then the function that calls the ui/server functions can just pass the same arguments and not worry about it. I know I could have used `...`, but wanted the boilerplate to be the same. Plus, those arguments will be used in future features that are planned, so I wanted to get ahead a bit to have clear structure right now.

I do agree with your other comment though, I will add some docs!

Gero1999 (approved):
Hey @m-kolomanski, thanks for the changes all looks good! Only thing which I am not sure if it is new is that the name of the list object appears in the UI (i.e, `$SERUM.intravascular.Analyte01`). Is it an easy fix? If not we can do it in a different issue

I don't think is a big issue though, so I will approve anyway.

m-kolomanski:
@Gero1999 Hmm, that should have been fixed with #186. I have added it as a bug report #228  and will fix it.

PR #235, opened 2025-02-27
GUI Enhancement: descriptive statistics module

### LINKED ISSUE #234: GUI: move descriptive statistics into NCA results
As these are not 'visualisation' but rather results, they should be in results section

js3110:
## Issue

Closes #234 

## Description

Move descriptive statistics location and create a module. In addition, change the default grouping options

## Definition of Done

Descriptive statistics works and is moved to NCA results

## Contributor checklist
- [x] New logic is documented

### DIFF
--- a/inst/shiny/modules/descriptive_statistics.R
+++ b/inst/shiny/modules/descriptive_statistics.R
@@ -7,7 +7,7 @@
 #' @param res_nca A reactive expression that returns the NCA results.
 #' @param grouping_vars A reactive expression that returns the grouping variables.
 #'
-#' @return A list containing the reactive expression for the summary statistics table.
+#' @returns A list containing the reactive expression for the summary statistics table.
 
 # UI function for the summary statistics module
 descriptive_statistics_ui <- function(id) {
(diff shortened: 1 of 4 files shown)


inst/shiny/modules/descriptive_statistics.R
```diff
@@ -0,0 +1,144 @@
...
+      cols_to_join <- c(classification_cols, unname(unlist(res_nca()$data$conc$columns$groups)))
+      stats_data <- inner_join(
+        res_nca()$result,
+        select(res_nca()$data$conc$data, any_of(cols_to_join))
+      ) %>%
+        filter(!(type_interval == "manual" & PPTESTCD != "aucint.last")) %>%
+        mutate(PPTESTCD = case_when(
+          type_interval == "manual" ~ paste0(PPTESTCD, signif(start), "-", signif(end)),
+          TRUE ~ PPTESTCD
+        ))
```
m-kolomanski, line 90: nitpick: Since we have only two options, `ifelse` would make more sense, unless we are future-proofing for something?

inst/shiny/modules/descriptive_statistics.R
```diff
@@ -0,0 +1,144 @@
...
+      },
+      content = function(file) {
+        log_info("Downloading summary statistics as CSV")
+        write.csv(summary_stats(), file)
+      }
+    )
+    
+    list(
+      summary_stats = summary_stats
+    )
```
m-kolomanski, line 142: questions:
1. Do we need to return a list? Seems like returning just `summary_stats` should be enough?
2. Seems like we are not capturing the returned reactive anyway inm `tab_nca.R`, do we need to return anything at all?

m-kolomanski (approved):
Hey, good PR, I like further modularization of application components.

I am accepting the changes, but leaving some non-blocking comments.

Also, please make sure that the linting passes before merging.

Gero1999 (approved):

PR #237, opened 2025-02-28
Bug: no impute cmax tmax

### LINKED ISSUE #182: Bug: PKNCA imputations do not disconsider observed parameters
### Description
There are parameters (i.e, `cmax`, `tmax`) that should be based merely on observations, and not imputed data. At first we though that this was controlled by PKNCA, but it is not, so it is a bug that eventually will need to be solved by PKNCA. In the meantime, as a temporary measure we will develop a solution for this problem that is external to the package. 

> Suggested solution 
Make a function that only and simply modifies the PKNCAdata$intervals, i.e, remove_impute_for_params(PKNCAdata, params). The function identifies rows/intervals where param = TRUE & !is.na(impute), set in those param = FALSE and create duplicates where param = TRUE, impute = NA

### To reproduce
Run NCA on IV bolus data in both conditions: considering t0 and not considering it. You can see that, erroneously, C0 and t0 values change in each run!

### Expected behaviour
t0 and c0 should in any case change when imputations are added

Gero1999:
## Issue

Closes #182 

## Description

`cmax` and `tmax` variables should not be suffering any imputation unlike the other parameters. This can be very dangerous in IV bolus studies (`ROUTE = "intravascular", ADOSEDUR = 0`)

For now, automatically, the app will exclude the imputations from this parameters by making additional interval rows for them.

## Definition of Done
- When t0 is checked in `NCA setup`, all parameters but tmax and cmax are imputed.

## How to test

`Map data > NCA setup: Select at least 1-3 doses > Run NCA > Check results & compare with data`
 
- [x] The `mydata()$intervals %>% filter(cmax | tmax) %>% pull(impute)` is all NA or empty string   
- [x] The results for cmax and tmax in the IV bolus subjects (11104, 11105, 11106) does not change when t0 is checked, but the aucinf values do. 
- [x] cmax expected values for 11104: `3.34`, `9.86`, `5.98` (DOSNO 1-3)
- [x] tmax is never 0

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer
For other specific value checking you can use this code:
```
adnca %>% dplyr::group_by(USUBJID, DOSNO, ROUTE, ANALYTE) %>% summarise(cmax = max(AVAL)) %>% filter(USUBJID == c("11104", "11105", "11106"))
```

Also notice that at some point the plan is to implement a better interface to represent/modify the imputations done in the App in


R/intervals_support_funs.R
```diff
@@ -0,0 +1,363 @@
...
+  )
+  ## 3. The target impute method is not already present and correctly positioned
+  after_vals <- lapply(strsplit(data$impute, "[ ,]+"), function(x) {
+    after_x <- which(x == target_impute)
+    if (length(after_x) == 0) return(NA)
+    if (after_x == length(x)) Inf else after_x
+  }) |> unlist()
+  target_rows <- target_rows & (after_vals != after | is.na(after_vals))
+
+  new_intervals <- data[target_rows, ]
```
m-kolomanski, line 143: suggestion: There is nothing inherently wrong with this code, but I find it a bit hard to read - it took me a while to deduce exactly what was going on.

My approach would look like so:
```R
  # Identify the target interval rows based on:
  ## 1. The target groups (perfect match)
  is_target_group <- {
    if (!is.null(target_groups)) {
      sapply(data[, names(target_groups), drop = FALSE], paste0) %in% sapply(target_groups, paste0)
    } else {
      rep(TRUE, nrow(data))
    }
  }

  ## 2. The target parameters (at least one calculated: not-FALSE/not-NA)
  target_params_data <- data[, target_params, drop = FALSE]
  is_target_param <- rowSums(replace(target_params_data, is.na(target_params_data), FALSE)) > 0

  ## 3. The target impute method is not already present and correctly positioned
  after_vals <- sapply(strsplit(data$impute, "[ ,]+"), \(x) {
    after_x <- which(x == target_impute)
    if (length(after_x) == 0) return(TRUE)
    if (after_x == length(x)) Inf else after_x
  })
  is_after <- after_vals != after | is.na(after_vals)

  target_rows <- is_target_group & is_target_param & is_after
  new_intervals <- data[target_rows, ]
```
This works the same based on the example code and passes unit tests.

This is just a suggestion, apply at your own convenience. If you fell like my approach is less readable - then stay with yours. Or apply parts of it if you like. Or we can see what other reviewers think.

R/intervals_support_funs.R
```diff
@@ -0,0 +1,363 @@
...
+
+  # Remove the target parameters calculation from the original target intervals
+  data[target_rows, target_params] <- FALSE
+
+  # Combine the new and original intervals
+  data <- rbind(data, new_intervals)
+
+  # Filter rows where all row values for param_cols are NA or FALSE
+  param_data <- data[, param_cols, drop = FALSE]
+  rows_no_params <- rowSums(!is.na(replace(param_data, param_data == FALSE, NA))) == 0
```
m-kolomanski, line 169: suggestion: Similar as in previous suggestion:
```suggestion
  rows_no_params <- rowSums(replace(param_data, is.na(param_data), FALSE)) == 0
```

R/intervals_support_funs.R
```diff
@@ -0,0 +1,363 @@
...
+      if (is.null(target_params) && is.null(target_groups)) {
+        # b. If it is in the global impute and no target parameters or groups, remove global impute
+        data$impute <- remove_impute_method(data$impute, target_impute)
+        return(data)
+      } else {
+        # c. If it is in the global impute but target parameters or groups, creates an impute column
+        data$intervals$impute <- data$impute
+        data$impute <- NA_character_
+      }
+    }
```
m-kolomanski, line 245: suggestion: We have some nice early returns, I think we can scrap some of the `if else` statements.
```suggestion
  # If the impute column is not present in the intervals...
  if (!"impute" %in% colnames(data$intervals)) {
    if (is.null(data$impute) || is.na(data$impute)) {
      # a. If neither is in the global impute method, return the data as it is
      warning("No default impute column or global method identified. No impute methods to remove")
      return(data)
    }

    # b & c. If global impute exists
    if (is.null(target_params) && is.null(target_groups)) {
      # b. If it is in the global impute and no target parameters or groups, remove global impute
      data$impute <- remove_impute_method(data$impute, target_impute)
      return(data)
    }

    # c. If it is in the global impute but target parameters or groups, creates an impute column
    data$intervals$impute <- data$impute
    data$impute <- NA_character_
  }
```

R/intervals_support_funs.R
```diff
@@ -0,0 +1,363 @@
...
+    )
+  }
+  ## 2. The target parameters (at least one calculated: not-FALSE/not-NA)
+  target_params_data <- data[, target_params, drop = FALSE]
+  target_rows <- target_rows & (
+    rowSums(!is.na(replace(target_params_data, target_params_data == FALSE, NA))) > 0
+  )
+  ## 3. The target impute method to be removed (contained in the string)
+  target_rows <- target_rows & grepl(target_impute, data$impute, fixed = TRUE)
+  new_intervals <- data[target_rows, ]
```
m-kolomanski, line 330: suggestion: I would implement similar workflow to my suggestion for similar code in `interval_add_impute.data.frame `. Again, subjective.

R/intervals_support_funs.R
```diff
@@ -0,0 +1,363 @@
...
+
+  # Remove the target parameters calculation from the original target intervals
+  data[target_rows, target_params] <- FALSE
+
+  # Combine the new and original intervals
+  data <- rbind(data, new_intervals)
+
+  # Filter rows where all row values for param_cols are NA or FALSE
+  param_data <- data[, param_cols, drop = FALSE]
+  rows_no_params <- rowSums(!is.na(replace(param_data, param_data == FALSE, NA))) == 0
```
m-kolomanski, line 356: suggestion: I would implement similar workflow to my suggestion for similar code in `interval_add_impute.data.frame `. Again, subjective.

R/intervals_support_funs.R
```diff
@@ -0,0 +1,363 @@
...
+
+  # Remove the impute from the other methods in each value
+  impute_vals <- ifelse(is.na(impute_vals), "", impute_vals)
+  strsplit(impute_vals, split = "[ ,]+") |>
+    lapply(FUN = setdiff, target_impute) |>
+    lapply(FUN = append, values = target_impute, after = after) |>
+    vapply(FUN = paste, collapse = ",", FUN.VALUE = "")
+}
+
+#' @export
```
m-kolomanski, line 88: issue: No need to export this function, I would use `@keywords internal` instead.\
```suggestion
#' @keywords internal
```

R/intervals_support_funs.R
```diff
@@ -0,0 +1,363 @@
...
+#' o_data <- interval_remove_impute(data = o_data,
+#'                                  target_impute = "start_conc0",
+#'                                  target_params = "half.life",
+#'                                  target_groups = data.frame(analyte = "Analyte1"))
+#' @export
+interval_remove_impute <- function(data, ...) {
+  UseMethod("interval_remove_impute", data)
+}
+
+#' @export
```
m-kolomanski, line 223: issue: No need to export this function, I would use `@keywords internal` instead.
```suggestion
#' @keywords internal
```

R/intervals_support_funs.R
```diff
@@ -0,0 +1,363 @@
...
+  impute_vals <- ifelse(is.na(impute_vals), "", impute_vals)
+  impute_vals <- strsplit(impute_vals, split = "[ ,]+") |>
+    lapply(FUN = setdiff, target_impute) |>
+    vapply(FUN = paste, collapse = ",", FUN.VALUE = "")
+
+  # Replace empty strings with NA_character_
+  ifelse(impute_vals == "", NA_character_, impute_vals)
+}
+
+#' @export
```
m-kolomanski, line 277: issue: No need to export this function, I would use `@keywords internal` instead.
```suggestion
#' @keywords internal
```

R/intervals_support_funs.R
```diff
@@ -0,0 +1,363 @@
...
+#' o_data <- interval_add_impute(o_data,
+#'                               target_impute = "start_conc0",
+#'                               target_params = "half.life",
+#'                               target_groups = data.frame(analyte = "Analyte1"))
+#' @export
+interval_add_impute <- function(data, ...) {
+  UseMethod("interval_add_impute", data)
+}
+
+#' @export
```
m-kolomanski, line 54: issue: No need to export this function, I would use `@keywords internal` instead.
```suggestion
#' @keywords internal
```

R/intervals_support_funs.R
m-kolomanski, line 1: suggestions: I have some more suggestions that could improve the readability of the file, but they are not critical.

#### Function ordering
I think the ordering of the functions in the file is a bit off. The best ordering would be I believe:
1. Exported / callable functions (the functions that call `UseMethod`), since they define the public API and should be on top.
2. Method implementation (so all `*.PKNCAdata` and `*.data.frame` functions).
3. Helper functions (`add_impute_method`, `remove_impute_method`).
Currently the functions are a bit mixed together (like the helper functions being between the method implementations) which makes it a bit hard to find a way around

#### Code repetition
Seems like both functions for adding and removing methods share some similarities - I think the file could use some more internal / helper functions to reduce code duplication and improve the readability.

tests/testthat/test-intervals_support_funs.R
m-kolomanski, line 1: issue: Please rewrite the tests using `describe()` and `it()` to match the convention we already have for unit tests.

issue: Please do not start your tests with handling errors or specific cases. Please put the _ideal case_ test at the front, so that a person working with the code has the simplest example of function working as expected up top, like in the `@examples` part of the documentation.

m-kolomanski (changes requested):

R/intervals_support_funs.R
Gero1999: I agree this looks more readable. I will add the equivalent as well in the other function. Thanks!

tests/testthat/test-intervals_helpers.R
```diff
@@ -0,0 +1,460 @@
...
+)
+
+o_conc <- PKNCA::PKNCAconc(d_conc, conc ~ time | ID / analyte, include_half.life = "include_hl")
+o_dose <- PKNCA::PKNCAdose(d_dose, dose ~ time | ID)
+o_data <- PKNCA::PKNCAdata(o_conc, o_dose, intervals = intervals)
+
+describe("interval_add_impute", {
+  it("throws an error if either data or target_impute is missing", {
+    expect_error(interval_add_impute(o_data), "Both 'data' and 'target_impute' must be provided.")
+  })
```
m-kolomanski, line 33: issue: please start the test suite with the most basic, ideal-case scenario for calling the function (i.e. like the one provided in `@examples`).

tests/testthat/test-intervals_helpers.R
```diff
@@ -0,0 +1,460 @@
...
+
+  it("handles impute column with FALSE values correctly", {
+    o_data_with_na_impute <- o_data
+    o_data_with_na_impute$intervals$impute <- NA_character_
+    result <- interval_add_impute(o_data_with_na_impute, target_impute = "new_impute")
+    expect_equal(result$intervals[, c("analyte", "half.life", "cmax", "impute")],
+                 data.frame(analyte = c("Analyte1", "Analyte2", "Analyte1"),
+                            half.life = c(TRUE, TRUE, TRUE),
+                            cmax = c(TRUE, TRUE, TRUE),
+                            impute = rep("new_impute", 3)))
```
m-kolomanski, line 59: suggestion: The tests work fine as they are, but to improve readability please consider including the expected result in the _setup_ part of the test, like:
```R
    o_data_with_na_impute <- o_data
    o_data_with_na_impute$intervals$impute <- NA_character_
    expected_result <- data.frame(analyte = c("Analyte1", "Analyte2", "Analyte1"),
                                  half.life = c(TRUE, TRUE, TRUE),
                                  cmax = c(TRUE, TRUE, TRUE),
                                  impute = rep("new_impute", 3))

    result <- interval_add_impute(o_data_with_na_impute, target_impute = "new_impute")

    expect_equal(result$intervals[, c("analyte", "half.life", "cmax", "impute")], expected_result)
```

This gives a cleaner distinction  between 1. setup 2. run and 3. assertion parts of the tests. This is not always necessary, like when we are using primitive values as expected results (`expect_equal(result, 5)` or something), but with results as complex ad data frames, this improves the readability.

m-kolomanski (changes requested):
Hey, the code looks good, thank you for all the tests, they are good but I have some suggestions to improve the structure / readability.

tests/testthat/test-intervals_helpers.R
Gero1999: I put one that is even more dummy/simple than the example & I wrote it inside the `it()`, let me know if this seems to make it easier to understand the function's purpose in the test suite

```
it("adds the impute method in the impute column of a dummy intervals dataframe", {
    simple_df <- data.frame(
      cmax = TRUE,
      impute =  c("", "m0", "m0,m1")
    )
    expected_res <- data.frame(
      cmax = TRUE,
      impute = c("mlast", "m0,mlast", "m0,m1,mlast")
    )
    res <- interval_add_impute(simple_df, target_impute = "mlast")
    expect_equal(res, expected_res)
  })
```

tests/testthat/test-intervals_helpers.R
```diff
@@ -0,0 +1,491 @@
...
+                                        "Analyte2", "Analyte1", "Analyte1"),
+                            half.life = c(TRUE, FALSE, TRUE, FALSE, TRUE, FALSE),
+                            cmax = c(FALSE, TRUE, FALSE, TRUE, FALSE, TRUE),
+                            impute = c("start_conc0,start_predose",
+                                       "start_conc0,start_predose,new_impute",
+                                       "start_predose",
+                                       "start_predose,new_impute",
+                                       "start_conc0",
+                                       "start_conc0,new_impute")))
+  })
```
m-kolomanski, line 221: nitpick(non-blocking): I saw that you have updated some tests for better readability, but there are still some cases when a data frame is defined in `expect_*` call, like this one. Again, this is just a question of readability, not something that is blocking, the test works fine.

m-kolomanski (approved):
LGTM, thank you for the work and all the changes! I have left two non-blocking nitpicks for consideration.

js3110 (changes requested):
hi cmax looks good, but I am getting two values for tmax for the bolus subjects, one 0 and one the real value.

Gero1999:
The problem is related with PKNCA calculating implicit parameters. `tmax` is used in `half.life` and then even if not selected is shown by default in the results. Because my strategy here is duplicating interval rows, then two results will pop up. 

The temporary solution will prevent duplicated parameters from happening, and take the last rows as preferent for those cases (which are always the ones affected by my functions).

js3110 (approved):
Thank you, LGTM!

PR #238, opened 2025-02-28
Enhancement: move filters section

### LINKED ISSUE #230: Enhancement: GUI: move filters section
### Description
Move filtering section to before column mapping, so the user can filter with raw data instead.

### Definition of Done
- [ ] Filters in data upload section
- [ ] User can filter for values that are then applied to the rest of the analysis

js3110:
## Issue

Closes #230 

## Description

Move filtering section to before column mapping, so the user can filter with raw data instead.

## Definition of Done

- [x] Filters in data upload section
- [x] User can filter for values that are then applied to the rest of the analysis

## How to test

Add filters at the start and check if the rest of the app works.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer


inst/shiny/modules/tab_nca.R
m-kolomanski, line 1: question: This change seems unrelated, should it be here?

inst/shiny/modules/tab_data.R
```diff
@@ -87,10 +87,39 @@ tab_data_server <- function(id) {
       ADNCA(new_adnca)
     })
 
+    adnca_filter <- reactiveVal(NULL)
```
m-kolomanski, line 90: issue: I think the name of this variable should be something like `adnca_filtered`. The word `filter` here is a bit confusing.

inst/shiny/modules/tab_data.R
```diff
@@ -87,10 +87,39 @@ tab_data_server <- function(id) {
...
+    observeEvent(list(input$submit_filters, processed_data()), {
+      # Extract filters from reactive values
+      applied_filters <- lapply(reactiveValuesToList(filters), \(x) x())
+
+      # Filter and overwrite data
+      filtered_data <- apply_filters(
+        ADNCA(), applied_filters
+      )
+      adnca_filter(filtered_data)
+    }, ignoreInit = FALSE)
```
m-kolomanski, line 117: 1. issue: Seems like there is no need for `adnca_filter` to be a `reactiveVal`, could be a `reactiveEvent` or `reactive(...) |> bindEvent()`
2. question: With this setup, when column mapping is applied, filters will also be auto-applied, without explicit user confirmation. Is that a desired behavior?

m-kolomanski (changes requested):

inst/shiny/modules/tab_nca.R
js3110: i think that i was getting a crash because there were no intervals.. not sure how it was related to changing the filtering but it seemed to happen only after I moved the filters section

inst/shiny/modules/tab_nca.R
```diff
@@ -113,9 +113,20 @@ tab_nca_server <- function(id, data, grouping_vars) {
...
+          cmax = TRUE,
+          tmax = TRUE,
+          auclast = FALSE,
+          aucinf.obs = FALSE
+        )
+
       mydata <- PKNCA::PKNCAdata(
         data.conc = myconc,
         data.dose = mydose,
+        intervals = intervals,
```
Gero1999, line 129: issue: Hello Jana, I am reluctant to this. I tested the branch taking it off and everything seems to keep working well. If it does not for you, let me know under what specific filter/actions the crush occurs

Gero1999 (changes requested):

inst/shiny/modules/tab_nca.R
js3110: @Gero1999 I went back to check, and its if you filter for only 1 subject, the app crashes, with the error: 
`Warning: Error in check.interval.specification: interval specification has no rows`

Also, with this code, we no longer get all the warning messages that say `Warning in PKNCAdata.default(data.conc = data.conc, data.dose = data.dose,  :
  STUDYID=XX01; PCSPEC=SERUM; DRUG=Analyte01; USUBJID=11201; ANALYTE=Analyte01; data_intervals=NULL: No intervals generated likely due to limited concentration data` etc

Let me know what you think! if theres another way to avoid this error that you prefer to use then we can try that

Gero1999 (approved):
Hey @js3110, I see the error now. Instead of a temporary fix, I suggest we find and address the root cause and open a new issue for it.

I will approve the pr, but please mark the relevant code lines with a #ToDo. The error most likely originates from the dataset or the `format_pkncadose_data()` function when creating the dose dataset for one subject. It's important to identify when and why this happens and open an issue for it.

js3110:
> Hey @js3110, I see the error now. Instead of a temporary fix, I suggest we find and address the root cause and open a new issue for it.
> 
> I will approve the pr, but please mark the relevant code lines with a #ToDo. The error most likely originates from the dataset or the `format_pkncadose_data()` function when creating the dose dataset for one subject. It's important to identify when and why this happens and open an issue for it.

Ok thanks! will do, although are you sure this is a temporary fix? the issue is due to there being no intervals set, so setting intervals is the fix, no?

Gero1999:
I am concerned that we could be just hiding the issue. The `intervals` is not a mandatory arg and `PKNCAdata` by default should be generating intervals based on the dose data (even if we don't currently use them in the app). I think is worthy we investigate it more at some point

Issue #210, opened 2025-02-19
Enhancement: remove apply button in dynamic slope plots


Gero1999:
### Description
Remove the apply button that is currently needed to save the changes in the slope selector plots when the user makes modifications. This has to be dynamically saved.

> Blocked until #174 #177 are merged
> Blocked until #209 is merged

### Definition of Done
The requirements for the feature to be complete:
- [ ] no `Apply` button, things are saved instantly
- [ ]  Add a check to make sure text REASONS have been included for ~~`Inclusions`~~ `Exclusions` when running `NCA`

PR #241, opened 2025-02-28
Enhancement: Slope Selector Pre NCA

### LINKED ISSUE #233: Enhancement: Slope selector displayed without running full NCA
### Description
Slope selector display before NCA is run, so user can already see the plots and make their own best fit selections.

### Definition of Done
- [x] plots available before and after the nca
- [x] before nca, no line of best fit, until the user makes manual slopes
- [x] after nca, displayed as normal, including and manual slopes added

js3110:
## Issue

Closes #233 

## Description

Update slope selector plots so that they appear before NCA has been run, and are still able to update based on manual slope selections.

## Definition of Done

- [x] plots available before and after the nca
- [x] after nca, displayed as normal, including and manual slopes added

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

Still some bugs in slope selector, but should be covered in #121 
Maybe need to make a new issue as the cmax and lambda.z.n rules are not being implemented as expected (manual selections should override the default settings, but are not currently)


inst/shiny/modules/manual_slopes_table.R
```diff
@@ -0,0 +1,217 @@
...
+          manual_slopes(edited_slopes)
+        })
+      })
+    })
+
+    #' saves and implements provided ruleset
+    observeEvent(input$save_ruleset, {
+      mydata(filter_slopes(mydata(), manual_slopes(), profiles_per_patient(), slopes_groups()))
+      pk_nca_trigger(pk_nca_trigger() + 1)
+    })
```
m-kolomanski, line 208: discussion: I see two issues here (I know they are a left-over from previous work, but maybe we wish to fix them here):

1. With this the **Apply** button is synonymous with **Run NCA**. Do we want that behavior? I think that from the UX perspective we should get rid of the **Apply** button altogether. The selections / exclusions are gathered and returned by the module automatically. Then the user should just click the **Run NCA** button and observer responsible for running the analysis should apply the manual slopes.
2. This module modifies the `mydata()` object directly. It should not. It returns the manual slopes upstream and the module that runs the data pipeline should update data before running the analysis.

We could solve those two issues within the scope of this PR, or I could take on improving the situation with #264.

Let me know what you think!

js3110: hi @m-kolomanski there is already an issue #210 for the apply button, so i didn't want to cover it here, although since its quite simple I will do it now and then close 210 as duplicated.
i think since #264 will be changing mydata a lot anyway in terms of how its processed, maybe its easier to take it on there?

js3110: actually, sorry i just looked into it a bit more and since both issues are quite linked, i would say its better to do them separately either in #264 or #210, so I'll leave this one as is if thats ok?

m-kolomanski: Yes, absolutely. Thank you for linking #210, I have missed this issue!

m-kolomanski (approved):
TMLG, thank you!

m-kolomanski:
BTW please remember to increment the package version before merging, it is now outdated

Gero1999 (approved):

Issue #225, opened 2025-02-26
Enhancement: Implement customization table for `l_pkconc`


m-kolomanski:
### Description
`l_pkconc` function takes in `formatting_vars_table` as an argument. This should be an option that is customizable by the user.

Implementing #224 is required for this task.

### Definition of Done
- [ ] `formatting_vars_table` is passed to the `l_pkonc` function based on user input.

PR #245, opened 2025-03-03
Enhancement: create table input widget for tlgs

### LINKED ISSUE #224: Enhancement: create table input widget for TLGs
### Description
A table widget for specifying TLG options should be implemented. This will give the user more freedom to specify e.g. formatting options on per-variable / per-column basis.

#223 is required for this to be accomplished.

### Definition of Done
- [ ] implemented new option widget for TLGs to allow per-variable customization of options
- [ ] the module returns a table with customizations specified by the user.

m-kolomanski:
## Issue
Closes #224, #225 

## Description
- Adds new type of option type for the TLG system: `table`
- Updates the definition of `l_pkconc` tlg to use the new table
- Updates the documentation and converts the contributing guidelines regarding TLGs into a vignette

## Definition of Done
- [x] implemented new option widget for TLGs to allow per-variable customization of options
- [x] the module returns a table with customizations specified by the user.
- [x] `formatting_vars_table` is passed to the `l_pkonc` function based on user input.

## How to test
Run the app, map the data, submit a TLG, go to listings. Try opening a formatting table. Try changing some values, like providing a custom labels or alignment of the text. Check if confirming updates the listing.

## Contributor checklist
- [x] New logic is documented


Gero1999 (changes requested):
Wow wow great work! Only a few ideas:

- [ ] suggestion: prevent the user from the error of putting round/digit actions on non-numeric/double columns (i.e, `AVISIT`, `TRT01A`)

- [x] suggestion/issue: consider removing the Add/Remove rows. also I can see that currently the `select + remove` action does not work as expected (removes last one)

- [ ] suggestion: if  possible, consider permanently linking the variables in the formatting tables with the ones selected in the other input variables

- [ ] nitpick: maybe consider making the table/modal message bigger so it is easier to see the entire `Label` column

- [x] nitpick: column width of `Digits` could be reduced or alternatively use a select option 0-4. That way the user won't introduce doubles or characters

m-kolomanski:
@Gero1999 Thank you for the feedback!

I have:
- enlarged the modal and the table;
- changed `Digits` to a select input;
- fixed the table not updating after removing a row -> Please let me know if we wish to keep it? What would be the reason to remove this feature? Seems like it might be needed if the user wants to add new formatting conditions.

In regards to the other two features, I have left them out on purpose. That is because I have not found a neat solution. Recently we have spoken about the TLG system getting very complex. I probably could implement those two features, but this would require a) much more time spend tinkering and b) add to the complexity of the whole thing. In addition, those features seemed like QOL to me rather than critical functionalities. I trusted that most of the users will be experts that know what sort of values fit AND if anything goes wrong, the application will not crash but just display an error stating that rendering failed with provided option.

I _could_ spend more time and probably implement these features if you wish, I would be happy to do so, but in my mind I preferred to miss non-critical features and move onto more pressing tasks. We can always revisit and refine this in the future.

Let me know what do you think!

Gero1999 (approved):
thanks for the changes, the layout is definitely amazing! We can leave the left things, not big deals. In most cases the users will just want to modify the format on these standard variables. 

But I would still not disconsider completely to come out at some point with an idea that doesn't force the user to include vars and then manually rows. Perhaps joining all together including in the table a column for the grouping action (table-grouping / list-grouping / display-var).

inst/shiny/modules/tlg/tlg_option_table.R
```diff
@@ -0,0 +1,158 @@
+#' Function generating an input widget for TLG option.
+#' @param id      id of the input widget
+#' @param opt_def definition of the option, as specified in the `yaml` file
+#' @param data    data object used for parsing labels, strings, infering placeholder values or
+#'                choices etc.
+tlg_option_table_ui <- function(id, opt_def, data) {
```
rabiibouhestine, line 6: the argument `data` is not used in `tlg_option_table_ui `. Is there a reason to have it?

inst/shiny/modules/tlg/tlg_option_table.R
```diff
@@ -0,0 +1,158 @@
...
+  actionButton(ns("open_table"), label = label)
+}
+
+#' Function generating an input widget server for TLG option.
+#' @param id      id of the input widget
+#' @param opt_def definition of the option, as specified in the `yaml` file
+#' @param data    data object used for parsing labels, strings, infering placeholder values or
+#'                choices etc.
+#' @returns a reactive with the input value
+tlg_option_table_server <- function(id, opt_def, data) {
```
rabiibouhestine, line 20: Consider adding more details to the docstrings, ex: expected structure / type of `opt_def` and `data`.

inst/shiny/modules/tlg/tlg_option_table.R
```diff
@@ -0,0 +1,158 @@
...
+                  names(data())
+                } else if (length(def$choices) == 1 && grepl("^\\$", def$choices)) {
+                  unique(data()[, sub("^\\$", "", def$choices)])
+                } else {
+                  def$choices
+                }
+              },
+              class = "dropdown-extra"
+            ),
+            stop("Unsupported extra")
```
rabiibouhestine, line 67: Maybe a more informative error message here? something like:

`stop("Unsupported type: ", def$type, " for column: ", colname)`

inst/shiny/modules/tlg/tlg_option_table.R
```diff
@@ -0,0 +1,158 @@
...
+    #' Confirm changes, apply the changes to output table.
+    observeEvent(input$confirm_changes, {
+      removeModal()
+      output_table(edits_table())
+    })
+
+    #' Cancel changes, close the modal, reset values.
+    observeEvent(input$cancel, {
+      removeModal()
+      shinyjs::runjs("memory = {};") # needed to properly reset reactable.extras widgets
```
rabiibouhestine, line 129: This line `shinyjs::runjs("memory = {};")` is repeated multiple times. Consider encapsulating this in a helper function to avoid repetition:
```
reset_reactable_memory <- function() {
  shinyjs::runjs("memory = {};")
}
```

rabiibouhestine (commented):
Great job, I made a few comments, will approve once addressed.

inst/shiny/modules/tlg/tlg_option_table.R
m-kolomanski: Yes, the `data` object is here for two reasons:
1. Even though it is not used now, it will be in the planned features. In order to have a clear structure for the module, I decided to go ahead of myself and add the argument. In addition, it is to
2. keep the module structure consistent across different input modules. The modules are not really called by the developer, instead they are invoked in a [lapply call iterating over options defined in the yaml file](https://github.com/pharmaverse/aNCA/blob/ec313e1c2702d1e3d95f7f9958be21cc63dea275/inst/shiny/modules/tlg/tlg_module.R#L185), so I wanted this part of the code to be able to pass the same arguments to all types of input modules. Again, I could have used `...` here, but I wanted to be verbose for the future.

rabiibouhestine (approved):
LGTM

PR #251, opened 2025-03-04
Bug: tlg plots fail with an error `all unnamed arguments must be length 1`

### LINKED ISSUE #226: Bug: TLG plots fail with an error: `All unnamed arguments must be length 1`

m-kolomanski:
## Issue

Closes #226 

## Description
The bug stemmed from changed dummy data (`Dummy_complex_data.csv`). Data includes several units. The default argument for `g_pkconc_ind` plot for generating ylab hence created a vector instead of a string, which broke `tern::g_ipp` call. This was not caught in unit tests because the tests still ran on the old data.

This PR:
- fixes the issue.
- updates dummy data for testing purposes.

In addition, fixes an issue where labels for widgets with no specified label are generated wrong.

## Definition of Done

- [x] TLG plots for `g_pkonc_ind` should render properly.
- [x] Labels for widgets where label is not provided should render properly.

## How to test
On `main`, run the app, map the data, submit TLG order, go to Graphs:
- Graph will not render, show an error instead
 - Labels for xmax/xmin/ymax/ymin widgets will include namespaced module id.

On this branch:
- Graphs should render properly.
- Labels should not include namespacing.

## Contributor checklist

## Notes to reviewer
Due to long loading times (with current dummy data `g_pkconc_ind` generates 92 plots!) and the nature of reactivity with `shinycssloaders`, the loading spinner freaks out a little. Any suggestions how to improve the situation are welcome.


Gero1999 (approved):
thank you for solving this, seems to be completely functional again. The only thing I noticed: 

- [ ] issue: Reset button not working 

However I guess this might be for another reason, and perhaps we can open an independent issue, at your choice. 

Regarding the loading issue, I want to point out that this is by far the TLG with the greatest demand because we need one plot per patient. The others will be in general more aggregated. And I am not technical but I guess that any sacrifice that can save speed would help us. I can talk with our product advisors/stakeholders if needed to propose them potential options: 

- Set by default 1 plot instead of all at the start of the app. Make the function just generate the plots we need for the display selected
- Eliminating ggplotly and going to plain ggplots

m-kolomanski:
Hey @Gero1999,

thank you for pointing out the issue with the reset button - I have checked and I have broken it when refactoring the TLG system. I have added this as a separate task #254.

I terms of speed, I do not think it is too bad? Some things just take time, R being not the fastest language - I have seen slower applications. IMHO the main thing is to inform the user that _something_ is happening, because when staring at a blank page for several seconds you start to wonder if something broke or not?

Of course, we could look into speeding up the process, but I would not consider it critical at the moment - let me know your thoughts.

@Gotfrid If you have any ideas how to fix the `shinycssloader` so it shows up properly (without too much refactoring 🙃) it would be great!

rabiibouhestine (approved):
LGTM

PR #255, opened 2025-03-06
Fix: do not display list names in TLG listings

### LINKED ISSUE #228: Bug: Name of the listing appears in the printed results
### Description
In the TLG - Listings tab, the name of the listing / object appears in the printed results (e.g. `$SERUM.intravascular.Analyte01`).

### To reproduce
Map the data, go to TLG tab, submit an order that includes some listings and go to the Listings tab. The result will have name of the entry visible in the results.

### Expected behaviour
Object name should not appear in the results.

m-kolomanski:
## Issue
Closes #228

## Description
Since we rely on `renderPrint` to display listings with original formatting, the output displays names of the elements, which looks ugly. This PR introduces changes that `unname()` the list of displayed TLGs in order to show index numbers instead of names.

## Definition of Done
- [x] List element name does not appear in the results of Listings TLGs.

## How to test
Run the app, map the data, go to TLGs tab, submit default order. Check if the Listings display includes index numbers instead of names.

## Contributor checklist

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.0.0.9016
+Version: 0.0.0.9017
 Authors@R: c(
     person("Ercan", "Suekuer", , "ercan.suekuer@roche.com", role = c("aut", "cre"),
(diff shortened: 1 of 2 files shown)


rabiibouhestine (approved):
LGTM

PR #256, opened 2025-03-06
feat: statistic filtering input & add DOSNO as groupby option - #246

### LINKED ISSUE #246: Enhancement: Summary statistics table filtering by statistic
### Description
Be capable of filtering the rows in the section `Results` >`Descriptive Statistics` based on the `Statistic` column

### Definition of Done
- [ ] Input to filter displayed rows in `Results` >`Descriptive Statistics` based on the `Statistic` column

Gero1999:
## Issue

Closes #246

## Description
Users have asked for the possibility of filtering by Statistic in the summary table 
I also noticed `DOSNO` was missing as a possible var to group_by

## Definition of Done
- [x] Input to filter displayed rows in `Results` >`Descriptive Statistics` based on the `Statistic` column
- [x] DOSNO can be used to group_by the data

## How to test
`Map data > NCA setup > Run NCA > Results: Descriptive Statistics > Check all filters work as expected`

## Contributor checklist
- [x] New logic is documented


js3110 (changes requested):
Looks good, but i would say DOSNO should automatically be selected as one of the groups.

js3110 (approved):
LGTM! thanks

m-kolomanski (approved):
LGTM!

question, suggestion: I have noticed that in the table that has all stats selected and 10 entries per page, the last row is the geomean for second subject. Then the second page shows all stats for the second subject, but the geomean. Would it be beneficial to show N rows per page, where N is the number of selected stats? So that all stats for a single subject / group are shown on single page.

Then if very little discrete stats are selected (like < 5), then do the product, like 2 subjects x 5 stats = 10 rows per page.

I do not know if this would improve the experience. Does not have to be done within this PR, I just had a thought and I am throwing it out.

Gero1999:
> LGTM!
> 
> question, suggestion: I have noticed that in the table that has all stats selected and 10 entries per page, the last row is the geomean for second subject. Then the second page shows all stats for the second subject, but the geomean. Would it be beneficial to show N rows per page, where N is the number of selected stats? So that all stats for a single subject / group are shown on single page.
> 
> Then if very little discrete stats are selected (like < 5), then do the product, like 2 subjects x 5 stats = 10 rows per page.
> 
> I do not know if this would improve the experience. Does not have to be done within this PR, I just had a thought and I am throwing it out.

Actually I like your idea of separating in groups the pages @m-kolomanski! The only thing is that I feel it would make more sense in the opposite way, separating them by Statistic so the researcher can compare easily the differences between the mean/median/SD of the different groups.  What do you think @js3110 @Shaakon35 ?

PR #258, opened 2025-03-06
bug: fix parameter selection in main NCA results for different analyte units

### LINKED ISSUE #257: Bug: Parameter filter input in main NCA results not working due to units
### Description
Parameter filter input in main NCA results not working. Due to the fact that `pivot_wider_pknca()` includes directly the units now in the results. This is needed to be done in case there are multiple analyte with different units. So now we need to figure out a different way to make it work easily.

### To reproduce
`Map data > Run NCA > Modify default Select Parameters`

### Expected behaviour
When a parameter is provided in the input `Select Parameters` all forms in units of that parameter should be displayed

Gero1999:
## Issue

Closes #257 

## Description
Parameter filter input in main NCA results was not working due to the fact that pivot_wider_pknca() includes directly the units now in the results. This is needed to be done in case there are multiple analyte with different units. So now we need to figure out a different way to make it work easily.

The proposed solution basically uses regex to map a parameter selection to all column derivations of that parameter (`parameter[unit]`)

## Definition of Done
- [x] Parameter selection can be done and will include all derived analyte units

## How to test
`Map data` > `NCA settings: Choose multiple analyte` > `Units Table: Modify the unit of a parameter for 1 analyte` > `Run NCA` > `Results NCA` > `Play with the Parameter picker input`

## Contributor checklist
- [x] New logic is documented


m-kolomanski (commented):
suggestion, up to discussion:

The solution seems fine, but I am always cautions of creating complex.

What do you thing about utilizing column attributes since we are already using the `label` attribute in the package?

In `pivot_wider_pknca_result` we could set a unit agnostic attribute describing PPTESTCD for each column and then in `nca_results.R` just do something like
```R
select(final_results(), where(~ isTRUE(attr(., "PPTESTCD") %in% input$params)))
```

What do you think about this kind of approach?

Gero1999:
> In `pivot_wider_pknca_result` we could set a unit agnostic attribute describing PPTESTCD for each column and then in `nca_results.R` just do something like
> 
> ```r
> select(final_results(), where(~ isTRUE(attr(., "PPTESTCD") %in% input$params)))
> ```
> 
> What do you think about this kind of approach?

I like your idea  of creating labels in `pivot_wider_pknca_result`, but actually I am thinking maybe it could be better to exploit it to make `generate_col_defs()` create proper hovering messages for the display (i.e, r.sq.adj[unitless] >> "R Squared Adjusted")

I guess that your main concern with this code is this specific regex action, which can seem complex:

``` 
      params_selected_cols <- grep(paste0("^", "(", paste0(input$params,
                                                           collapse = "|"), ")", "\\[.*"),
                                   names(final_results()), value = TRUE)
```

If you want I can counter propose doing something else that is not a regex to guess the parameter - columname relation. Like a named vector created with `res_nca()` 

```
param_columname_vr <- res_nca()$result %>% 
 mutate(columname = paste(PPTESTCD, PPSTRESU)   %>% 
 pull(PPTESTCD, columname)
 
 select(finaL_results(), param_columname_vr[input$params])
```

Would this approach look safer to you @m-kolomanski?

m-kolomanski:
I think it would be great to utilize this for things like hover labels. I would think it should be a separate issue though? 

Adding the attributes does not break anything inherently, and might be useful in such cases.

Also, I was reading your comment when you edited it, and the proposed solution of creating a kind of _mapping_ that then can be used to select the columns would also be a great solution, much better than regexpr, if we did not want to go into modifying `pivot_wider_pknca_result`.

Gero1999:
Yes, I think that  my mapping idea will also work. But I actually think your solution already solves 2 things in one. The mapping without regex and also the user get a better display of the inputs. 

It is true it might seem like a feature, but I cannot imagine how to separate it from the bug now. Is any actual way we can split them?

Gero1999:
We can maybe create the issue-feature and link this time this branch to two issues?

m-kolomanski:
I think that just adding the unit-agnostic attribute and using it for selecting the right columns despite the units is 100% related to solving the issue, no need to change anything in the PR.

But implementing those attributes for different purposes, like the labels, I would create a separate feature task and a separate PR for that.

Gero1999:
Ok then I will open the issue and block it for now, so anyone can take on it once this branch is merged

m-kolomanski (approved):
Thank you for the changes and implementing the labels, looks great 💪

PR #259, opened 2025-03-06
Bug/flag rules crash due to units #250

### LINKED ISSUE #250: Bug: Flag rules crush the app (cause: finalresults units)
### Description
Flag rules in NCA settings are not working. Likely due to a change in the `finalresults()` variable. Following the rest of the code logic we should make the flag rules independent of this object and instead use `res_nca()` to prevent problems. 

### To reproduce
Map Data > NCA > Setup/Settings > i.e, RSQADJ ✅ : > Run NCA

### Expected behaviour
- [ ] No crush when the `To reproduce` action is executed
- [ ] Server logic should be independent from objects that are coming from other tabs

Gero1999:
## Issue

Closes #250 

## Description
Fix: Flag rules in NCA settings are not working. Likely due to a change in the `finalresults()` variable

The proposed solution simply considers that the final_results contain column names with the units added, using the res_nca().

Change description.

## Definition of Done
**Testing feature functionality**
- [x] No crush should happen 
- [x] Check that results with lower r.sq.adj (or bigger AUCPEO/AUCPEXT) are flagged in red

## How to test
`Map Data > NCA > Setup/Settings > i.e, RSQADJ ✅ : > Run NCA`

## Contributor checklist
- [x] New logic is documented


inst/shiny/modules/nca_results.R
```diff
@@ -36,11 +36,18 @@ nca_results_server <- function(id, res_nca, rules, grouping_vars) {
           gsub("^rule_", "", x = _) |>
           gsub("_", ".", x = _, fixed = TRUE)
 
+        pptestcd_in_fr <- res_nca()$result %>%
```
m-kolomanski, line 39: nitpick(non-blocking): `*_in_fr` is not quite obvious, I am guessing this stands for _final results_? I would maybe go for something like `final_pptestcd` instead, but that is subjective

m-kolomanski (approved):
LGTM

m-kolomanski:
FYI, when testing this feature I have encountered a bug, which I have decided to add to the board as a separate task, since it seemed it was not related to the changes at hand.

#260

Gero1999:
thanks for the revision and finding that bug Mateusz!

m-kolomanski (approved):

PR #268, opened 2025-03-11
Fix: reset widgets button for TLGs not working

### LINKED ISSUE #254: Bug:  Reset button for TLG options not working
### Description
Reset button for TLG options is not working - nothing happens when clicked.

### Expected behaviour
Reset button should reset the all options for particular TLG to their default values.

m-kolomanski:
## Issue

Closes #254 

## Description

Each of the option input modules for TLGs takes in a reactive `reset_trigger`. When the value for this trigger is changed, the module resets its value to default one.

As a default, the `input$reset_widgets` button of the `tlg_module` is passed as a trigger to all the options modules.

## Definition of Done

- [x] Reset widgets button restores the default values for all input widgets

## How to test

Run the app, map the data, go to TLG tab and submit an order. Try changing some values for text, numeric, select and table inputs. Click on the **Reset widgets** button. Check if default values are reinstated.

## Contributor checklist


Gero1999 (approved):
LGTM, thanks for the fix! 

I noticed an unrelated thing that may be an easy-to-fix bug or not even a bug, but in `Graphs/pkcg01-linear` the input `Group variables` does not allow other choices than the ones set by default

m-kolomanski:
@Gero1999, thank you for spotting the issue, I would kindly ask you please to create a separate issue for this task and describe whether we want all colnames in the picklist or some specific columns only?

rabiibouhestine (approved):
LGTM

Issue #249, opened 2025-03-04
Bug: Partial AUCs problem when start is not 0


Gero1999:
### Description
The partial AUCs are crashing the app when added

### To reproduce
`Map data` > `NCA Settings` > `AUC partial ✅` > `Add AUC` > Run NCA

### Expected behaviour
- [ ] Run any partial AUC without issues
- [ ] Suggestion: Reconsider the layout. If needed open another issue

js3110:
blocked until #272 merged

PR #272, opened 2025-03-14
Refactor: `tab_nca.R`

### LINKED ISSUE #264: Refactor: `tab_nca.R` to rename variables, use `reactive` when possible and modularize further
### Description
This would be a step towards simplifying the data flow in the application.

- use `reactive` instead of `reactiveVal` whenever possible
- avoid modules modifying any top object directly, instead those should return a modified copy that is the passed further down the pipeline
- variable renaming:
  - `mydata` -> `pknca_data_object`
  - `output$manual_slopes2` -> `output$slope_rules`
- modularization:
  - move NCA settings saving to a separate module
- seems like there is a bunch of code that handles `manual_slopes` object which is not used in the module (L198-221)
- add more documentation regarding what the workflow is

I believe the next step after implementing the proposed changes would be addressing issue #121 and implementing the workflow [described by Gerardo](https://github.com/pharmaverse/aNCA/issues/121#issuecomment-2627132115). When this module data flow is simplified, we could semi-detach the `slope_selector`, so that it also can run NCA analysis and allow the user to select the slopes before running the actual pipeline.

@Gero1999 @js3110 Please let me know your thoughts.

m-kolomanski:
## Issue

Closes #264 

## Description

Refactors the `tab_nca.R` module by: 
  - improving the data pipeline
  - simplifying the reactive logic
  - renaming variables
  - moving logic code into testable functions
  - implementing some more logs
  - moving some code to separate modules

**Minimal** changes have been done to other modules in order to make them work with the changes to `tab_nca.R`. The sub-modules should be refactored and fitted better in their own right.

This PR **breaks** some of the features (user defined intervals, settings download). Due to the complexity and the size of the PR, I would suggest restoring those features in their own tasks.

In addition, the longer the refactoring takes, the more conflicts will be generated, which will extend the life of this PR and so on. Therefore, I would suggest focusing on the issue at hand (`tab_nca.R` module) and working on other issues iteratively.

## Notes to reviewer

Like with the previous refactor, I would suggest disregarding the diff and going top-to-bottom on the relevant files (`tab_nca.R` and `PKNCA.R`). You can then investigate the sub-modules to see how the logic is passed down. Please do bear in mind that most of the sub-modules were coded for different reactivity schema and I have added ad-hoc solutions in order to fit them to the new data flow.

Any feedback on general structure and logic improvements welcome.

If you find any features that are not working properly, please also let me know how critical fixing those is. If it is not of upmost importance, I would suggest creating separate bug issues.

## Contributor checklist
- [x] New logic is documented


m-kolomanski:
Hey @js3110 @Gero1999,

as a part of this PR I want to move some purely logic code outside of the shiny application to improve the distinction between analytics pipeline and the app, to make the code better structured and more testable.

I have created two new functions:

- [PKNCA_create_data_object](https://github.com/pharmaverse/aNCA/blob/c5a4e95dda9bbf7c6b12525754efb84fe3200ce9/R/PKNCA_create_data_object.R)
- [PKNCA_calculate_nca](https://github.com/pharmaverse/aNCA/blob/c5a4e95dda9bbf7c6b12525754efb84fe3200ce9/R/PKNCA_calculate_nca.R)

Would you mind jumping in and adding some documentation to this code, examples and ideally some basic unit tests? We are not only using `PKNCA` functions for calculations, but also doing some other transformations (setting default settings, updating units, some joins and additional calculations etc). It would be beneficial to have the details in the docs, but I myself am not 100% sure what is happening and why exactly.

js3110:
@m-kolomanski I've added some docs for the data creation and will create some tests now! please let me know if you think the docs are clear enough, or I can always add more information ifneeded 😄

js3110:
@m-kolomanski I have added tests to the data object creation function. It would be good if the tests for "handles missing columns" can be updated in the function to give the user a pop up with the error message in the app, as we want them to be able to see the error without having to use the console.

m-kolomanski:
@Gero1999 There is part of the code that handles intervals from user input that messes with the reactivity a lot:
```R
    intervals_userinput <- reactive({
      # Collect all inputs for the AUC intervals
      input_names_aucmin <- grep("^timeInputMin_", names(input), value = TRUE)
      input_names_aucmax <- grep("^timeInputMax_", names(input), value = TRUE)

      starts <- unlist(lapply(input_names_aucmin, \(name) input[[name]]))
      ends <- unlist(lapply(input_names_aucmax, \(name) input[[name]]))

      # Make a list of dataframes with each of the intervals requested
      intervals_list <- lapply(seq_along(starts), function(i) {
        mydata()$intervals %>%
          mutate(
            start = start + as.numeric(starts[i]),
            end = start + as.numeric(ends[i])
          ) %>%
          # only TRUE for columns specified in params
          mutate(across(where(is.logical), ~FALSE)) %>%
          # Intervals will always only compute AUC values
          mutate(across(c("aucint.last"), ~TRUE)) %>%
          # Identify the intervals as the manual ones created by the user
          mutate(type_interval = "manual")
      })

      # Make sure NAs were not left by the user
      intervals_list <- intervals_list[!is.na(starts) & !is.na(ends)]
      intervals_list
    })
```
Listening on all `names(input)` in our case overwhelms the system a bit and and really weird things start to happen.

This is part of the `nca_setup.R` module, while this PR focuses on `tab_nca.R`. The above code needs major refactor to make it work as it should, but I really do not want to extend current PR above what is necessary (it is already quite big).

Would it be acceptable if I have **disabled** this feature for now and refactored it in a later PR? I can handle it, but I would prefer to do it separately in order not to overwhelm the reviewer (and myself too) with the amount of changes.

js3110:
> @Gero1999 There is part of the code that handles intervals from user input that messes with the reactivity a lot:
> 
> ```r
>     intervals_userinput <- reactive({
>       # Collect all inputs for the AUC intervals
>       input_names_aucmin <- grep("^timeInputMin_", names(input), value = TRUE)
>       input_names_aucmax <- grep("^timeInputMax_", names(input), value = TRUE)
> 
>       starts <- unlist(lapply(input_names_aucmin, \(name) input[[name]]))
>       ends <- unlist(lapply(input_names_aucmax, \(name) input[[name]]))
> 
>       # Make a list of dataframes with each of the intervals requested
>       intervals_list <- lapply(seq_along(starts), function(i) {
>         mydata()$intervals %>%
>           mutate(
>             start = start + as.numeric(starts[i]),
>             end = start + as.numeric(ends[i])
>           ) %>%
>           # only TRUE for columns specified in params
>           mutate(across(where(is.logical), ~FALSE)) %>%
>           # Intervals will always only compute AUC values
>           mutate(across(c("aucint.last"), ~TRUE)) %>%
>           # Identify the intervals as the manual ones created by the user
>           mutate(type_interval = "manual")
>       })
> 
>       # Make sure NAs were not left by the user
>       intervals_list <- intervals_list[!is.na(starts) & !is.na(ends)]
>       intervals_list
>     })
> ```
> 

Hi @m-kolomanski , I just started working on #249 , and this code is involved in the bug, so I would say you can disable it here, since I am going to be making changes to it in the other branch anyway!
ps. if you have any advice on how I can make this cleaner, please let me know! because currently I am struggling to find a fix to the issue

inst/shiny/modules/tab_data.R
```diff
@@ -51,8 +51,7 @@ tab_data_server <- function(id) {
     # Call the column mapping module
     column_mapping <- data_mapping_server(
       id = "column_mapping",
-      adnca_data = adnca_filtered,
-      on_submit = \() updateTabsetPanel(session, "data_navset", selected = "Review Data")
```
js3110, line 55: Question: why has the on_submit been removed? It is a feature that we would like to keep as after column mapping we want to send the user straight to the review section.

js3110 (commented):
Thank you so much for the work Mateusz, it looks like you have done a lot!
I have found a couple of bugs:

1. Manual slope selections do not work after clicking apply. I would suggest trying to fix it within this PR if its just some simple refactoring, but if its more complicated then create a new issue. But then that would be extremely high priority as this feature is very important.
2. Great that the units table is now working after the results, however it does not update based on the previous units table. If i change a unit before running the nca, then in the results its correct but if I open the units table it still has the original unit, so then if I save that table without changing it again, the units go back to original.

The units issue is less of a priority, so could just be a new bug issue.

inst/shiny/modules/tab_data.R
m-kolomanski: It has not been removed, but moved to `tab_nca.R` file, the plan being to switch the tabs only when the `PKNCA` object is created correctly, otherwise show an error.

I see that this does not work correctly now, and in hindsight it was not the most intuitive solution, I will fix that.

I will fix other issues you have noticed within the scope of this PR, thanks for the feedback!

m-kolomanski:
@js3110 I believe I have fixed the noted issues, please take another look!

js3110 (approved):
LGTM! thank you for the effort!

on testing, i ended up with a bug, see the video attached

https://github.com/user-attachments/assets/64025d30-7609-4c48-8434-31e8f3dd8bd2

not actually sure how it happened, and I havent been able to recreate it. But just in case its relevant i thought I would let you know.

js3110 (changes requested):
Sorry, unapproving because I have noticed a bug while working on another issue in which I set this branch as the upstream.

- When you deselect the should_impute_c0 checkbox, the settings disappear from the display and I am unable to run NCA

rabiibouhestine (changes requested):
Please take a look at my comment, great work otherwise 👍🏼

inst/shiny/modules/tab_nca/nca_setup.R
```diff
@@ -498,94 +502,105 @@ nca_setup_server <- function(id, data, mydata, res_nca) { # nolint : TODO: compl
...
       input$method, input$nca_params, input$should_impute_c0,
       input$select_analyte, input$select_dosno, input$select_pcspec,
-      intervals_userinput()
+      intervals_userinput(), mydata(), session$userData$units_table()
     ), {
+      req(mydata())
+      req(
+        input$method, input$nca_params, input$should_impute_c0, input$select_analyte,
+        input$select_dosno, input$select_pcspec, intervals_userinput()
+      )
```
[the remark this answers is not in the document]
m-kolomanski: Good spot, thanks!

m-kolomanski:
@js3110 @rabiibouhestine Thank you for the feedback, I have fixed the issues, please take another look!

js3110 (changes requested):
Hi! I have noticed some more bugs sorry!!

1. If I select > 1 analyte, and  change the units of one parameter (for both analytes), the results table adds an extra column for every parameter with the units [NA]
2. Adding Manual slopes leads to all of the slopes for every subject being labelled as 'manual'. The calculations I think are correct, but only the subjects modified should be labelled manual.

js3110:
2. Adding Manual slopes leads to all of the slopes for every subject being labelled as 'manual'. The calculations I think are correct, but only the subjects modified should be labelled manual.

Actually regarding this, I have just noticed it is also this way in Main so I will create a separate bug issue!

m-kolomanski:
@js3110 PTAL

js3110 (approved):
LGTM! thanks :)

rabiibouhestine (approved):
LGTM

PR #273, opened 2025-03-14
Enhancement: bioavailability calculations

### LINKED ISSUE #203: Enhancement: bioavailability calculations
### Description

Example table:

Image

### Definition of Done

js3110:
## Issue

Closes #203 

### Description
Currently PKNCA is unable to calculate parameter F (bioavailability). Until that feature has been implemented, we want a temporary function to calculate bioavailability. This should include the following options:

F should be written as a percentage %

The app should be able to automatically determine if there are subjects with both IV and EX routes for each group (taken from `grouping_vars`), or not, and decide which of 1 or 2 it will calculate. The mean should be calculated by the grouping vars. The results should be along the main results. The type of AUC (mean or individual) should also be included in the results.

### Definition of Done
- [x] Bioavailability calculation for individuals, and grouped data
- [x] Automatic calculations with only user input for parameters
- [x] Results table with ratios displayed
- [x] Ratios included in descriptive statistics

## How to test

Use dummy data DummyRO_MRoutes, select bioavailability. Check main results and descriptive statistics for f_ parameters.

## Contributor checklist
- [x] New logic is documented


js3110:
Hi @m-kolomanski and @Gero1999 !

This PR is still a draft as I still need to write some tests for the new functions, but can I ask you both to take a look at it first, so that I can be sure before I write the tests that no major changes need to be made! Thanks!

m-kolomanski:
Hey!

In terms of tests, from technical perspective tests _should_ not care about the implementation. At a quick glance, I have some suggestions, but those would not change the results of the function.

We should write test for what we _want the function to do_, not really _what it does_. So we want x to go in, y to go out and z be handled without crashing, so the question right now is more _what we want the function to do_, which tbh I am no expert at and I do not know exactly ;p

I do not see any substantial flaws in the testable code logic, sorry but I do not have anything valuable to add at this point.

js3110:
thanks @m-kolomanski , you're right, I think theres not much you can add to this for now! @Gero1999 it would be more useful to hear some feedback from you, are you happy with the way that I have implemented this? In terms of creating an updated res_nca with the new parameters?

js3110:
on second thoughts I'm not actually sure how to write tests for these functions as they require PKNCAresults objects with specific features, so you can just go ahead and review it!
I have been thinking that we need a solution for this for the many functions that require results, like maybe having a file that creates a testing dataframe and  PKNCAresults that covers all the types of data we need? or can we just use the dummy data as input for the tests? as trying to create a test dataframe in each testing file is quite difficult, especially when using the PKNCA functions. what do you think @m-kolomanski ?

Gero1999:
> thanks @m-kolomanski , you're right, I think theres not much you can add to this for now! @Gero1999 it would be more useful to hear some feedback from you, are you happy with the way that I have implemented this? In terms of creating an updated res_nca with the new parameters?

hey sorry somehow I missed this thread I will give a look but in principle the idea sounds good

inst/shiny/modules/nca_setup.R
```diff
@@ -508,6 +518,19 @@ nca_setup_server <- function(id, data, mydata, res_nca) { # nolint : TODO: compl
       req(mydata())
       mydata <- mydata()
 
+      # Add picker input if bioavailability calculations are possible
+      if (mydata$dose$data$std_route %>% unique() %>% length() == 2) {
```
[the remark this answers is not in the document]
Gero1999: issue: what would happen if user does not select `aucinf.obs`, `aucinf.pred` and/or `auclast` among the parameters?

R/bioavailability.R
```diff
@@ -0,0 +1,162 @@
...
+#' @export
+calculate_bioavailability <- function(res_nca, selected_aucs) {
+
+  #check if selected_aucs are available
+  if (is.null(selected_aucs) || length(selected_aucs) == 0) {
+    return(NULL)
+  }
+
+  # Extract and clean AUC selection
+  auc_vars <- gsub("^f_", "", selected_aucs)
```
Gero1999, line 37: suggestion/nitpick: put it an input of this function and use directly the input$bioavailability

inst/shiny/modules/nca_results.R
```diff
@@ -94,16 +98,14 @@ nca_results_server <- function(id, res_nca, rules, grouping_vars) {
...
         names()
 
-      group_cols <- setdiff(names(res_nca()$data$intervals),
-                            c(names(PKNCA::get.interval.cols()))) |>
-        # Here cols of interest are also added
-        c("Exclude", "impute", "flagged")
-
       final_results <- final_results() %>%
-        select(any_of(c(group_cols, sort(params_sel_cols))))
+        select(-(c(any_of(params_rem_cols), conc_groups)))
```
Gero1999, line 108: issue: some columns are showing off last - TRT01A, TRT01P, AGE, RACE, SEX, DOSEA, ROUTE

Gero1999:
Also I would add these 2 extra quests that I am not sure by reviewing the code. Maybe they could be good tests. What would happen if:

- Data presents cohorts where f can be calculated, and others in which not. What would be the exclude value as well?
- Data presents cohorts with some DOSNO with both routes, but some with only 1

inst/shiny/modules/nca_setup.R
```diff
@@ -599,17 +622,20 @@ nca_setup_server <- function(id, data, mydata, res_nca) { # nolint : TODO: compl
...
-        aucpext.pred_threshold = input$aucpext.pred_threshold,
+          rule_aucpext_pred = input$rule_aucpext_pred,
+          aucpext.pred_threshold = input$aucpext.pred_threshold,
 
-        rule_span_ratio = input$rule_span_ratio,
-        span.ratio_threshold = input$span.ratio_threshold
+          rule_span_ratio = input$rule_span_ratio,
+          span.ratio_threshold = input$span.ratio_threshold
+        ),
+        bioavailability = input$bioavailability
```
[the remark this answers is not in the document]
js3110: no its for results. It will have a conflict with #272 but the changes are similar so should be easy to fix!

R/bioavailability.R
js3110: ? i don't understand can you clarify please

inst/shiny/modules/nca_results.R
js3110: yes, this is fine as they are the additional columns so can be at the end. sorting them is too complicated and also unnecessary

R/bioavailability.R
Gero1999: I interpreted wrong this code so np

js3110:
> * Data presents cohorts where f can be calculated, and others in which not. What would be the exclude value as well?

Exclude value will match with whatever the aucinf exclude is, and if its an IV subject, there will be no F anyway, so the cell will just be empty.

> * Data presents cohorts with some DOSNO with both routes, but some with only 1

DOSNO is not involved in the calculation, so would have no change to results

PR #205, opened 2025-02-13
Bug/settings based on pknca

Gero1999:
## Issue
- Settings upload does not update dose numbers in Settings Tab of NCA
- Settings seem to not work when no inclusions/exclusions are in the settings file considered
- Settings need to be updated to include multiple analytes

Closes #120

## Description
The implementation would try to simplify the way the settings file is created and read adapting it to the PKNCA objects.
Based on feedback will be in principle independent from the original data... 

## Definition of Done

- [ ] Settings upload should find the dose number (profile), analyte and pcspec for the settings, and update the corresponding widgets in settings tab so that these are selected in the selectInput

- [ ] Settings upload should work fine even if no inclusions/exclusions are considered

## How to test
1) Perform different NCA settings, run NCA & download settings on results page 
2) Reinitiate the App > Map data > upload the file in NCA settings > Check all relevant custom changes were reconsidered

How to test features not covered by unit tests.

## Contributor checklist
- [ ] Code passes lintr checks
- [ ] Code passes all unit tests
- [ ] New logic covered by unit tests
- [ ] New logic is documented
- [ ] Package version is incremented

## Notes to reviewer

Anything that the reviewer should know before tacking the pull request?

PR #274, opened 2025-03-17
Bug/settings for pknca

### LINKED ISSUE #120: Bug: Settings upload does not update dose number

Gero1999:
## Issue
- Settings upload does not update dose numbers in Settings Tab of NCA
- Settings seem to not work when no inclusions/exclusions are in the settings file considered
- Settings need to be updated to include multiple analytes

Closes #120

## Description
The implementation would try to simplify the way the settings file is created and read adapting it to the PKNCA objects.
Based on feedback will be in principle independent from the original data... 

## Definition of Done

- [ ] Settings upload should find the dose number (profile), analyte and pcspec for the settings, and update the corresponding widgets in settings tab so that these are selected in the selectInput

- [ ] Settings upload should work fine even if no inclusions/exclusions are considered

## How to test
1) Perform different NCA settings, run NCA & download settings on results page 
2) Reinitiate the App > Map data > upload the file in NCA settings > Check all relevant custom changes were reconsidered

## Contributor checklist
- [ ] New logic is documented

## Notes to reviewer


Gero1999:
I remade & modularise the code into two files: `save_settings` /  `load_settings`. Now the code works dynamically as well with `mydata` object (before it used `res_nca`)

- All savings seem correct
- All loadings seem correct for all NCA-setup inputs

The code to load the slope adjustments from the settings files is also there and working, but I am missing how to link it with the variable `manual_slopes()`. However, in order to prevent what happened last time with this issue (#205) I would like to not create many conflicts with other branches and keep this for later once we have a final/functional version of the slope selection section

Let me know your thoughts

inst/shiny/modules/load_settings.R
```diff
@@ -0,0 +1,339 @@
...
+        # If there were changed rows update the concentration data & exclusions table
+        if (nrow(changed_rows) > 0) {
+
+          # Update the main object value
+          data$conc$data <- new_conc_data
+          mydata(data)
+
+          # Update the manual slopes object
+          conc_groups <- unname(unlist(mydata()$conc$columns$groups))
+          manual_slopes <- new_conc_data %>%
```
js3110, line 289: Have you tried to return this as an output for the server? then in nca_setup it can be also returned and then inputted to the manual_slopes server module? 

I think if possible, we should try to complete this within this issue, since manual slopes is one of the main things that settings should be handling. But if its too complex, its best in another issue!

js3110 (changes requested):
Hi Gerardo, I've had a look but I am unable to download the xlxs file (its saying site wasn't available and trying to save as .htm), and I also wasn't able to upload the .rds file as it said I have exceeded maximum file size. So I can't test whether or not it works!

I also added some comments / questions in the code, please take a look :)

js3110:
> Hi Gerardo, I've had a look but I am unable to download the xlxs file (its saying site wasn't available and trying to save as .htm), and I also wasn't able to upload the .rds file as it said I have exceeded maximum file size. So I can't test whether or not it works!

For info, I saved settings with two analytes and two doses, and added one extra parameter.

inst/shiny/modules/load_settings.R
```diff
@@ -0,0 +1,339 @@
...
+              dose = setts$dose_columns %>%
+                filter(ind == "dose") %>%
+                pull(values),
+              time.nominal = setts$dose_columns %>%
+                filter(ind == "time.nominal") %>%
+                pull(values),
+              exclude = setts$dose_columns %>%
+                filter(ind == "exclude") %>%
+                pull(values),
+              duration = setts$dose_columns %>%
```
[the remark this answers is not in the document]
Gero1999: i don't use it but I will add it for simplicity

inst/shiny/modules/load_settings.R
```diff
@@ -0,0 +1,339 @@
...
+      intervals_userinput_setts <- setts$intervals %>%
+        ungroup() %>%
+        filter(type_interval == "manual") %>%
+        left_join(setts$dose$data, by = c(unlist(unname(setts$dose$columns$groups)), "DOSNO")) %>%
+        mutate(start = start - !!sym(dose_time_col),
+               end = end - !!sym(dose_time_col)) %>%
+        select(start, end) %>%
+        filter(!duplicated(paste0(start, end)))
+
+      if ((nrow(intervals_userinput_setts) > 0)) {
```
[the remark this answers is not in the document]
Gero1999: hmmm I would prefer to only do it if is reviewed/merged before this pr. We can open a new issue afterwards if it is not

inst/shiny/modules/tab_nca/upload_settings.R
```diff
@@ -0,0 +1,342 @@
...
+            datatable(mismatched_points %>% select(-IX))
+          })
+        }
+
+        # If there were changed rows update the concentration data & exclusions table
+        if (nrow(changed_rows) > 0) {
+
+          # Update the main object value
+          data$conc$data <- new_conc_data
+          mydata(data)
```
m-kolomanski, line 288: issue: Data object is not a reactive value, should not be updated directly. It is a reactive. When this module updates the widgets, it should trigger the correct events in `nca_setup.R` that will update returned rules.

inst/shiny/modules/tab_nca/upload_settings.R
```diff
@@ -0,0 +1,342 @@
...
+}
+
+#' Load Settings Server Module
+#'
+#' This module handles the server-side logic for loading project settings from a file,
+#' including updating the UI elements: selectInputs, parameters, imputations, slope adj...
+#'
+#' - id The module's ID.
+#' - mydata A reactive expression containing the project data to be updated.
+#' - parent_session The parent Shiny session.
```
m-kolomanski, line 21: issue: Please give more context as to why the session is passed. I know why because I saw the previous iterations, but this might be confusing for someone joining the project.

inst/shiny/modules/tab_nca/upload_settings.R
```diff
@@ -0,0 +1,342 @@
...
+        }
+      }
+      ########################################################################################
+    })
+  })
+}
+
+# Create a function to update the selectInputs based on the matches between settings and data
+# If there are mismatches send a notification to the user
+.update_select_with_setts <- function(var_setts_col, var_data_col, setts_df,
```
m-kolomanski, line 312: nitpick, subjective: The function name is not really clear to me, maybe something like `.update_nca_setup_widgets` would be better?

inst/shiny/modules/tab_nca/download_settings.R
```diff
@@ -1,79 +1,106 @@
...
+#' Save Settings Server Module
+#'
+#' The module handles the server logic for saving project settings from NCA setup & slope adj.
+#' The file can be downloaded in two formats: RDS or XLSX.
+#'
+#' - id The module's ID.
+#' - mydata A reactive expression containing the project data to be saved.
+#' - parent_session The parent Shiny session.
+#'
+download_settings_server <- function(id, mydata, parent_session) {
```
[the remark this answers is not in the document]
Gero1999: hey @m-kolomanski ! just to know how to make a better solution, the problem with `parent_session` is because it is a heavy object for the memory of the module? Or is more an issue of misunderstanding because of the lack of comments? 

Another solution I can think of is just taking the character vector `input_names_parent_session` as an arg. Then using regex I can get the corresponding inputs in the same way session does. Will that be a better solution?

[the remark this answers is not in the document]
Gero1999: @m-kolomanski that will work for the `download_settings`, but in the `load_settings` I am using it to update many other inputs (select_analyte, select_dosno, select_pcspec, method, nca_params, should_impute_c0)

What should I do then for those? I can personally think of:

1) Guessing the parent session namespace using regex on the current
2) Instead of doing it a module, doing it simply a sourced code script, if that makes sense (?)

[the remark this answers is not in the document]
m-kolomanski: So in the case of loading the settings, passing the session makes more sense, since updating the inputs directly would be good from the performance perspective, thats why I have not commented there.

But you might consider scraping the module and simply creating an R function for gathering all options from the uploaded file, then having `nca_setup` read the file, gather the settings and update its own widgets.
```R
# nca_setup.R
observeEvent(inpu$settings_upload, {
  settings_file <- readRDS(input$settings_upload$datapath) # or if-else statement supporting whatever file extensions

  uploaded_settings <- extract_settings_from_file(settings_file)

  purrr::iwalk(uploaded_settings, \(x, y) .update_select_with_setts(...))
})
```

But you could also have the `load_settings` return a reactive with the list of uploaded settings, then update the widgets in `nca_setup`:
```R
uploaded_settings <- upload_settings_server("upload_settings", mydata, session, auc_counter, manual_slopes)

observeEvent(uploaded_settings(), {
  purrr::iwalk(uploaded_settings(), \(x, y) .update_select_with_setts(...))
}}
```

I know I have pushed for modularizing as much as possible, and app structure looks great to me atm, but if two modules would need to be so coupled together to the extent of modifying each others inputs directly, then it should probably be just one module.

inst/shiny/modules/tab_nca/upload_settings.R
Gero1999: it goes one by one, I have rename it to `.updateSelectInput_with_setts`, to make it sound more similar to the shiny function it uses. Hope this seems more intuitive now, if not let me know!

inst/shiny/modules/load_settings.R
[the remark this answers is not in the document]
js3110: it is now merged so you can add it!

inst/shiny/modules/tab_nca/upload_settings.R
```diff
@@ -0,0 +1,382 @@
...
+#'
+#'Its arguments are:
+#' - id The module's ID.
+#' - processed_pknca_data A reactive expression containing the project data to be updated.
+#' - parent_session The parent Shiny session (tab_nca). Used to update all mentioned inputs
+#' - auc_counter A reactive value for counting AUC intervals.
+#' - manual_slopes A reactive value for handling manual slopes.
+#'
+upload_settings_server <- function(id, processed_pknca_data, parent_session,
+                                   auc_counter, manual_slopes) {
```
js3110, line 31: Instead of having `manual_slopes ` as an input, I would create it as an output of the module
Then, when you call the server, define `manual_slopes `(similar to with the rules, and processed data), then that can be used as an input to the `slope_selector_server`

inst/shiny/modules/tab_nca/upload_settings.R
```diff
@@ -0,0 +1,382 @@
...
+      if (nrow(setts$conc$data) > 0) {
+
+        # Using the exclusions/selections of the settings update the concentration dataset
+        new_conc_data <- dplyr::left_join(
+          data$conc$data,
+          setts$conc$data %>%
+            select(
+              any_of(c(conc_setts_cols, "is.included.hl",
+                       "is.excluded.hl", "exclude_half.life", "REASON")
+              )
```
js3110, line 260: this won't work because the included and excluded values are reset in `filter_slopes()`, which is called before the NCA is run. The function uses `manual_slopes `to then modify the inclusions and exclusions, so they need to be defined there instead.

So use the inclusions and exclusions to create a manual_slopes object here, that you then return as an output?

inst/shiny/modules/tab_nca/upload_settings.R
```diff
@@ -0,0 +1,382 @@
...
+          })
+        }
+
+        # If there were changed rows update the concentration data & exclusions table
+        if (nrow(changed_rows) > 0) {
+
+          # Update the manual slopes object
+          data$conc$data <- new_conc_data
+          conc_groups <- unname(unlist(processed_pknca_data()$conc$columns$groups))
+          manual_slopes <- new_conc_data %>%
```
js3110, line 307: this is good, just need to actually return it because here it is being created but then nothing happens.
Also maybe double check if its `Type `and `Range `or `TYPE `and `RANGE`

m-kolomanski:
The issue should be handled in #422, @Gero1999 do you think we can we close this one?

PR #280, opened 2025-03-20
fix: crash na interval_remove_impute

### LINKED ISSUE #279: Bug: crash for impute method NA
### Description
intervals_helpers function causing a crash if impute method is NA

### To reproduce
Upload dataset

### Expected behaviour
App should work

Gero1999:
## Issue

Closes #279

## Description
intervals_helpers function causing a crash if impute method is NA as argument --> 
I modified the two functions behaviour over NA/empty strings to return the same input and I added a test for each 

## Definition of Done
App should work even if start_impute default method is NA

## How to test
Use a dataset with NA by default in impute (i.e, already a start concentration present at dose time)

I added a row to the dummy complex data so we can track this from now on! USUBJID 11111, DOSNO 1, PCSPEC Serum (Infusion example)

## Contributor checklist
- [x] New logic is documented


js3110 (approved):
LGTM! and works with my big dataset, thanks!

js3110 (approved):
LGTM! thanks

m-kolomanski (approved):
LGTM, thank you for the fix!

PR #282, opened 2025-03-21
Bug: Partial AUCs crashing

### LINKED ISSUE #249: Bug: Partial AUCs problem when start is not 0

js3110:
## Issue

Closes #249 

## Description

Partial AUCs were crashing and not working. I have rearranged to create a table instead of numeric INputs.
Works for all types of intervals (as long as there are data points within these intervals
## Definition of Done

- [x] Partial AUCs work well for all types, eg 0 to 12 or 12 to 24

## How to test

Add partial AUC to table

## Contributor checklist
- [x] New logic is documented


Gero1999 (changes requested):
Hey thanks for the changes, I feel this solution/enhancement should end up being better.! I also like the UI changes. Only some small points:

- [x] question: is this now intended to be merged in `main.R` instead of `refactor/tab-nca-r`?
- [ ] suggestion: for the names/labels of the inputs I would personally prefer `start`-`end` instead of `min`-`max`
- [ ] issue/bug: Add row does not work for me. These are the two situations I can report
-  When the initial inputs are not filled the row is not added (maybe this is intentional?)
- When inputs are filled and then I press add row the app crashes:

```
Warning: Error in pmap: ℹ In index: 1.
Caused by error in `.f()`:
! unused arguments (Min = .l[[3]][[i]], Max = .l[[4]][[i]])
  131: <Anonymous>
  130: signalCondition
  129: signal_abort
  128: rlang::abort
  127: cli::cli_abort
  126: h
  125: .handleSimpleError
  124: call_with_cleanup
  121: pmap_
  120: pmap
  119: eventReactiveValueFunc [modules/tab_nca/nca_setup.R#516]
   75: mydata
   68: <reactive:plot_data> [modules/tab_nca/slope_selector.R#298]
   52: r
   46: observe
   45: <observer>
    2: shiny::runApp
    1: aNCA::run_app [C:/Users/rodrig60/Documents/aNCA/R/run_app.R#46]
```

Gero1999 (changes requested):
> In regards to naming, i can't name them start and end since the other variables start and end are used when creating the auc_data : `intervals_list <- pmap(auc_ranges, function(min, max) { intervals %>% mutate( start = start + min, end = start + (max - min), across(where(is.logical), ~FALSE), aucint.last = TRUE, type_interval = "manual" ) })`


Hey regarding this I think there should be no problem, you can keep if you want in the pmap arg `function(min, max)`, it won't produce problems. I just would like that for the user (UI) and if possible also internally `auc_ranges` has `start` & `end` (or similar words i.e, `aucint_start` `aucint_end`)

These terms are more familiar for PKNCA users and can make the code easier for developers as well

inst/shiny/modules/tab_nca/nca_setup.R
```diff
@@ -12,191 +12,188 @@ nca_setup_ui <- function(id) {
...
+                  ns("aucpext.obs_threshold"),
+                  "",
+                  value = 20,
+                  step = 1,
+                  min = 0,
+                  max = 100
+                )
+              )
             )
           )
```
m-kolomanski, line 140: suggestion(non-blocking):
I would consider wrapping that into a function:
```
conditional_numeric_input <- function(
  session = shiny::getDefaultReactiveDomain(), id, label, value, step, min, max
) {
  ns <- session$ns
  rule_id <- paste0("rule_", id)

  tagList(
    column(
      width = 6,
      checkboxInput(ns(rule_id), paste0(label, ":"))
    ),
    column(
      width = 6,
      conditionalPanel(
        condition = paste0("input['", ns(rule_id), "'] == true"),
        div(
          class = "nca-numeric-container",
          numericInput(
            ns(id),
            "",
            value = value,
            step = step,
            min = min,
            max = max
          )
        )
      )
    )
  )
}
```

but I recognize this is cosmetic and this PR is not about refactoring, so this is non-blocking suggestion, we can leave this for another day

m-kolomanski (approved):
Hey, good job with the code, menu is looking nice too!

I am accepting the PR, but there is one issue I would like you to consider before merging:

Currently the whole `processed_pknca_data` is invalidated on every change to `auc_data()`, meaning that when the user provides double digits it will be recalculated twice - e.g. for 24h, once when the user inputs **2** and the second time when *4** is inputted. To remedy that, please add `shiny::debounce(500)` to the `processed_pknca_data` reactive:

```R
    processed_pknca_data <- eventReactive(list(
      input$method, input$nca_params, input$should_impute_c0,
      input$select_analyte, input$select_dosno, input$select_pcspec,
      mydata(), session$userData$units_table(), auc_data()
    ), {
        # code
    } %>%
      shiny::debounce(500)
```

This will give the user some time to input the whole number before the reactive is invalidated.

Also, make sure the checks are passing before merging.

In addition, I have left one cosmetic, non-blocking suggestion.

js3110:
> Currently the whole `processed_pknca_data` is invalidated on every change to `auc_data()`, meaning that when the user provides double digits it will be recalculated twice - e.g. for 24h, once when the user inputs **2** and the second time when _4_* is inputted. To remedy that, please add `shiny::debounce(500)` to the `processed_pknca_data` reactive:


Thanks @m-kolomanski ! I am actually going to leave this, because I have conveniently already added shiny debounce (and some refactoring) in #294 , so theres no point in making a worse version of it here!

In regards to the other comment, these section of UI is not related to this PR (as thats rule sets and not partial AUCs), so I think it would make more sense to do refactoring of those parts in a different issue? I can write it now!

m-kolomanski:
@js3110 Alright, all clear 👍 I am not sure if adding the issue is necessary - we might as well do so, but set it to very low priority, as there is nothing wrong inherently with the code.

js3110:
> @js3110 Alright, all clear 👍 I am not sure if adding the issue is necessary - we might as well do so, but set it to very low priority, as there is nothing wrong inherently with the code.

Ok, sounds good, maybe I add it to a similar issue for refactoring the parameter selection, then its two in one.
Also, while I have you here- can you see why the check is failing for this PR? Its saying its failed but I can't see any errors so not sure what the problem is @m-kolomanski

js3110:
@Gero1999 I updated the naming, PTAL :)

m-kolomanski:
@js3110 It is failing because #289 updated the CI to fail at warnings. You will need to either rebase or merge `main` into your branch.

Eh, TBH I think since we know what the issue is and there is nothing wrong with your PR (I believe), we could be lenient on this and just merge with checks failing, if the rebasing/merging process is annoying.

js3110:
@Gero1999 , fixed conflicts, ready for testing!

Gero1999 (approved):
lgtm! :D

PR #285, opened 2025-03-24
Enhancement: Add labels for parameters & include CDISC.PPTESTCD for PP/ADPP

### LINKED ISSUE #261: Enhancement: Add labels for parameters in PKNCA
### Description
Changes to improve the display in `NCA results` setion. 

We count with a `generate_col_defs()` that creates hovering column names for reactable objects. We would like that the `pivot_wider_pknca()` functions already provides labels to the parameter columns. It would be also advantegous if the picker input allows the user to choose these clean column names instead of the literal ones. 

### Definition of Done
- [ ] Table in `NCA results` display hovering messages for each parameter column (i.e, for `r.sq.adj[unitless] > `Adjusted R Squared``). If existent, use a CDISC nomenclature for the metadata. 
- [ ] Picker input of that table for the parameters works with the clean attr names as choices.
- [ ] PP/ADPP should include the CDISC labels as their reported PPTESTCD values
- [ ] Eliminate fraction & unitless from  r.sq.adj[unitless] > r.sq.adj

> This is blocked until merging #259

Gero1999:
## Issue

Closes #261

## Description

For both implementations I created the same function `translate_terms` from one nomenclature to another using a specified mapping file (added in new folder: `www/metadata/Parameter_labels_PKNCA-CDISC.csv`)

### Labels for parameters in NCA results section
We count with a generate_col_defs() that creates hovering column names for reactable objects. We would like that the pivot_wider_pknca() functions already provides labels to the parameter columns. It would be also advantegous if the picker input allows the user to choose these clean column names instead of the literal ones.

> Note: Also as requested by @Shaakon35 parameter names `r.sq.adj` & `span.ratio` won't include the redundant "[unitless]" or "[fraction]" in their names

### CDISC in Parameter Datasets PP/ADPP
All names will be generated using now `translate_terms` to provide all official CDISC terms 

## Definition of Done
- [x] User can now hover parameters in `Results NCA`
- [x] Empty units are not longer shown i.e, "r.sq.adj[unitless]" > "r.sq.adj" 
- [x] Select Input for results NCA allows now choosing pretty names for the user

## How to test
Map data > Run NCA > Check Results NCA table 
Map data > Run NCA > Check Parameter Datasets > PP/ADPP

## Contributor checklist
- [x] New logic is documented


inst/shiny/www/metadata/Parameter_labels_PKNCA-CDISC.csv
m-kolomanski, line 1: issue: Please do not use `www/` for storing data. `www/` is for files that should be accessible directly by the browser, and this .csv file should not.

inst/shiny/modules/translate_nomenclature.R
```diff
@@ -0,0 +1,29 @@
...
+#' @param target_col Character indicating the column name in the metadata file of the target terms.
+#' Default is "CDISC".
+#' @return A character vector of translated terms. Input terms not available in mapping_col will be
+#' returned with the same value.
+#' @examples
+#' input_terms <- c("adj.r.squared", "ae", "nonexistent_term")
+#' translate_terms(input_terms)
+translate_terms <- function(input_terms, mapping_col = "PKNCA", target_col = "CDISC") {
+  # Load the metadata file
+  metadata <- read.csv("www/metadata/Parameter_labels_PKNCA-CDISC.csv")
```
m-kolomanski, line 18: issue: Do not load files upon function call unless necessary - this way, every time this function is called, the app will waste time on loading the file.

suggestion: Instead of adding this file as a part of the application, I think it would be worth to attach this file as a data object to the package, so that both the app and the command-line user can reference it directly, as it is done with datasets like `mtcars` or `iris`.

inst/shiny/modules/translate_nomenclature.R
```diff
@@ -0,0 +1,29 @@
...
+
+  # Create a named vector for translation
+  translation_vector <- setNames(metadata[[target_col]], metadata[[mapping_col]])
+
+  # Translate terms
+  translated_terms <- ifelse(input_terms %in% names(translation_vector),
+                             translation_vector[input_terms],
+                             input_terms)
+
+  return(translated_terms)
```
m-kolomanski, line 28: nitpick: no need for explicit return
```suggestion
  translated_terms
```

inst/shiny/modules/translate_nomenclature.R
```diff
@@ -0,0 +1,29 @@
...
+#'
+#' This function translates a character vector of terms from one nomenclature to another using a
+#' mapping file.
+#'
+#' @param input_terms A character vector of terms to be translated.
+#' @param mapping_col Character indicating the column name in the metadata file of the input terms.
+#' Default is "PKNCA".
+#' @param target_col Character indicating the column name in the metadata file of the target terms.
+#' Default is "CDISC".
+#' @return A character vector of translated terms. Input terms not available in mapping_col will be
```
m-kolomanski, line 11:
```suggestion
#' @returns A character vector of translated terms. Input terms not available in mapping_col will be
```

m-kolomanski (changes requested):

inst/shiny/www/metadata/Parameter_labels_PKNCA-CDISC.csv
Gero1999: I used `usethis::use_data()` and created a new folder `/data` in the working dir. Is that ok? 

If not will still the lazy loading work if I do a direct call in the documentation `data.R`?

R/export_cdisc.R
```diff
@@ -106,31 +106,11 @@ export_cdisc <- function(res_nca) {
     ungroup() %>%
     #  Recode PPTESTCD PKNCA names to CDISC abbreviations
     mutate(
-      PPTESTCD = recode(
```
Gero1999, line 109: issue: `.pptestcd_dict` is also not needed if we use `translate_terms()`

inst/shiny/www/metadata/Parameter_labels_PKNCA-CDISC.csv
m-kolomanski: I would use `data/` if we want this to be accessible by the whole package and `inst/shiny/data/` if this is data related to the application itself, like dummy data to be loaded. I think putting this file in `data/` is fine.

Gero1999: that makes a lot of sense, thanks!

Shaakon35 (approved):

Gero1999:
Hey @m-kolomanski now this should be fine. By request from Valentin, I made some additional changes since your review, most relevant ones:

- **Changed all labels for parameters in all sections.** Now everything should be consistent with CDISC standards. I basically just modified the PPTESTCD column in `res_nca()`. I took into consideration the transformation process in the `units_table` for this.

- **Changed flag rules application in final_results** The new PPTESTCD names forced me to do it. I used the opportunity to simplify a little bit this part using the reactive `rules`.

m-kolomanski (approved):
Hey, the PR looks great!

I just wanted to inquire about the name of the data file - what does `dict` stand for? For Python users, this could get confused with `dictionary` data type, while the actual data is a data frame. I would suggest more neutral name, maybe `pknca_cdisc_terms` or similar?

PR #286, opened 2025-03-24
Enhancement: bioavailability calculations 

### LINKED ISSUE #203: Enhancement: bioavailability calculations

js3110:
## Issue

Closes #203
Replaces #273 

## Description

Currently PKNCA is unable to calculate parameter F (bioavailability). Until that feature has been implemented, we want a temporary function to calculate bioavailability. This should include the following options:

F should be written as a percentage %

The app should be able to automatically determine if there are subjects with both IV and EX routes for each group (taken from `grouping_vars`), or not, and decide which of 1 or 2 it will calculate. The mean should be calculated by the grouping vars. The results should be along the main results. The type of AUC (mean or individual) should also be included in the results.

## Definition of Done

- [x] Bioavailability calculation for individuals, and grouped data
- [x] Automatic calculations with only user input for parameters
- [x] Results table with ratios displayed
- [x] Ratios included in descriptive statistics

## How to test

Use dummy data DummyRO_MRoutes, select bioavailability. Check main results and descriptive statistics for f_ parameters.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

This is a replica of #273 as rebasing with the new main was too complicated

I haven't written any unit tests, as I am not sure the best way to create a PKNCAresults object for testing.
Since we will need this to test other functions as well, maybe its better to create a separate issue to create a testing PKNCAresult?

### DIFF
--- a/R/zzz.R
+++ b/R/zzz.R
@@ -17,6 +17,7 @@
     "log10_SE", "next_dose", "next_nom_dose",
     "nom_dose_time", "nom_interval_next", "nom_interval_prev",
     "start", "start_dose", "time_dose", "type_interval", "var_name", "view",
-    "zero_str", "PPTESTCD", "PPSTRESU", "PPTESTCD_unit"
+    "zero_str", "PPTESTCD", "PPSTRESU", "PPTESTCD_unit", "Grouping_EX", "Route", "Dose",
+    "Grouping_IV", "AUC_IV", "Dose_IV", "AUC_EX", "Dose_EX", "pknca_cdisc_terms", "PPTESTCD_cdisc"
   ))
 }


m-kolomanski (changes requested):
Hey, good job on the PR, good separation between logic and application - I like that we keep the changes in the modules to a minimum. I have left some comments, mostly suggestions, there is an issue with lacking documentation.

Also, it would be great to add unit tests.

We can create re-usable fixtures for testing using `testhat`. I do not think creating a separate issue for that is necessary - you could start off in this PR, and then subsequent PRs could use those fixed PKNCA objects for their tests. And then we can refactor tests for previously merged functionalities while we are at them.

To do this, you can simply create `tests/testthat/setup.R` and define any global fixtures there, like `PKNCAconc_fixture`, `PKNCAdose_fixture`, `PKNCAresults_fixture` etc. Those variables / constants will be available in all test files.

js3110:
> To do this, you can simply create `tests/testthat/setup.R` and define any global fixtures there, like `PKNCAconc_fixture`, `PKNCAdose_fixture`, `PKNCAdata_fixture`, `PKNCAresults_fixture` etc. Those variables / constants will be available in all test files.

Hi @m-kolomanski for this, is it ok to use an uploaded dummy dataset as the input to create the PKNCAdata?
Can we load it from the data folder already, or do I need to upload a separate version in the tests folder?
Manually coding a dummy data is doable for the simple functions, but for things like bioavailability becomes far too complicated

m-kolomanski:
@js3110 _Ideally_ we would want separate testing dataset in `tests/testhat/data/` folder, BUT I think there are benefits in loading the same data we have in the app. If at some point we decide to change the mock data, our tests will run on the up-to-date version.

If we want the testing data to cover cases that are not present in the mock data, then I would go for creating a copy. As new functions come, we can then adjust it without changing the in-app experience.

Otherwise, if it is gonna be exactly the same data as we have in app, I think we can just load the mock data from the app.

tests/testthat/setup.R
m-kolomanski, line 1: issue: Please rename the variables to indicate that those are global constants shared across the tests, preferably by using UPPERCASE and adding `_FIXTURE` at the end to indicate what this variable is.

Variables in `setup.R` are defined as global values for all tests suites. This file is run only once, at the start of the testing. If a test file assigns a variable of the same name, it will get overwritten - and this might cause a lot of mess down the testing line. Therefore, to be clean, the names for the fixture need to be unique and indicate what those values are and where to look for them.

m-kolomanski (changes requested):
Hey, the code and tests look good, thank you for all the work! I have one issue to do with global variable naming, please take look.

m-kolomanski (approved):
LGTM!

Regarding the failing checks, this is due to so-called non-standard evaluation performed by functions from `tidyverse`, like `dplyr`. When you pass column names to functions like `select` or `mutate`, you pass them as variables. R CMD check then looks for declarations of those variables, but cannot find them in the context. In order to make the check pass, you will need to add global binding for those colnames. Please see `zzz.R` file, there is a function that adds the bindings - you will need to add your colnames to the vector.

Gero1999 (approved):
lgtm! thanks

Issue #71, opened 2024-10-11
Enhancement: Add two additionnal plot TLGs


Shaakon35:
### Prerequisite
https://github.com/pharmaverse/aNCA/issues/53

### Description
Based on the enhancement 53, on the same template, please add two more TLGs:
g_pkconc_LIN.R
g_pkconc_mean_LIN.R

### Definition of Done
The requirements for the feature to be complete:
- [ ] the two new TLGs should work as g_pkconc_ind_LIN.R
- [ ] Titles, footnotes, xlabel, ylabel, legend, etc are correctly populated

PR #295, opened 2025-03-27
Feat : Added R/g_pkconc.R and test-g_pkconc.R files

KOBANAK:
## Issue
Matches issue #71  

## Description
Implementation of pkgc02 : Combined PK profiles over time 
https://insightsengineering.github.io/tlg-catalog/stable/graphs/pharmacokinetic/pkcg02.html

Definition of Done
- [x] the TLG work as g_pkconc_ind_LIN.R
- [x] Titles, footnotes, xlabel, ylabel are correctly populated

## How to test
In the TLG tab, click on "Graphs" and pkcg02 - linear or pkcg02 - log  tab

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer
Discussed during aNCA meetings : 
- Minor bug in legend => The legend icon differs for some patients.
- g_pkconc_mean plot not added for now, to add in separate issue and fix plotly issue.


KOBANAK:
Hello Mateusz,
Many thanks for your feedback!

I have updated the files based on your comments.  I agree it is tidier to keep all the pkcg plots within a single file like g_pkcg.R. The PR template is filled out, and I have incremented the package version accordingly.

Apologies for the delay, it took me a bit of time to navigate everything properly.

m-kolomanski:
Hey, technically the submission looks good, thank you!

I would need to ask either @Gero1999 or @js3110 to verify if the plots are correct.

There are small issues I have encountered:
- the max range for log plots seems off, way to high:
- Screenshot from 2025-04-08 15-02-49

- when the user specifies the color variable, the legend is lost (even if still USUBJID is selected):
Screenshot from 2025-04-08 15-02-39


I dunno if that is expected or not.

Other than that, seems like the names for both pkcg01 and pkcg02 functions are quite similar - can that be confusing? Should we rename the functions to `g_pkcg01_lin`, `g_pkcg01_log`, `g_pkcg02_lin`, `g_pkcg02_log` to be more consistent?

Gero1999 (changes requested):
Hey great work! But I confirm some of the small problems found by Mateusz, which are mainly in LOG plots:

- [x] In both plots the default Xlabel / Ylabel are not the labels of the variables but their literal names (AVAL, AFRLT - instead of "Analyte Value", "Act. Rel. Time from Analyte First Dose" 

### General problems affecting also other plots
Perhaps these ones need another issue to solve it everywhere, correct me if I am wrong @m-kolomanski
- [x] TLG order/submission table does not differentiate pkgc01 functions (does not say linear & log)
- [ ] color variable when selected is not showing any legend in any case
- [x] Y Ticks in the log plots are not displayed as expected
- [x] Ymax by default in the logs plots are too high for some reason

m-kolomanski:
@Gero1999 We could create separate issue for the changes you have mentioned, but this PR is not very complex, I think some small changes to `pkcg01` could fit here as well. Up to you, I do not think it will make a lot of difference.

KOBANAK:
Hi Mateusz,

I've updated the function names and the legend now appears when a color variable is selected. As for the Y-axis ticks issue and max default value issue of log plots it will be addressed in a separate issue (discussed today with Gerardo and Valentin). 
Thanks

m-kolomanski (changes requested):
issue: Both `ggh4x` and `scales` were recently removed from `Imports`, since they are used only in one situation - the **SBS** view, which is not even used in application. Therefore those packages were moved to `Suggested` and appropriate checks were applied to verify that the user has those packages installed in a case they call it from console:
```R
  if (scale == "SBS") {
    if (!requireNamespace("ggh4x", silently = TRUE))
      stop(
        "Side-by-side view requires `ggh4x` package, please install it with ",
        "`install.packages('ggh4x')`"
      )
    if (!requireNamespace("scales", silently = TRUE))
      stop(
        "Side-by-side view requires `scales` package, please install it with ",
        "`install.packages('scales')`"
      )
    ## ...
  }

``
Please see what the code for `g_pkcg01` looks like on the `main` branch.

I see that this PR a) removes the checks and b) imports the two packages back. I see no reason for that - please move the packages to Suggests, remove `@import` statements from documentation and reinstate the checks as above.

Gero1999 (approved):
lgtm! :)

m-kolomanski (changes requested):
Hey, thank you for all the changes, looks very nice, but I still have a couple of comments.

The log scale plots seem off, is this what we are going for? Is this expected from the dummy data?
Screenshot from 2025-05-05 07-31-46

KOBANAK:
Thank you for the previous feedback. I've incorporated the suggested changes and upated the g_pkcg for 100% test coverage

m-kolomanski (approved):
Hey,

LGTM, please make sure everything is working after resolving conflicts.

Thank you for all the changes and all the work!

PR #301, opened 2025-03-31
Bug: Manual slopes saying manual for all profiles

### LINKED ISSUE #283: Bug: Manual Slope selector changes all results to manual
### Description
When adding a manual slope selection in the slopes tab, the numeric results are fine (i think), but all of the subjects have the slope type renamed to 'Manual' 

### To reproduce
Add a manual slope selection, view results in Slopes tab

### Expected behaviour
Only the slopes changed should be labelled manual, all other should be Best Fit

js3110:
## Issue

Closes #283 

## Description
When adding a manual slope selection in the slopes tab, the numeric results are fine (i think), but all of the subjects have the slope type renamed to 'Manual'. In addition, the method column is no longer in the slopes results.

## Definition of Done

- [x] Manual slopes only change method of individual profiles to manual
- [x] Method column in slopes results

## How to test

Add manual slopes

## Contributor checklist
- [x] New logic is documented


m-kolomanski:
Hey, would you mind providing more details as to how to test the changes? I do remember there being a part of the application which provided the info if the slopes were manual or not, but I am gonna be honest - I am testing it on `main` and I cannot find it anywhere, neither in the app UI nor the data itself.

js3110:
Hi @m-kolomanski , if you run the results, then click on Slopes in the results section, then a table appears. Scroll to the right and there is a columbn "LAMZMETHOD" which will then tell you the method

Gero1999 (changes requested):
Hey, good catch! Only one small nitpick, if you want to also made up CDISC names for our own created `lambda.z.ix` and `lambda.z.method` make sure they follow the standard requirements (i.e, `LAMZMETHOD` should be max 8 char).

Gero1999:
> on Slopes Information in the results section, then a table appears. Scroll to the right and there is a columbn "LAMZMETHOD" which will then tell you the method

Just to add more context to @m-kolomanski  in the testing, any user changes in `NCA Setup > Slope Selector` must be informed in the results. If any selection/exclusion was performed, then in `Results > Slopes Information` the corresponding profile will have `LAMZMETHOD = "Manual"`. If the user made no modification, "LAMZMETHOD" = "Automatic"

m-kolomanski:
@js3110 The `main` branch does not have such a column.

Screenshot from 2025-04-01 12-41-14

or at least does not show for me for some reason?

I see now that this branch does have a **LAMZMTD** column, seems to work great.

However, since this is described as a bug, I was looking for the _before_ state of the app to see how it behaves currently, but seems like this info is not displayed at all? So is it a _bug_ as in something is not working or an _enhancement_ as in we are adding the display?

Either way - it does not matter much to me, but I would appreciate more detailed descriptions on how to test the feature, what to look for, how to verify that the changes actually work. It caused me a tiny bit of confusion, I think if a person with less familiarity was to review the PR they could struggle.

Gero1999 (approved):
super good, thanks! 💪🏽 🙌🏽

Issue #242, opened 2025-02-28
Bug: Use nominal time instead of actual time for ratios


js3110:
### Description
Actual time causes data to be filtered out when they should be able to calculate a ratio. Instead use Nominal time as this is more clear about when the sample should have been taken, so the ratios should be calculated.

### Expected behaviour
Nominal time used to determine when specimens have a shared time point, to calculate ratio

PR #303, opened 2025-04-01
Bug: matrix ratios units and time refactor

### LINKED ISSUE #243: Bug: weird units conversion for tissue ratios
### Description
Tissue ratios sometimes convert to units that mean the ratio value is really small

### Expected behaviour
Remove units from ratio calculations, instead just have columns that info the units for each specimen, then the ratio is just 1:1 of units.

js3110:
## Issue

Closes #243 and #242

## Description

Due to issues with the `units` package, ratios were being unnecessarily converted to smaller values. Since this problem is with another package, the easiest solutiion is to ignore the units and just calculate 1:1 units for the ratios.
Also changing the time column used from TIME to NFRLT, as this makes more sense in regards to ratio cacluations, since they are matched by time of sample.

## Definition of Done

- [x] remove units calculation from ratios
- [x] Display units of each specimen
- [x] Calculate ratios correctly
- [x] Calculate with nominal instead of actual time

## How to test

NCA -> Additional Analysis -> calculate a ratio

## Contributor checklist
- [x] New logic is documented


m-kolomanski:
Hey, curious to know what is the solution if matrices are stated in different units, like `ng/mL` and `ug/mL`? Wouldn't that throw the ratio calculation off? Should we expect the units in the data to be already standardized? Should there be a check if `unique(Spec1_Unit) == unique(Spec2_Unit)`? Do we want to drop the support for multiple units?

I am no domain expert, I get that there were issues with the `units` package calculations, but this change seems a bit odd to me, I am probably missing something and trying to understand.

js3110:
Hi @m-kolomanski , don't worry I get it!
normally, if concentrations are non-solids, they will all be in the same unit (eg ug/mL). It is unlikely that those will be different (except in our dummy data but that is for testing purposes). What is more common, is a non solid in ug/mL and then a solid in ng/g or similar. This is where the conversion was giving smaller values than necessary.

I spoke to our stakeholders about this, and they said that as long as they can see the units for each specimen (which I have now included in Spec1_Units and Spec2_Units), then they are happy because they can know roughly what the ratio unit is. And typically for these types of calculations, no conversion is done and the ratio is simply concentration1/concentration2.

I hope that makes sense! let me know if you have any more questions :)

m-kolomanski (approved):
Thank you for the explanation! The PR looks good to me.

Shaakon35 (approved):

PR #304, opened 2025-04-01
fix: rename parameters for bioavailability

### LINKED ISSUE #203: Enhancement: bioavailability calculations

js3110:
## Issue

Closes #203 

## Description

Fix for unnoticed bug after renaming of parameters

## Definition of Done

- [x] Bioavailability works

## How to test

Check bioavailability with MRoutes dataset

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

I merged and then noticed that actually the bioavailability wasn't working any more after #285 , so had to rename


m-kolomanski:
Hey, what is _MRoutes dataset_? From what I can see, all dummy datasets available in the repo have singular route of administration.

js3110:
Hi @m-kolomanski the DummyMRoutes csv should be in the data folder, although it’s possible it’s uploaded only on my local and not in the remote. I’ll double check!

m-kolomanski (approved):
LGTM

Gero1999 (approved):
lgtm!

PR #305, opened 2025-04-01
Fix: test coverage workflow failing on staged dependencies

m-kolomanski:
## Description

The workflow for test coverage failed because it could not restore system dependencies. In `admiralci` workflow the `run-system-dependencies` is set to `false` as default. In all other workflows we have it set to `true` - this change introduces the same for test coverage workflow.


rabiibouhestine (approved):
LGTM

Gero1999 (approved):
lgtm!

Issue #287, opened 2025-03-24
Enhancement: Add tests for functions to reach 100% coverage


Gero1999:
### Description
We need to focus on creating tests for some functions that lack any test at all. Here is a list of the priorities, once you create a branch that covers one please include it as done in this issue and/or add a comment to this issue. 

The objective is to reach 100% coverage for all files. Feel free to assign yourself to the file/functions you want to work in and crate a branch for it! Also add a ✅ if the work was done and the PR was merged to the main branch. Feel free also to tag/ reach out in this thread for questions

`> covr::package_coverage()`

| Task/Script                          | Completion Percentage | Assigned User |
|--------------------------------------|-----------------------|---------------|
| R/export_cdisc.R                     | 0.00%                 |  @Gero1999              |
| R/flexible_violinboxplot.R           | 0.00%                 | @Gero1999               |
| R/lambda_slope_plot.R                | 0.00%                 |@js3110               |
| R/pivot_wider_pknca_results.R        | 0.00%                 |  @Gero1999              |
| R/run_app.R                          | 0.00%                 |   covrignore  (#362)          |
| R/zzz.R                              | 0.00%                 |  covrignore  (#362)       |
| R/utils.R                            | 58.33%                |  @Gero1999             |
| R/label_operators.R                  | 59.09%                | @jimrothstein               |
| R/bioavailability.R                  | 69.74%                |  @Gero1999               |
| R/general_lineplot.R                 | 77.46%                |   @Gero1999         |
| R/filter_breaks.R                    | 78.79%                |  @jimrothstein         |
| R/g_pkconc_ind.R                     | 79.14%                |  @Gero1999             |
| R/create_start_impute.R              | 83.87%                |   @Gero1999            |
| R/conversions.R            | 92.31%                |  @Gero1999              |
| R/apply_filters.R                    | 94.12%                |  @Gero1999              |
| R/PKNCA.R                            | 94.87%                |    @Gero1999           |
| R/general_meanplot.R                 | 96.10%                |   @Gero1999            |
| R/utils-slope-selector.R             | 96.72%                |  @Gero1999               |
| R/dose_profile_duplicates.R          | 98.39%                |  @Gero1999              |
| R/format_data.R                      | 98.73%                |  @Gero1999              |
| R/calculate_summary_stats.R          | 100.00% ✅            |               |
| R/intervals_helpers.R                | 100.00% ✅            |               |
| R/l_pkconc.R                         | 100.00% ✅            |               |
[...]

Issue #189, opened 2025-02-04
Enhancement: QC of Dataset before NCA to prevent app crashing


js3110:
### Description
Create a "Data QC" tab in the Data section, that will QC the data and ensure that there is nothing that will cause the NCA section to crash. For example, no duplicates in the times for each sample per patient.

It would be good if the user can then select an option to decide how to proceed?
 1. Clean data themselves and reupload 
 2. Manually input value to mutate
 3. Modify duplicated value TIME by + 0.01 to prevent error ???  @Gero1999 what do you think- any suggestions??

### Definition of Done
The requirements for the feature to be complete:
- [ ] Quality control tab in Data Section
- [ ] Checks for duplicated data grouped by USUBJID, ANALYTE, PCSPEC, TIME and DOSNO
- [ ] If duplicated data present, give user a warning and show which rows are duplicated
- [ ] Allow user to select option for how to proceed

If any other bugs in the data are found to cause the app to crash, they can also have checks included in here.

PR #306, opened 2025-04-01
Enhancement: Adding extra variable to PP/ADPP

### LINKED ISSUE #277: Enhancement: Adding extra variable to PP/ADPP
### Description
There are some extra variables that are not mandatory but needed to be reported in:
* TLGs
* PP/ADPP
* Perhaps others (?)
Some of them are defined per study, and most should have a column associated in the data. We need to allow the user to either include a column or write something themselves when relevant (i.e, Sponsor ID). 

### Definition of Done
Some of the input extra optional information needed:
> **PP/ADPC**
- [x] PPSPID: Sponsor ID = "/F:EDT-[STUDYID]_PKPARAM_aNCA"
- [x] PPRFTDTC: Same one as variable PCRFTDTC (NA if PCRFTDTC does not exists), should be a format like this: "2021-01-06T09:25"
- [x] rename the name of the output from "PP_Dataset_2025-03-17.csv" to "PP_[STUDYID]_2025-03-17.csv"
- [x] # missing variables to add in ADPP:
AGE
AGEU
SUBJID = scan(usubjid, 2, '-'); (take the last digits after the last "-")
PARAM = PPTEST;
PARAMCD = PPTESTCD;
PARAMCAT = PPCAT;
AVISIT
- [x] PPGRPID: It is currently : paste(ANALYTE, DOSETRT, PCSPEC, "CYCLE", DOSNO), it should be paste(DOSETRT, "-", PCSPEC, "-", AVISIT of the dose (e.g. DOSNO = 4, the dose corresponding to DOSNO= 4 has AVISIT that we need to use), if AVISIT does not exist, use DOSNO)
- [x] PPSTINT and PPENINT are populated based on Actual times, please populate them based on Nominal times additionally:
if PPSTINT/PPENINT is always in hours, please set it to 'PTxH' (e.g. "PT0H"   or   "PT360H"),
PPSTINT/PPENINT should be NA if there is no interval
- [x] PPORRES  PPSTRESC  PPSTRESN should be rounded with 12 digits (e.g. 12345.1234567, or 1.23456789999) - rounding
- [x] Make sure that the changes in PP are also done in ADPP
- [x] Make sure that PPREASND does not contain more than 200 char
- [x] remove DOMAIN from ADPP
- [x] When downloading PP/ADPP, make sure there is not a weird header in the Excel format

</body>
</html>

Feel free to add more if you can think of them

Gero1999:
## Issue

Closes #277 

## Description
Based on a review of the PP/ADPP files currently generated by the app we notices some of the most important changes that needed to be done to consider/adequately calculate the different variables. 

#### Main changes:
- **Export_CDISC and parameter_datasets R files**. These are the files involved in the generation and display of PP/ADPP
- **Dummy Data**. I included the PPSPID, PCRFDTC, PCGRPID columns. Also recoded USUBJID, so now also includes the STUDYID in its name (as it is standard).
- **test of pkgc01**. This testings were directly dependent on the previous data, in specific in USUBJID, so I also needed to consder that. 
- **Created a new file for format conversion functions.** I created in there an ISO conversion function and added the preexistent get_conversion_factor. 

## Definition of Done
- [x] rename the name of the output from "PP_Dataset_2025-03-17.csv" to "PP_[STUDYID]_2025-03-17.csv"
- [x] When downloading PP/ADPP, make sure there is not a weird header in the Excel format
- [x] PPRFTDTC: Same one as variable PCRFTDTC (NA if PCRFTDTC does not exists), should be a format like this: "2021-01-06T09:25"
- [x] PPSPID: The one provided in the dataset as PCSPID ((if not Sponsor ID = "/F:EDT-[STUDYID]_PKPARAM_aNCA"))
- [x] # missing variables to add in ADPP:
AGE
AGEU
SUBJID = scan(usubjid, 2, '-'); (take the last digits after the last "-")
PARAM = PPTEST;
PARAMCD = PPTESTCD;
PARAMCAT = PPCAT;
AVISIT
- [x] PPGRPID: It is currently : paste(ANALYTE, DOSETRT, PCSPEC, "CYCLE", DOSNO), it should be paste(DOSETRT, "-", PCSPEC, "-", AVISIT of the dose (e.g. DOSNO = 4, the dose corresponding to DOSNO= 4 has AVISIT that we need to use), if AVISIT does not exist, use DOSNO)
- [x] PPSTINT and PPENINT are populated in 'PTxH' format (e.g. "PT0H"   or   "PT360H"), They are NA if not AUCINT parameters.
PPSTINT/PPENINT should be NA if there is no interval
- [x] PPORRES  PPSTRESC  PPSTRESN should be rounded with 12 digits (e.g. 12345.1234567, or 1.23456789999) - rounding
- [x] Make sure that the changes in PP are also done in ADPP
- [x] Make sure that PPREASND does not contain more than 200 char
- [x] remove DOMAIN from ADPP

## How to test
Map columns in data > NCA setup > Select multiple doses (dosno) > Run NCA > Parameter Datasets > PP/ADPP

Visually check one by one that all Definition of Done aspects are correctly covered!

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer


R/export_cdisc.R
```diff
@@ -126,28 +154,63 @@ export_cdisc <- function(res_nca) {
...
+    common_prefix_length <- i
+  }
 
-  return(list(pp = pp, adpp = adpp))
+  # Return the common prefix if it exists, otherwise return an empty character vector
+  if (common_prefix_length == 0) {
+    return(character(0))
+  } else {
+    return(substr(sorted_strings[1], 1, common_prefix_length))
+  }
```
m-kolomanski, line 215: suggestion:
```suggestion
  if (common_prefix_length == 0) 
    return(character(0))
  
  substr(sorted_strings[1], 1, common_prefix_length)
  }
```

R/format_conversions.R
m-kolomanski, line 1: suggestion: Maybe at this point the file name should be just `conversions` or `unit_conversions` or maybe `units`? WDYT?

tests/testthat/test-format_conversions.R
```diff
@@ -49,3 +49,73 @@ describe("get_conversion_factor", {
...
+    expect_equal(convert_to_iso8601_duration(Inf, "d"), "PinfD")
+    expect_equal(convert_to_iso8601_duration(NA, "d"), NA)
+  })
+
+  it("handles unsupported units gracefully", {
+    expect_error(
+      convert_to_iso8601_duration(10, "unsupported"),
+      "Unsupported unit. Accepted units start with 'y', 'm', 'w', 'd', 'h', or 's'."
+    )
+  })
```
m-kolomanski, line 71: question: What about input like:
```R
r$> convert_to_iso8601_duration(10, "dinvalid")
[1] "P10D"
```
The function takes the first letter and converts the unit based on that. Is that expected?

R/format_conversions.R
```diff
@@ -0,0 +1,81 @@
...
+  # Try to standardize the unit using units::set_units
+  unit <- tryCatch(
+    {
+      units::set_units(1, tolower(unit), mode = "standard") %>% units::deparse_unit()
+    },
+    error = function(e) {
+      # If an error occurs, return the original unit
+      unit
+    }
+  )
```
m-kolomanski, line 70: question: Why do we need this part of the code? I understand that we want to standarize the unit in cases like
```R
r$> units::set_units(1, tolower("minute"), mode = "standard") %>% units::deparse_unit()
[1] "min"
r$> units::set_units(1, tolower("hour"), mode = "standard") %>% units::deparse_unit()
[1] "h"
```
but then we base the whole conversion on the first letter anyway:
```R
r$> substr("minute", 1, 1)
[1] "m"
r$> substr("hour", 1, 1)
[1] "h"
```
and I do not see any unit in the examples or the unit tests for which this standarization is required?

tests/testthat/test-format_conversions.R
Gero1999: @m-kolomanski yes, this this is its main weakness. It will basically assume is `Day` because it starts with "d" and is supposed to be a time unit. I just did not want to overcomplicate it and keep it flexible in case the user uses some way of writing the unit that is not recognized by the `units` package. 

I personally think that the assumptions the function is currently taking will be more than enough for the App case. But let me know your honest opinion, do you think I should perhaps use a more complex regex or similar to invalidate inputs?

m-kolomanski: I understand.

We can keep the function that we thing is _good enough for our case_, especially that with units handling we can spend hours upon hours on developing this function and still not cover all the bases.

The two things that I would do if we want to go this route:
- Do not `@export` this function and allow it only for internal package use - as it will not be a robust enough solution for a end user to use. We know the quirks and can use the function within the package safely, but user who wants to use our package to create some kind of analysis pipeline will not.
- Explain the caveats in the documentation.

R/format_conversions.R
Gero1999: @m-kolomanski The thing is that `min` and `month` have the same start, and the ISO format needs to be able to distinguish them. `min` is well covered by the `units` package, so it will identify it

R/export_cdisc.R
```diff
@@ -126,28 +154,63 @@ export_cdisc <- function(res_nca) {
...
+find_common_prefix <- function(strings) {
+  # Sort the vector of strings
+  first_last_str <- sort(strings)[c(1, length(sorted_strings))]
+
+  # Split the first and last elements into individual characters
+  first_last_chars <- strsplit(sorted_strings, "")
+
+  # Extract the two character vectors
+  first_chars <- first_last_chars[[1]]
+  last_chars <- first_last_chars[[2]]
```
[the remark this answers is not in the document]
Gero1999: @m-kolomanski I have refactored this fun completely and added some docs, let me know if now seems more clear:

```
.find_common_prefix <- function(strings) {
  # Get the strings with the greatest prefix mismatch
  diff_strings <- sort(strings)[c(1, length(strings))] |>
    # Separate the strings by letters
    strsplit("")
  
  # Get the common prefix by using the first mismatch between the two strings
  first_mistmatch <- which(diff_strings[[1]] != diff_strings[[2]])[1]
  substr(strings[1], 0, first_mistmatch - 1)
}
```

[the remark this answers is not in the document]
Gero1999: @m-kolomanski I have refactored this fun completely and added some docs, let me know if now seems more clear:

```
.find_common_prefix <- function(strings) {
  # Get the strings with the greatest prefix mismatch
  diff_strings <- sort(strings)[c(1, length(strings))] %>%
    # For the comparison make all have same number of letters
    sapply(function(x) substr(x, 0, min(nchar(strings)))) %>%
    # Separate the strings by letters
    strsplit("")
  
  # Get the common prefix by using the first letter mismatch between the two strings
  first_mistmatch <- which(diff_strings[[1]] != diff_strings[[2]])[1]
  substr(strings[1], 0, first_mistmatch - 1)
}
```

R/format_conversions.R
m-kolomanski: Ok, I understand. But, that is covered by `if (unit %in% c("h", "min", "s"))` statement. I guess that `units` are useful when the units are either `minute` or `minutes`, because they get coerced to `min`:
```R
r$> convert_to_iso8601_duration(5, "minute")
[1] "PT5M"

r$> convert_to_iso8601_duration(5, "minutes")
[1] "PT5M"
```

but then providing just `m` assumes it is a month:
```R
r$> convert_to_iso8601_duration(5, "M")
[1] "P5M"

r$> convert_to_iso8601_duration(5, "m")
[1] "P5M"
```

Is that expected and correct?

R/export_cdisc.R
```diff
@@ -126,28 +154,63 @@ export_cdisc <- function(res_nca) {
...
+           PARAMCD = PPTESTCD, PARAM = PPTEST, PARAMCAT = PPCAT) %>%
+    select(any_of(c(group_cols, adpp_cols, "RACE", "SEX", "AGE", "AGEU", "AVISIT")))
+
+  # Keep StudyID value to use for file naming
+  studyid <- if ("STUDYID" %in% names(pp_info)) unique(pp_info$STUDYID)[1] else ""
+
+  return(list(pp = pp, adpp = adpp, studyid = studyid))
+}
+
+find_common_prefix <- function(strings) {
```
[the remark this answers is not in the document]
Gero1999: I will create a different branch for this linked with #287

R/format_conversions.R
Gero1999: that is correct and expected for the worst case scenario. In standard circumstances we expect the user to adhere to the regular time denominations (practically 100% times)

m-kolomanski: Ok, all clear, thank you for the explanations!

R/export_cdisc.R
```diff
@@ -93,19 +100,38 @@ export_cdisc <- function(res_nca) {
...
       PPTESTCD = translate_terms(PPTESTCD, mapping_col = "PKNCA", target_col = "PPTESTCD"),
+      PPTEST = translate_terms(PPTESTCD, mapping_col = "PPTESTCD", target_col = "PPTEST"),
       DOMAIN = "PP",
       # Group ID
-      PPGRPID =  paste(ANALYTE, PCSPEC, paste("CYCLE", DOSNO,  sep = " "), sep = "-"),
+      PPGRPID = {
+        if ("PCGRPID" %in% names(.)) PCGRPID
+        else if ("AVISIT" %in% names(.)) paste(ANALYTE, DRUG, PCSPEC, AVISIT, sep = "-")
+        else paste(DRUG, PCSPEC, DOSNO, sep = "-")
+      },
```
Shaakon35, line 110:
```suggestion
      PPGRPID =  dplyr::case_when(
        "AVISIT" %in% names(.) ~ paste0(ANALYTE, PCSPEC, AVISIT, sep = "-")
        "VISIT" %in% names(.) ~ paste0(ANALYTE, PCSPEC, VISIT, sep = "-")
        .default ~ paste(ANALYTE, PCSPEC, DOSNO, sep = "-")
      ),
```
Specs:
For all parameters, create a value comprised of the parameter category [PREPP.PPASTYP], parameter specimen [PREPP.PPASMED] and Visit [PREPP.PPVISIT] (separated by a '-').
For example. RRO4893594-SERUM-DAY 1

no need of DRUG as the NCA is done per metabolite
I guess VISIT could be there in some cases when AVISIT is not available

R/export_cdisc.R
```diff
@@ -93,19 +100,38 @@ export_cdisc <- function(res_nca) {
...
       # Specific ID variables
-      PPSPID = "TBD",
-      # TODO Results in Standard Units if ORRESU is not in standard units
-      PPSTRESN = as.numeric(PPSTRES),
-      PPSTRESC = as.character(PPSTRES),
+      PPSPID = {
+        if ("PCSPID" %in% names(.)) PCSPID
+        else if ("EXSPID" %in% names(.)) EXSPID
+        else NA
+      },
```
Shaakon35, line 121:
```suggestion
      PPSPID = PPSPID = paste0("/F:EDT-", STUDYID, "_PKPARAM_aNCA"),
```
Per requirements of PP,
PPSPID = "Concatenate "/F:EDT-" with the dataset/analysis name [PREPP.PPDSNAME]"

R/export_cdisc.R
```diff
@@ -93,19 +100,38 @@ export_cdisc <- function(res_nca) {
...
       DOMAIN = "PP",
       # Group ID
-      PPGRPID =  paste(ANALYTE, PCSPEC, paste("CYCLE", DOSNO,  sep = " "), sep = "-"),
+      PPGRPID = {
+        if ("PCGRPID" %in% names(.)) PCGRPID
+        else if ("AVISIT" %in% names(.)) paste(ANALYTE, DRUG, PCSPEC, AVISIT, sep = "-")
+        else paste(DRUG, PCSPEC, DOSNO, sep = "-")
+      },
       # Parameter Category
       PPCAT = if ("PARAM" %in% names(.)) PARAM else ANALYTE,
```
Shaakon35, line 112: Would be nice to know which variable we use, because here it's PARAM or ANALYTE (which is the same variable), but in the previous line we just take ANALYTE

R/export_cdisc.R
```diff
@@ -126,28 +154,56 @@ export_cdisc <- function(res_nca) {
...
+        startsWith(PPTESTCD, "AUCINT"),
+        convert_to_iso8601_duration(start, RRLTU),
+        NA
+      ),
+      PPENINT = ifelse(
+        startsWith(PPTESTCD, "AUCINT"),
+        convert_to_iso8601_duration(end, RRLTU),
+        NA
+      ),
+      PPREASND = substr(exclude, 1, 200)
```
Shaakon35, line 167: I think it would be necessary to see what different values of PPREASND we can have to try to make it shorter beforehand, I am unsure how is built this variable

R/export_cdisc.R
```diff
@@ -117,7 +143,9 @@ export_cdisc <- function(res_nca) {
       # Datetime
       PPDTC = Sys.time() %>% format("%Y-%m-%dT%H:%M"),
       PPRFTDTC = {
-        if ("PCRFTDM" %in% names(.)) {
+        if ("PCRFTDTC" %in% names(.)) {
```
Shaakon35, line 146: A check on the format is needed here as well

Shaakon35 (changes requested):

R/export_cdisc.R
```diff
@@ -22,8 +22,17 @@
 #' @export
 export_cdisc <- function(res_nca) {
 
+  # Define group columns in the data
+  group_cols <- unique(
+    unlist(
+      c(res_nca$data$conc$columns$groups,
```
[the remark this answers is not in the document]
Gero1999: the second one corresponds to the dose records columns: res_nca$data$_**dose**_$columns$groups

R/export_cdisc.R
```diff
@@ -48,36 +57,34 @@ export_cdisc <- function(res_nca) {
...
+                 "AAGE",
+                 "AAGEU",
+                 "TRT01P",
+                 "TRT01A",
+                 "AVAL",
+                 "AVALC",
+                 "AVALU")
 
   pp_info <- res_nca$result  %>%
-    filter(is.infinite(end) | PPTESTCD == "auclast") %>%
```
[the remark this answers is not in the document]
Gero1999: it was an old way to exclude the partial AUCs defined by the user. But we don't do nor need this anymore

R/export_cdisc.R
Gero1999: this is a good point, is old code and in the app we use ANALYTE until we change it. Once we change it it will change in everything!

R/export_cdisc.R
```diff
@@ -48,36 +57,34 @@ export_cdisc <- function(res_nca) {
...
+                 "SEX",
+                 "RACE",
+                 "ACTARM",
+                 "AAGE",
+                 "AAGEU",
+                 "TRT01P",
+                 "TRT01A",
+                 "AVAL",
+                 "AVALC",
+                 "AVALU")
```
Shaakon35, line 85:
```suggestion
  adpp_cols <- c("STUDYID",
                 "USUBJID",
                 "PPSEQ",
                 "PPGRPID",
                 "PPSPID",
                 "PARAMCD",
                 "PARAM",
                 "PPCAT",
                 "PPSCAT",
                 "PPREASND",
                 "PPSPEC",
                 "PPDTC",
                 "PPSTINT",
                 "PPENINT",
                 "SUBJID",
                 "SITEID",
                 "SEX",
                 "RACE",
                 "ACTARM",
                 "AAGE",
                 "AAGEU",
                 "TRT01P",
                 "TRT01A",
                 "AVAL",
                 "AVALC",
                 "AVALU")
```

PPCAT needed, 
I removed PARAMCAT 

Somehow when I download the ADPP dataset, I get additional variables that we should remove: 
DRUG ANALYTE std_route PCSPEC

Could you please have a look?

Shaakon35 (changes requested):

R/export_cdisc.R
Gero1999: They are premade text messages, never have seen one greater than 100 characters

R/export_cdisc.R
Shaakon35: @Gero1999 
1. What is important to know in PP or ADPP is one which Analyte the NCA has been run, so we don't need the Drug information, the matching DRUG + ANALYTE is only needed for the derivation of times in the data preparation, then you don't need it at all in the PK parameters. As the specs mention, we need PPASTYP (analyte), PPASMED (ppscpec) and PPVISIT (avisit), that is enough.
2. Do you mean that the ifelse would not crash while the case_when would?
3. Fine

R/export_cdisc.R
Shaakon35: I checked and there is nothing about PCSPID and EXSPID, these IDs are very different from what we want

R/export_cdisc.R
Gero1999: 1. Ok then I will change
2. Yep 
3. On it! 💪🏽

R/export_cdisc.R
Gero1999: Those are the grouping variables for the intervals/parameter-calculations, I personally think it would be better to keep them for general use because internally in the App they make sense for identification

R/export_cdisc.R
Gero1999: @Shaakon35 Won't DTC format be already good? The problem is that to read the date I need to already assume a format. So if it is not DTC when the column is called DTC I won't have to know in which way the date is written. 

An alternative step that I think will be more safe would be to implement a checking in all DTC/DTM columns when the data is mapped. I have included this in #189

Gero1999:
Hey @Shaakon35 @m-kolomanski 

In order to not overcomplicate the issue/pr, I will:

- continue developing a ADPC final in CDISC_export.R in this pr #318
- add tests for final CDISC_export.R functions in this pr #287 

Let me know if there is anything else I need to do apart from those. Hope the pr is now good!

R/export_cdisc.R
Shaakon35: @Gero1999 Sounds good, what we need to keep in mind is that this column needs the correct format so that we can create the xpt (sas) dataset with the datetime format.

m-kolomanski (changes requested):
Hey, I have some more minor issues that I think need ironing out.

R/export_cdisc.R
```diff
@@ -93,19 +100,35 @@ export_cdisc <- function(res_nca) {
...
+      PPTEST = translate_terms(PPTESTCD, mapping_col = "PPTESTCD", target_col = "PPTEST"),
       DOMAIN = "PP",
       # Group ID
-      PPGRPID =  paste(ANALYTE, PCSPEC, paste("CYCLE", DOSNO,  sep = " "), sep = "-"),
+      PPGRPID = {
+        if ("PCGRPID" %in% names(.)) PCGRPID
+        else if ("AVISIT" %in% names(.)) paste(ANALYTE, PCSPEC, AVISIT, sep = "-")
+        else if ("VISIT" %in% names(.)) paste(ANALYTE, PCSPEC, VISIT, sep = "-")
+        else paste(ANALYTE, PCSPEC, DOSNO, sep = "-")
+      },
```
Shaakon35, line 111:
```suggestion
      PPGRPID = {
        if ("AVISIT" %in% names(.)) paste(ANALYTE, PCSPEC, AVISIT, sep = "-")
        else if ("VISIT" %in% names(.)) paste(ANALYTE, PCSPEC, VISIT, sep = "-")
        else paste(ANALYTE, PCSPEC, DOSNO, sep = "-")
      },
```
PCGRPID not needed, it really not the same format

Shaakon35 (changes requested):

R/export_cdisc.R
Shaakon35: For sure, we need PPCAT = PARAMCAT (PARAMCAT is not CDISC standard)

@Gero1999 If you look, we already have them:
We already have ANALYTE, which is PPCAT
We already have PCSPEC, which is PPSPEC (We should not have PC variables in PP or ADPP)
As mentioned in https://github.com/pharmaverse/aNCA/pull/306#discussion_r2026351421, DRUG is not useful for the PP and ADPP
std_route is not CDISC standard and anyways, it is not useful to have DRUG information in PP and ADPP

**That's why I would remove PARAMCAT, DRUG, ANALYTE, std_route, PCSPEC and add PPCAT**

Gero1999:
Now all should hopefully be solved and we have proper PP/ADPP, thanks for all feedback and suggestions! @m-kolomanski @Shaakon35  💪🏽

R/export_cdisc.R
Shaakon35: @Gero1999

m-kolomanski (approved):
LGTM, thank you for all the work!

Shaakon35 (approved):

PR #294, opened 2025-03-27
Bug: nca widget run every change

js3110:
## Issue

Closes #275 

## Description
Loading widget was loading every time user updates setup options. This was due to the NCA being rerun every time, in order to update the slopes.

To make the process more organised and clean, I have split the updating of PKNCAdata object into two: one for slopes and one for the main NCA. For this, I created the function `PKNCA_update_data_object`
The slopes data is that used to run the NCA for the slope plots, and the other data is used for the main NCA. They are set up in the same way to ensure consisitency between slope plots and results, but are separated to prevent delays in data loading and lags in the interval setup.

## Definition of Done

- [x] Calculations of lambda.z in plots are less frequent
- [x] Loading widget does not appear unless Run NCA is clicked

## How to test

Load data and play around with adjusting settings, viewing slopes and intervals etc.

## Contributor checklist
- [x] Code passes lintr checks
- [x] Code passes all unit tests
- [x] New logic covered by unit tests
- [x] New logic is documented
- [x] Package version is incremented

## Notes to reviewer

Things that I think still need modifications, just maybe in separate issues:
- parameter input selection needs to change, as with pickerInput the box closes after every modification, so you can't select multiple in one go
- intervals table could be changed so it displays more simple information (not sure that parameters are needed in the table)

PR #309, opened 2025-04-01
Refactor: remove `pk_nca_trigger` reactive, rely on the button value …

### LINKED ISSUE #210: Enhancement: remove apply button in dynamic slope plots

m-kolomanski:
## Issue

Closes #210 

## Description

- Removes **Apply** button near slope selector and `pk_nca_trigger` reactive connected to that button. The **Apply** button is redundant, as it basically just triggers the NCA analysis, just as **Run NCA** button.
- Adds check to see if REASON was provided for all selection/exclusions provided by the user.
- Reworks the error handling bit slightly to nicely display errors coming from NCA calculations.

## Definition of Done

- [x] No **Apply** button, things are saved instantly (this is with changes in #294)
- [x] Add a check to make sure text REASONS have been included for `Inclusions` when running `NCA`

## How to test

Run the app, map the data, go to NCA -> Slope Selector. Add a selection/exclusion. Change the new row to TYPE == "Exclusion" - it will be missing REASON as default. Run the NCA analysis. It should throw an error. If you provide any string in the REASON field of the table, NCA should go fine. 

## Contributor checklist
- [x] New logic is documented


Gero1999 (changes requested):
Thank you for the changes! Only some small nitpicks:

- [x] Consider graphical placeholder as if no REASON was provided. Feel free also to simply eliminate the placeholder as a solution to this. I think if we have the notification it is not that much of a need.
- [ ] question for @js3110: If I am correct only `TYPE = Exclusions` are mandatory to have a REASON but not `TYPE = Selection` , am I right? I think this is my fault, because I wrote it wrong in the original issue (I corrected it now) #210
- [ ] Include in the notification the section/location that require the changes `Setup > Slope Selector`. I think if users run NCA in another section and find the error can become confused.

m-kolomanski:
Hey @Gero1999,
>  Consider graphical placeholder as if no REASON was provided. Feel free also to simply eliminate the placeholder as a solution to this. I think if we have the notification it is not that much of a need.

Could you elaborate? Do you mean like a placeholder in the text widget?

> Include in the notification the section/location that require the changes Setup > Slope Selector. I think if users run NCA in another section and find the error can become confused.

What would be the expected UX here? When this specific error is detected, should the tab change to Slope Selector? Or should we change the error message to be more like `Please provide the reason for all exclusions in the Slope Selector tab`? I am expecting that the user is either in **NCA settings** or **Slope Selector** when they run the NCA, as other tabs do not include any information before that is done.

js3110:
> * [ ]  question for @js3110: If I am correct only `TYPE = Exclusions` are mandatory to have a REASON but not `TYPE = Selection` , am I right?

Yes, I think currently only exclusions are mandatory, but its still preferred if they give a reason for selection as well. In the future, we want to go towards having a dropdown box with options to make it more standardised, but still waiting for feedback on what those options should be!

Gero1999:
> Hey @Gero1999,
> >  Consider graphical placeholder as if no REASON was provided. Feel free also to simply eliminate the placeholder as a solution to this. I think if we have the notification it is not that much of a need.
> 
> Could you elaborate? Do you mean like a placeholder in the text widget?

@m-kolomanski Currently when the user makes a selection/exclusion graphically via the plots there is a default text value in REASON. Therefore, if you run NCA right after it there won't be a notification, even if the user per ce did not introduce any real reason

m-kolomanski:
@Gero1999 I have implemented the feedback, PTAL!

Gero1999 (approved):
oleee looks amazing to me, thanks for the changes! :)

R/utils-slope_selector.R
```diff
@@ -8,25 +8,41 @@
...
+
+    if (any(exclusions$REASON == "")) {
+      stop(
+        "No reason provided for the following exclusions:\n",
+        "PCSPEC USUBJID ANALYTE DOSNO RANGE\n",
+        filter(exclusions, REASON == "") %>%
+          select(PCSPEC, USUBJID, ANALYTE, DOSNO, RANGE) %>%
+          apply(1, \(x) paste0(x, collapse = " "))
+      )
+    }
```
rabiibouhestine, line 38: Could you extract the code 

```
        filter(exclusions, REASON == "") %>%
          select(PCSPEC, USUBJID, ANALYTE, DOSNO, RANGE) %>%
          apply(1, \(x) paste0(x, collapse = " "))
```

outside the `stop()` for better readability please?

rabiibouhestine (changes requested):
One small request, otherwise looks good!

R/utils-slope_selector.R
m-kolomanski: @rabiibouhestine Thanks for the feedback - done, PTAL!

rabiibouhestine (approved):
LGTM

PR #310, opened 2025-04-02
fix: fixed results table not updating upon flag change and re-run

### LINKED ISSUE #260: Bug:  NCA Results table not refreshing upon re-running the NCA analysis with changed flag parameters
### Description
When re-running the NCA analysis with different flags, the actual results table stays the same and does not update. Might concern more modules than just the results table.

### To reproduce
1. Run the app, map the data, run NCA analysis with default settings and inspect the table.
2. Add flag rule like setting RSQADJ to 0.99. in settings
3. Run the NCA analysis again
4. The table will not update

### Expected behaviour
The results should update every time the user re-runs the analysis during a single session with different options.

m-kolomanski:
## Issue

Closes #260 

## Description

When re-running the NCA analysis with different flags, the actual results table stays the same and do not update.

## Definition of Done

- results table updates when the analysis is re-run with different settings

#### main branch
1. Run the app, map the data, run NCA analysis with default settings and inspect the table in NCA results
2. Go back to Setup -> NCA settings -> Flag Rule Sets and check RSQADJ. Set it to 0.99.
3. Run the NCA analysis again -> The results table will not update

You can also run the app again and set RSQADJ = 0.99 on the first pass - the readouts that are below the thresholds will be flagged and the appropriate rows will be red.

#### this branch
Go through the same workflow. On step 3 and running the analysis again with different settings, the table should update and include rows marked red.

## Contributor checklist


inst/shiny/modules/tab_nca/nca_results.R
```diff
@@ -57,6 +57,7 @@ nca_results_server <- function(id, pknca_data, res_nca, rules, grouping_vars, au
       final_results <- pivot_wider_pknca_results(results)
 
       # Apply flag rules
+      rules <- rules()
```
rabiibouhestine, line 60: could we rename `rules` to `current_rules` here to avoid confusion ? 

`current_rules <- rules()`

rabiibouhestine (changes requested):
nice catch, I have one tiny change request :)

inst/shiny/modules/tab_nca/nca_results.R
m-kolomanski: Yup, good point. PTAL!

rabiibouhestine (approved):
LGTM

PR #314, opened 2025-04-02
Bug: AUCint not working for end = Inf

### LINKED ISSUE #281: Bug: AUCint doesn't work with basic settings
### Description
If the user selects AUCint as one of the parameters, for the main data (not for partial AUC), they are given an error saying `Error with interval start=0, end=Inf: the condition has length > 1`

This is happening because in `format_pknca_intervals()` there is a line of code `mutate(end = lead(as.numeric(time_dose), default = Inf))` which sets the end time to Inf if there are no subsequent doses. 

### To reproduce
Select aucint.inf.pred as a parameter.

### Expected behaviour
A separate interval will need to be created when aucint is selected, in the same way that is done with partial aucs.
The interval should have start = start of dose, and end = end of dose/start of next dose. 
This interval only needs to be created for the doses that have Inf in the end, as the others will already be fine. Then jsut ensure that aucint is TRUE for these intervals.

js3110:
## Issue

Closes #281 

## Description

AUCIntpred was causing the app to crash if selected, because the end = Inf and therefore there was no interval for calculation.
I have set to remove AUCINTIFP from parameters where end = Inf.

When TAU is included in the app this will create a separate interval for AUCTAU calculations, but this code should stay the same anyway, as either way calculating AUCINT for Inf is not possible
## Definition of Done

- [x] AUCInt can be calculated for all intervals except last

## How to test

Select AUCINTIFP for parameters and run NCA

## Contributor checklist
- [x] New logic is documented


Shaakon35 (approved):

m-kolomanski (approved):
LGTM

PR #319, opened 2025-04-03
308 refactor noisy warnings pknca

### LINKED ISSUE #308: Refactor: noisy warnings from `pk.nca()`
### Description
Every time `PKNCA::pk.nca()` function is run within the application, we get quite noisy warnings in the console:

```R
Warning: STUDYID=XX01; PCSPEC=CSF; DRUG=Analyte01; USUBJID=11101; ANALYTE=Analyte01: No intervals for data
Warning: STUDYID=XX01; PCSPEC=CSF; DRUG=Analyte01; USUBJID=11101; ANALYTE=Metab. of Analyte01: No intervals for data
Warning: STUDYID=XX01; PCSPEC=SERUM; DRUG=Analyte01; USUBJID=11101; ANALYTE=Metab. of Analyte01: No intervals for data
```

would be good to handle those warnings appropriately - should it be something to be worried about? Maybe we should silence the PKNCA warnings and log our own in more concise manner?

### To reproduce
Run the application, map the data. This will trigger first wave of warnings. Then every time the NCA is run, more warnings will appear.

### Expected behaviour
No warnings in the console.

Gero1999:
## Issue

Closes #308 

## Description
Every time PKNCA::pk.nca() function is run within the application, we get quite noisy warnings in the console:
```
Warning: STUDYID=XX01; PCSPEC=CSF; DRUG=Analyte01; USUBJID=11101; ANALYTE=Analyte01: No intervals for data
Warning: Too few points for half-life calculation (min.hl.points=3 with only 2 points)
```
would be good to handle those warnings appropriately - should it be something to be worried about? Maybe we should silence the PKNCA warnings and log our own in more concise manner?

## Definition of Done
- [x] Warnings from pk.nca and/or PKNCA_calculate_nca() that are not troublesome/relevant should be gone for good!

## How to test
Run the application, map the data. This will trigger first wave of warnings. Then every time the NCA is run, more warnings will appear. 

## Contributor checklist
- [x] New logic is documented


js3110 (approved):
Looks good to me ! thanks!

Just a small suggestion, since its on the topic of warning messages, could you also change the `interval_remove_impute.data.frame` warning, as its technically not a warning, just a message. `No impute method specified. No changes made.`

either remove it or change the type to maybe a log trace or something!

R/PKNCA.R
```diff
@@ -314,13 +317,38 @@ PKNCA_update_data_object <- function( # nolint: object_name_linter
...
+        }
+      }
+    )
+  } else if (warning_level == "none") {
+    results <- suppressWarnings(PKNCA::pk.nca(data = pknca_data, verbose = FALSE))
+  } else {
+    stop(paste0("Invalid `warning_level` = ",
+                warning_level,
+                ". Options: 'all', 'partial', 'none'"))
+  }
```
m-kolomanski, line 341: discussion: I am not a fan of this solution.

For one, if we encounter unknown errors, they will not be logged nicely:
```R
    results <- withCallingHandlers({
      warning("blah")
      PKNCA::pk.nca(data = pknca_data, verbose = FALSE)
    },
    ...
```
Screenshot from 2025-04-04 07-43-34

Secondly, do we expect to use this function in other contexts with different warning levels? Are we gonna be calling this function in other parts of the application? If not, I believe the argument for stating warning levels is a bit overengineered, as it will be used only once and with the same value, changing it will require code changes. It would make more sense if the user was able to specify the warning level from within the app - which I do not think is in the plans?

Either way, I think it would make more sense if the application handled the warnings and the logic function was allowed to throw whatever it wants to throw. I imagine the app as a kind of _guided experience_ through the analysis, but advanced user should be able to create their own pipeline using functions available in the package. Therefore, the application could handle and muffle some common warnings for _curated workflow_, but the function when used standalone should throw all warnings and we should let the user handle them themselves.

Let me know if my assumptions and thoughts are in line with your expectations regarding the app and the package. There is nothing inherently wrong with the code, it just does not suit my vision of how the package/app should work, but that might be misaligned with your plans.

R/PKNCA.R
```diff
@@ -283,6 +285,7 @@ PKNCA_update_data_object <- function( # nolint: object_name_linter
 #' with the start and end times for each dose, from first and most recent dose.
 #'
 #' @param pknca_data Data object created using PKNCA::PKNCAdata() function.
+#' @param warning_level Character string with three possible values
```
m-kolomanski, line 288: issue: For a param like that we would need more documentation - what are the levels? What do they mean, what kind of warnings they allow/muffle?

R/PKNCA.R
Gero1999: hey @m-kolomanski. My though process on this was that if we did it over the entire `PKNCA_calculate_nca()` call, then we could potentially miss other internal non `pk.nca` warnings that could be relevant.  

Therefore I included the fun argument thinking in the possibility to investigate differently during debugging slope_selector / tab_nca sections. But I have no problem in changing the solution to something different, as my vision is not very technical. Just to make sure I understand, your preferred solution would be to silence these warnings without affecting internally the function right?

m-kolomanski (changes requested):
issue: Running `grepl` agains a warning object `w` will never return `TRUE`, you need to check the actual message of the warning object:
```R
if (any(grepl(paste(irrelevant_regex_warnings, collapse = "|"), conditionMessage(w)))) {
```

question: With the current setup, the application will print the listed _irrelevant_ warnings, while omitting any other. Is that the desired behavior? I would have thought we want to silence the _irrelevant_ warnings and print any others - to do so, we need to reverse the result of the `grepl`:
```R
if (!any(grepl(paste(irrelevant_regex_warnings, collapse = "|"), conditionMessage(w)))) {
```

question: Do we just want to log the warning to the console? Would there be any benefit to `showNotification` and informing the user of the warning directly in UI?
```R
if (any(grepl(paste(irrelevant_regex_warnings, collapse = "|"), w))) {
  log_warn(conditionMessage(w))
  showNotification(conditionMessage(w), type = "warning")
}
```

discussion: Also, now I am thinking, there is also an option that we will not silence the known warnings completely, but instead log them as **debug** events. But I do not know if those warnings have any debugging benefit? If they are completely dismissible, then we might as well silence them completely. But if they might be useful in debugging any specific bugs or edge-cases, then we could log them as such:
```R
if (any(grepl(paste(irrelevant_regex_warnings, collapse = "|"), w))) {
  log_warn(conditionMessage(w))
} else {
  log_debug(conditionMessage(w))
}

### OR
log_level(
  if (any(grepl(paste(irrelevant_regex_warnings, collapse = "|"), w)) "DEBUG" else "WARN"
  conditionMessage(w)
)


```
Ofc the point that the app user should not see those logs, since they are accounted for - I have created issue #324 to add ability to specify log levels, so that users are not pummeled with non-informative logs, but developers still can have access to them. Let me know what you think on that regard - this is very loose suggestion and I do not know if those logs would be of any use at all.

Gero1999:
hey @m-kolomanski I agree with your points. In summary:

1,2,3) I will correct the warning function to **not** consider the mentioned warnings and also to show a notification for the "important" ones
```
          warning = function(w) {
            if (!grepl(paste(irrelevant_regex_warnings, collapse = "|"),
                       conditionMessage(w))) {
              log_warn(conditionMessage(w))
              showNotification(conditionMessage(w), type = "warning", duration = 5)
            }
            invokeRestart("muffleWarning")
          })
```
4) Regarding the discussion, I personally think could makes things cleaner for developing. This is definitely more your call. But I am also thinking that maybe making the logs available in the App could be a worthy discussion so users can report to us precisely what was their problem if the find issues. What is your opinion @Shaakon35 ?

m-kolomanski:
FYI, we can also do some magic and set up the logger in such a way that only INFO+ will be logged to console, but we will also save all the logs in a log file - so then, when a bug is found, we can ask a user to sent over their log file and have the access to all the traces and debugging info from the app.

m-kolomanski (approved):
See my other comments and discussion, but in the current form for me the PR is good. If we wish to expand the feature in this PR, please request re-review. Otherwise, I think it is fine to accept the PR as-is and implement any additional features in the scope of other issues.

PR #321, opened 2025-04-03
Enhancement: update setup widgets

### LINKED ISSUE #296: Enhancement: update setup widgets inc parameter selection
### Description
Parameter selection currently only allows one parameter to be added at a time, and is difficult to see all the parameters clearly.
In addition, the code for the rule sets could be refactored into a function.

It would be good to refactor the parameter selection, and maybe create a different type of widget. for example a table with checkboxes, or a selectInput, where all the parameters can be seen better and selected without having to stop each time.

### Definition of Done
- [ ] Parameter selection easy to read and can select multiple in one go
- [ ] code for widgets is clean and minimal repeats

js3110:
## Issue

Closes #296 

## Description

Parameter selection widget was previously pickerInput, which made for difficultly selecting multiple at a time. It was also hard to properly see everything that was selected. Here, I have refactored it to a reactable instead. This way, many can be selected at once, and its easier to see clearly what the parameters are.

## Definition of Done

- [x] Parameter selection improved
- [x] Code optimised for widgets

## How to test
View parameter selection and setup widgets:
`Column Mapping: Submit mapping` > `NCA` > `Setup` > `NCA settings` > `NCA parameters to calculate`: 

## Contributor checklist
- [x] New logic is documented


js3110:
@Gero1999 @lucyaspridis @m-kolomanski 
I would appreciate some input on this PR before I finish it off, whether you think this layout is the best option, and code wise whether it could be improved. Let me know! thanks :)

Gero1999:
hey a very nice input on this might also be @rabiibouhestine

js3110:
> * Reduce the number of parameters selectable by the user. There are indeed many they don't need (like all with the `.iv` suffix)

What do we do if they have an IV study and want these parameters?

> * Subdivide the input choices using list elements

with this I have tried to make it a bit more organised, by subdividing into categories, but I am open to someone giving me a better solution!! 

I agree, I don't know that this is the best option. The problem with pickerInput is that you can only select one at a time, and with selectInput the widget gets very full and takes up too much of the screen, and doesn't look too great. I already tried selectInput while i was making this, but it was even worse than the pickerInput...

m-kolomanski:
I actually get an error when trying to run the application (this is after mapping the data, when I switch to NCA tab):
```R
Warning: Error in filter: ℹ In argument: `grepl("auc|aumc", PKNCA) | grepl("auc", Depends)`.
Caused by error:
! object 'Depends' not found
```

so seems like this is due to the `PKNCA_update_data_object` function, particularly this part of the code:
```R
    params_auc_dep <- pknca_cdisc_terms %>%
      filter(grepl("auc|aumc", PKNCA) | grepl("auc", Depends)) %>%
      pull(PKNCA)
```

`pknca_cdisc_terms` does not have Depends column after your commit, which breaks this function.

js3110:
Yes, I realised that I updated the cdisc_terms but with an outdated version of the file, and I am waiting for @Gero1999 to add the Depends column to our csv copy, so that I can then update it with all the columns :)
Not really sure why it worked before but then stopped working, as I don't recall merging with main....

js3110:
hi @m-kolomanski it should be working now!

m-kolomanski (commented):
For me, the UI looks awesome. I would probably go for something more custom and nicer than `verbatimTextOutput`. I would suggest creating a container with `display: grid` and several columns, then in the observer for each param creating a `p` or `span` element and rendering those in the container - this would give us similar layout, but without the `print` aesthetic. This could be wrapped in a `card` element to give it a nice bit of separation to the table.

inst/shiny/modules/tab_nca/nca_setup.R
```diff
@@ -28,43 +61,35 @@ nca_setup_ui <- function(id) {
...
-              choices = c("f_aucinf.obs", "f_aucinf.pred", "f_auclast"),
-              multiple = TRUE,
-              selected = NULL
+        ),
+        accordion_panel(
+          title = "Parameter Selection",
+          reactableOutput(ns("nca_parameters")),
+          card(
+            full_screen = FALSE,
+            card_header("Selected NCA Parameters"),
```
m-kolomanski, line 90: nitpick: For me, if all options in the table are open, the table seems a little bit too close to the card with pills - I would consider adding a small margin for the spacing to look a bit nicer:
```suggestion
            full_screen = FALSE,
            style = "margin-top: 2em;",
            card_header("Selected NCA Parameters"),
```

m-kolomanski (commented):
Hey, the code looks awesome, I have just some minor stylistic nitpicks.

I would accept it straight away, but I have noticed the PR adds some empty files in the `vignettes/` folder - I do not see any reason for that?

m-kolomanski (approved):
LGTM, thank you!

BTW, I have noticed changes to the download settings. As of now the download settings do not work, even on `main`, so the changes do not matter. I thought we would have an issue on the board for this - but I do not see any. Should we add it? I could start working on this bug.

js3110:
> BTW, I have noticed changes to the download settings. As of now the download settings do not work, even on `main`, so the changes do not matter. I thought we would have an issue on the board for this - but I do not see any. Should we add it? I could start working on this bug.

I believe they are being worked on in #274 
I made the changes here so that Gerardo will see the conflicts and change them in his PR as well

Gero1999 (approved):
still personally think the UX seems overengineered but seems to work fine so I will just approve! 😅

js3110:
@Gero1999 i agree lol, hopefully the UX guys will come up with something better! the important thing is that it works and allows multiple selection at once lol

Issue #291, opened 2025-03-26
Enhancement: PP/ADPP/ADPC in other formats


Gero1999:
### Description
If sas7bdat format file is not possible, we will try to add another alternative file formats so SAS users can also enjoy reading the app outputs. So for now our best considerations

### Definition of Done
PP/ADPP files should be exportable in:
- [ ] xpt or packet (if not sas7bdat) --> including all proper class formats of the file
- [ ] csv --> already present
- [ ] rds or rda

Shaakon35:
@Gero1999 Please re-use the code I created to generate PP and make the checks we need:
/_Projects/Training_5152/CPDS/NCA/aNCA/Pillar3/aNCA_Pillar3_and_PP/Step 5/PP_aNCA.R

This issue is blocked by https://github.com/pharmaverse/aNCA/issues/277

PR #329, opened 2025-04-08
Feat: final ADPC in export_cdisc for Parameter Datasets #318

### LINKED ISSUE #318: Enhancement: export final ADPC with CDISC standards
### Description
The final ADAM ADPC file can be generated through the App in the same way as PP/ADPP. In fact, it would make sense the three of them are together.

### Definition of Done
- [ ] ADPC with CDISC standards (including the derived ANL01FL)
- [ ] option to download is not relevant, as it will be later continued in #291

Gero1999:
## Issue

Closes #318

## Description
The final ADAM ADPC file can be generated through the App in the same way as PP/ADPP. In fact, it would make sense the three of them are together.

## Definition of Done

## How to test
- [x] ADPC with CDISC standards (including the derived ANL01FL)

## Contributor checklist
- [x] New logic is documented


js3110 (changes requested):
Hey, thanks! Looks really good. One small note:
in the ADPC there are extra columns that shouldn't be there:
`conc_groups`, `is.included.hl` and `REASON`

Gero1999:
> REASON

You are right @js3110 thanks! However, won't you agree that it will make a lot of sense to have `REASON` or a similar column to inform on why a point was excluded (?). I just did not found anything on CDISC/Roche but for me this seems like a big mistake

js3110:
> You are right @js3110 thanks! However, won't you agree that it will make a lot of sense to have `REASON` or a similar column to inform on why a point was excluded (?). I just did not found anything on CDISC/Roche but for me this seems like a big mistake

I would say that this would be better to have in ADPP, not ADPC, since its related to parameters not to the PC?
But I agree it should be included somewhere! @Gero1999

Gero1999:
> I would say that this would be better to have in ADPP, not ADPC, since its related to parameters not to the PC? But I agree it should be included somewhere! @Gero1999

The problem is that ADPP has one row per parameter not per sample, so there is no way to say why the X specific record was not considered

Gero1999:
@js3110  For simplification I will just remove them now, and perhaps re-add them in #291 if required by them

m-kolomanski (approved):
Nice, the code looks good. I have left some minor suggestions that are a bit subjective, I am leaving it up to you if you feel like implementing them.

R/export_cdisc.R
```diff
@@ -206,3 +168,132 @@ find_common_prefix <- function(strings) {
...
+    "PPDTC",
+    "PPSTINT",
+    "PPENINT",
+    "SUBJID",
+    "SITEID",
+    "SEX",
+    "RACE",
+    "ACTARM",
+    "AAGE",
+    "AAGEU",
```
js3110, line 262:
```suggestion
    "AGE",
    "AGEU",
```

spelling mistake? or is it meant to be like that

R/export_cdisc.R
```diff
@@ -168,25 +107,48 @@ export_cdisc <- function(res_nca) {
...
+  pp <- cdisc_info %>%
+    select(all_of(CDISC_COLS$PP))
 
-  adpp <- pp_info %>%
+  adpp <- cdisc_info %>%
     # Rename/mutate variables from PP
     mutate(AVAL = PPSTRESN, AVALC = PPSTRESC, AVALU = PPSTRESU,
            PARAMCD = PPTESTCD, PARAM = PPTEST) %>%
-    select(any_of(c(adpp_cols, "RACE", "SEX", "AGE", "AGEU", "AVISIT")))
+    select(any_of(c(CDISC_COLS$ADPP, "RACE", "SEX", "AGE", "AGEU", "AVISIT")))
```
js3110, line 117: SEX, RACE, AGE, AGEU are already in the $ADPP columns list, so don't need to be here too.
Can AVISIT also be  added to the  list instead of manually here?

Question: what happens if the columns are not in the dataset?

js3110 (approved):
LGTM!!

PR #336, opened 2025-04-09
feat: inlcude `AUCINT` inselected parameters if any AUCs are provided

### LINKED ISSUE #326: Bug: partial AUCs table problems with selectInput in results & not saving unless Add Row
### Description
When a partial AUC is selected by the user, it is expected that the results contain automatically the AUCINT parameter (PPTESTCD) in the selectInput of the `NCA results`

### To reproduce
`Map columns` > `NCA` > `NCA setup` > Select partial AUCs i.e, `0-100` > `NCA results`
AUCINT is not in `NCA results > Select Parameters` there unless you previously indicated it in `NCA settings > Select Parameters`

### Expected behaviour
AUCINT should automatically be present in there and forced when the user asks for it implicitely in `Select partial AUCs`.

m-kolomanski:
## Issue

Closes #326 

## Description
When a partial AUC is selected by the user, it is expected that the results contain the AUCINT parameter (PPTESTCD) in the selectInput of the NCA results. This was caused due to mismatch of column names and custom labels - normally, the column name for partial AUC interval would be `AUCINT`, however in our implementation the limits of the interval are specified in column name (e.g. `AUCINT_5-50`). This lead to issues when assigning label attributes and later parameters selection.

## Definition of Done

- [x] AUCINT should automatically be present in there and forced when the user asks for it implicitely in Select partial AUCs.

## How to test
Map columns > NCA > NCA setup > Select partial AUCs, e.g. provide 5 and 55, run the NCA analysis.

#### On main branch
**AUCINT** will not be present in Results -> NCA Results -> Select parameters widget.

#### On this branch
**AUCINT** will be present in the results parameter selection. In addition, the results table should include separate columns for each provided interval.

## Contributor checklist


rabiibouhestine (approved):
LGTM

Gero1999 (changes requested):
Hey, I only see 2 problems with the proposed solution:

- [ ] If AUC partial is selected, then AUCINT is calculated for all intervals, not only the partial ones selected

- [ ] the condition `if (any(apply(auc_data(), 1, \(x) !all(is.na(x)))))`  will still apply even if the user did not populate the entire row. Should be perhaps `all(!is.na()`

I investigated more, and I think the problem is more related on how `auc_data()` is updated in the `PKNCA_update_data_object()` call in `nca_setup.R`. Unless the user presses "Add row" after adding the AUCpartial details in the UI, the object is not updated there and therefore the intervals are not added. Personally, I think the preferred solution would be on the direction of correcting that, but if we manage to solve the problems with this that can also be ok!

js3110:
> I investigated more, and I think the problem is more related on how `auc_data()` is updated in the `PKNCA_update_data_object()` call in `nca_setup.R`. Unless the user presses "Add row" after adding the AUCpartial details in the UI, the object is not updated there and therefore the intervals are not added. Personally, I think the preferred solution would be on the direction of correcting that, but if we manage to solve the problems with this that can also be ok!

hey! are you sure about this? when I try in main the partial aucs are added if you add them in the table, no requirement for Add Row. All add row does is add a new row to the table.

js3110:
@Gero1999 maybe you could upload a video/image that shows the partial aucs not updating without add row? because I am unable to recreate

js3110:
> Here is a screen recording of the problem I saw:

Thanks @Gero1999 ! The problem here is that you click Run NCA too soon- something I created issue #341 for!
The settings are set to a 2.5 second debounce to allow for the user to make multiple changes at once, so if you run the nca before 2.5 seconds, the changes are not applied. Therefore this is a separate issue :)

m-kolomanski:
@js3110 @Gero1999 Hey, sorry for missing the point of the issue. Can we reiterate what is the root of the problem and the expected solution? I am unsure at the moment. Thank you!

Gero1999:
hey @m-kolomanski , the user problem with the solution will be this:

> * [ ]  If AUC partial is selected, then AUCINT is calculated for all intervals, not only the partial ones selected
> * [ ]  the condition `if (any(apply(auc_data(), 1, \(x) !all(is.na(x)))))`  will still apply even if the user did not populate the entire row. Should be perhaps `all(!is.na()`


The thing is that partial AUCs selected by the user are already populated with `aucint` for `pk.nca` and included as independent rows. They are also renamed in the results from `aucint.last` to `AUCINT` + `start-end`, so it is not in the `pknca_cdisc_terms$PPTESTCD` object. So the likely easiest solution would be change this part in the code:

```
observeEvent(final_results(), {
      req(final_results())

      param_pptest_cols <- intersect(unname(var_labels(final_results())), pknca_cdisc_terms$PPTEST)
```

Hope that makes it more clear, let me know if not!

m-kolomanski:
@Gero1999 @js3110 I think I get it now. So the issue is with the labels - we are selecting columns based on `var_labels`. I can see now that during data processing we create column **AUC_X-Y,** but this is not a valid term, so it does not get labeled as **AUCINT[unit]**. If we assigned the label correctly, everything should fall in place so I would suggest modifying `.add_label_attribute` so that the label gets assigned to the interval columns - does that sound coherent?

BTW, thank you for taking the time to explain this issue, I with to understand the data processing of the application better!

js3110:
hi @m-kolomanski , that sounds good to me! 
The PKNCA is aucint.last and the CDISC is AUCINT
Then for user clarity it needs to be labelled somehow as start-end (so they know the interval)
But if you see a better way/different way to do it that makes it easier to ensure that the results and parameter selections work, then feel free to change!

m-kolomanski:
@js3110 @Gero1999 I have implemented the solution as discussed.
@rabiibouhestine The solution has changed completely since your review, so please kindly take another look!

js3110 (approved):
LGTM! Thanks :)

R/pivot_wider_pknca_results.R
```diff
@@ -150,6 +151,29 @@ pivot_wider_pknca_results <- function(myres) {
     distinct() %>%
     pull(PPTESTCD_cdisc, PPTESTCD_unit)
 
+  # if custom AUC intervals are present, add them to the mapping vector
+  aucint_index <- grep("AUCINT", names(mapping_vr))
+  if (!rlang::is_empty(aucint_index)) {
```
Gero1999, line 156: issue: AUCINT requested by the user via `nca_params` will also be affected (replicating the units in the column). Perhaps this can be an alternative solution, feel free to use another if you don't like it!

```
add_label_attribute <- function(df, myres) {

  mapping_vr <- myres$result %>%
    mutate(
      PPTESTCD_unit = case_when(
        type_interval == "manual" ~ paste0(PPTESTCD, "_", start, "-", end),
        PPSTRESU != "" ~ paste0(PPTESTCD, "[", PPSTRESU, "]"),
        TRUE ~ PPTESTCD
      ),
      PPTEST_cdisc = translate_terms(PPTESTCD,
                                     mapping_col = "PPTESTCD",
                                     target_col = "PPTEST")
      ) %>%
    select(PPTEST_cdisc, PPTESTCD_unit) %>%
    distinct() %>%
    pull(PPTEST_cdisc, PPTESTCD_unit)

  mapping_cols <- intersect(names(df), names(mapping_vr))
  attrs <- unname(mapping_vr[mapping_cols])

  df[, mapping_cols] <- as.data.frame(mapply(function(col, bw) {
    attr(col, "label") <- bw
    col
  }, df[, mapping_cols], attrs, SIMPLIFY = FALSE))
  df
}
```

Gero1999 (changes requested):
only one small comment, but is almost perfect!

m-kolomanski:
@Gero1999 Your approach makes much more sense, thank you for pointing that out! I have found something slightly different, please let me know what you think!

R/pivot_wider_pknca_results.R
```diff
@@ -137,15 +145,22 @@ pivot_wider_pknca_results <- function(myres) {
...
+          ifelse(PPSTRESU != "", paste0("[", PPSTRESU, "]"), "")
+        ),
+        PPSTRESU != "" ~ paste0(PPTESTCD, "[", PPSTRESU, "]"),
+        TRUE ~ PPTESTCD
+      ),
+      PPTESTCD_cdisc = translate_terms(PPTESTCD, mapping_col = "PPTESTCD", target_col = "PPTEST")
+    ) %>%
     select(PPTESTCD_cdisc, PPTESTCD_unit) %>%
     distinct() %>%
     pull(PPTESTCD_cdisc, PPTESTCD_unit)
```
Gero1999, line 166: suggestion: Just a refactoring to
1) simplify the nested/implicit `ifelse`s in the `case_when`
2) to rename variable `PPTESTCD_cdisc` > `PPTEST_cdisc`

```suggestion
    mutate(
      PPSTRESU = ifelse(
        PPSTRESU != "",
        paste0("[", PPSTRESU, "]"),
        ""
      ),
      PPTESTCD = ifelse(
        type_interval == "manual",
        paste0(PPTESTCD, "_", start, "-", end),
        PPTESTCD
      ),
      PPTESTCD_unit = paste0(PPTESTCD, PPSTRESU),
      PPTEST_cdisc = translate_terms(PPTESTCD, mapping_col = "PPTESTCD", target_col = "PPTEST")
    ) %>%
    select(PPTEST_cdisc, PPTESTCD_unit) %>%
    distinct() %>%
    pull(PPTEST_cdisc, PPTESTCD_unit)
```

Gero1999 (approved):
thanks for the changes, it actually makes sense to add the unit! I just made a non-blocking suggestion, but feel free to ignore it!

PR #339, opened 2025-04-11
Refactor: `tab_tlg.R` and `tlg_module.R` should not use `session$userdata` object

### LINKED ISSUE #265: Refactor: `tab_tlg.R` and `tlg_module.R` so it does not use `session$userData` object
### Description
This would be a step towards simplifying the data flow in the application.

Currently, the tlg-related modules use global **session$userData$data** object to fetch the data for display.  This is in order to avoid verbosely passing data objects down the line, however it might lead to confusing data flow where it is not clear where the data comes from and what kind of data the module works with. When the application is refactored and modularized, and with workflow improvements this is no longer needed and passing the correct data object down the stream would be more readable and easier to maintain.

@js3110 @Gero1999 I would ask for your opinion here too, but it is kinda irrelevant - this issue is for some _technical debt_ that I have willingly taken some time ago and with all the improvements that happened (and will still happen) it will be no longer needed and simply will need to be fixed ;p

m-kolomanski:
## Issue

Closes #265

## Description
This PR is focused on refactoring a tech debt that was previously taken due to the structure of the app. With a data flow refactoring efforts, debt can be paid and the structure improved.

Currently, the tlg-related modules use global **session$userData$data** object to fetch the data for display.  This is in order to avoid verbosely passing data objects down the line, however it might lead to confusing data flow where it is not clear where the data comes from and what kind of data the module works with. When the application is refactored and modularized, and with workflow improvements this is no longer needed and passing the correct data object down the stream would be more readable and easier to maintain.

This PR also adds some more docstrings to TLG-related modules.

## How to test
Run the application, map the data, go to TLG tab and submit an order. Verify that the functionalities work correctly and TLG modules have access to the data.

## Contributor checklist
- [x] New logic is documented


rabiibouhestine (approved):
LGTM

js3110 (approved):
Looks good to me! everything seems to be working

PR #340, opened 2025-04-11
Bug: Summary statistics table not considering PPSTRESU

### LINKED ISSUE #284: Bug: Summary statistics table not considering PPSTRESU
### Description
The summary statistic table when aggregating data is not considering if the groups present different PPSTRESU, which can lead to erroneous calculations when the user aggregates the analytes. This is a minor problem, as most users won't do this, but still something that may be better be addressed to prevent related issues. 

### To reproduce
Map data > Select multiple analyte > Change units for tmax in only one analyte > Run NCA > Check summary table  without grouping by analyte

### Expected behaviour
Units should be standardized to one  before doing the statistic so the results are reliable. However this should only affect the summary statistics table output

Gero1999:
## Issue

Closes #284 

## Description
The summary statistic table when aggregating data is not considering if the groups values present different PPSTRESU, which can lead to erroneous calculations when the user aggregates the analytes or when they use the package function . This is a minor problem, as most users won't do this, but still something that may be better be addressed to prevent related issues.

The implementation makes sure the function performs a conversion to the mode unit within the group. Tests were also added to cover the case. 
> Note: Additionally, when units are empty such as in `R2ADJ`, there won't be empty brackets in the output (i.e, `R2ADJ[]`)

## Definition of Done
- [x] Proper statistic calculations 
- [x] Column units represent always the used unit
- [x] Optional: `R2ADJ` does not show brackets

## How to test
Map data > Select multiple analyte > `Change units for tmax in one run / not in the other for Metabolite` > Run NCA > Check summary table without grouping by analyte. 

Values are equivalent to the case when unit changes were or not performed

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer


m-kolomanski (approved):
LGTM

js3110 (approved):
LGTM, thanks!

PR #343, opened 2025-04-14
Enhancement/add tests pivot wider pknca results (#287) & create fixture test datasets

Gero1999:
Contributes to #287 

### Main implementation
* Added tests to `tests/testthat/test-pivot_wider_pknca_results.R` for 100% coverage (covr). Included checks for reshaping main intervals, columns, handling AUC intervals, rounding numeric values, adding appropriate labels, and handling exclude values correctly.
- Created new testing data for all cases and independent of package functions. 
- Found bug in bioavailability for created testing data. Needed to change `bioavailability.R`

### Additional modifications
* Updated the `@param myres` documentation to specify additional assumptions about the input data, such as CDISC denomination of time variables and the inclusion of specific columns like `type_interval`, `PPSTRES`, and `PPSTRESU`

* Added a new column `time_col` to handle time-related data more flexibly.

## Definition of Done

- [x] tests have a 100% coverage for `pivot_wider_pknca.R` (`covr`)
- [x] tests seem logical

## How to test
- check tests in file `tests/testthat/test-pivot_wider_pknca_results.R`

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer


R/pivot_wider_pknca_results.R
```diff
@@ -6,7 +6,15 @@
 #' each column name corresponds with a calculated parameter and between brackets its
 #' corresponding units. AUC intervals, if present, are be added as additional columns.
 #'
-#' @param myres The output of PKNCA::pk.nca
+#' @param myres The output of PKNCA::pk.nca. It makes some additional assumptions:
+#'   1) CDISC denomination of actual and nominal time variables (AFRLT, ARRLT, NFRLT, NRRLT).
+#'   2) Intervals must include a column (`type_interval`) to differentiate between the
+#'   custom AUC ranges ("manual") and main parameter calculations ("main").
+#'   3) Includes `PPSTRES` and `PPSTRESU` variables in results dataset.
+#'   4) Columns `start_dose` and `end_dose` must express the actual start and end times
```
js3110, line 14: `start_dose` and `end_dose` are the start and end time relative to the most recent dose, whereas `start` and `end` are the times relative to the first dose

tests/testthat/test-pivot_wider_pknca_results.R
```diff
@@ -0,0 +1,200 @@
...
+  ) %>%
+  # Function assumes start_dose, end_dose, PPSTRES, PPSTRESU
+  mutate(
+    start_dose = start - !!sym(myres$data$dose$columns$time),
+    end_dose = end - !!sym(myres$data$dose$columns$time),
+    PPSTRESU = PPORRESU,
+    PPSTRES = PPORRES,
+    # Function assumes PPTESTCD is following CDISC standards
+    PPTESTCD = translate_terms(PPTESTCD, "PKNCA", "PPTESTCD")
+  )
```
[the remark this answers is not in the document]
Gero1999: @m-kolomanski  Two questions regarding this:

1) The dummy example used in `setup.R` is unfortunately very limited for most tests. Perhaps I can make a new one more complex using as reference the one in `test-interval_helpers.R`. Would that work and do you think would be worthy?

2) Should we make this `setup.R` used for all tests dependent on the functions `PKNCA_create_data_object`, `PKNCA_update_data_object`,  `PKNCA_calculate_nca`. Is that ok?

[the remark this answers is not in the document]
m-kolomanski: 2 is a very good point that I have completely overlooked - we should not rely on functions in the package to create testing data for the package. In the testing pipeline, `setup.R` is run first, technically we have not tested those functions yet, so they might generate invalid results, which will then to incorrect tests. Thank you for bringing this up.

This setup will need to be reworked.

Creating global testing object would be optimal, but if we require a lot of customizations between tests, then it makes sense to go the bespoke route and setup the data before every test, so we have greater control over it.

tests/testthat/setup_testing_data.R
```diff
@@ -0,0 +1,312 @@
...
+# 3) PKNCA results include `PPSTRES` and `PPSTRESU` variables.
+#    - Ensures compatibility with CDISC standards.
+# 4) `start_dose` and `end_dose` columns express the actual start and end times of the dose,
+#    relative to the first dose given to the subject.
+#    - Used for dose interval calculations.
+# 5) CDISC denomination of PK parameters (e.g., LAMZNPT, LAMZLL, LAMZ).
+#    - Temporarily needed to derive LAMZNPT and LAMZMTD.
+
+# Create Testing Concentration Data
+TEST_CONC_DATA <- data.frame(
```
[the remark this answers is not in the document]
Gero1999: just to produce a testing dataset that covers a variety of PKNCA data scenarios and is independent from other aNCA functions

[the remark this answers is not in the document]
m-kolomanski: I agree with @js3110 - we should keep all the pre-testing setup to a single `setup.R` file and consolidate testing data.

tests/testthat/test-pivot_wider_pknca_results.R
```diff
@@ -0,0 +1,165 @@
+# Use the testing data to produce results
+pwres <- pivot_wider_pknca_results(TEST_PKNCA_RES)
```
m-kolomanski, line 2: issue: We should not run the tested function outside of the testing suite. I think in this case, we should also test if the output is any valid before going into more detailed tests. It should look something like this:

```R
describe("pivot_wider_pknca_results", {
  pwres <- expect_no_error(pivot_wider_pknca_results(TEST_PKNCA_RES))

  it("produces a data.frame", {
    expect_s3_class(pwres, "data.frame")
  })
  ...
```

tests/testthat/test-pivot_wider_pknca_results.R
```diff
@@ -0,0 +1,165 @@
...
+
+#' Validate PKNCA Parameters
+#'
+#' This function validates that the reshaped PKNCA results contain the expected values
+#' for each parameter based on the original `TEST_PKNCA_RES` object.
+#'
+#' @param TEST_PKNCA_RES The original PKNCA results object.
+#' @param pwres The pivoted results from `pivot_wider_pknca_results`.
+#' @return A logical value indicating whether all checks passed.
+#' @export
```
m-kolomanski, line 12: issue: `@returns`
issue: `@export` is not needed, this is not a package function. In addition, this clashes semantically - functions with `.` should be local functions, so by definition not exported anywhere else.

tests/testthat/test-pivot_wider_pknca_results.R
```diff
@@ -0,0 +1,165 @@
...
+      "LAMZSPN", "LAMZIX",
+      "LAMZMTD", "Exclude"
+    )
+    expect_true(all(expected_columns %in% colnames(pwres_only_main)))
+
+    # Check number of rows is the expected (each dose has 1 row in the pivoted data)
+    expect_true(nrow(pwres_only_main) == nrow(res_only_main$data$dose$data))
+
+    # Check parameter values match the ones in the PKNCA results object
+    expect_true(.validate_pknca_params(res_only_main, pwres_only_main))
```
m-kolomanski, line 96: issue: Should be more like
```suggestion
    expect_no_error(.validate_pknca_params(res_only_main, pwres_only_main))
```
since if there is an issue, the validate function will throw an error - and this is what we want to have reported. Then the validate function can return whatever (or nothing) and the test will still be as informative. The returned value is not relevant in the context of testing - the errors are.

tests/testthat/test-pivot_wider_pknca_results.R
```diff
@@ -0,0 +1,165 @@
...
+      "NFRLT", "NRRLT", "ADOSE", "ROUTE", "ADOSEDUR",
+      "CMAX[ng/mL]", "TMAX[hr]", "TLST[hr]", "LAMZ[1/hr]",
+      "R2", "R2ADJ", "LAMZLL[hr]",
+      "LAMZNPT[count]", "CLSTP[ng/mL]", "LAMZHL[hr]",
+      "LAMZSPN", "LAMZIX", "LAMZMTD", "Exclude",
+      "AUCINT_0-2[hr*ng/mL]", "AUCINT_2-4[hr*ng/mL]"
+    )
+    expect_true(all(expected_columns %in% colnames(pwres)))
+
+    expect_true(.validate_pknca_params(TEST_PKNCA_RES, pwres))
```
m-kolomanski, line 115:
```suggestion
    expect_no_error(.validate_pknca_params(TEST_PKNCA_RES, pwres))
```

tests/testthat/test-pivot_wider_pknca_results.R
```diff
@@ -0,0 +1,165 @@
...
+
+  it("rounds numeric values to three decimals", {
+    expected_num_param_cols <- c(
+      "CMAX[ng/mL]", "TMAX[hr]", "TLST[hr]", "LAMZ[1/hr]",
+      "R2", "R2ADJ", "LAMZLL[hr]", "LAMZNPT[count]", "LAMZSPN",
+      "AUCINT_0-2[hr*ng/mL]", "AUCINT_2-4[hr*ng/mL]"
+    )
+    rounded_values <- pwres[, expected_num_param_cols]
+    expect_true(all(apply(rounded_values, 2,
+                          function(x) all(na.omit(abs(x - round(x, 3))) < 1e-9))))
```
m-kolomanski, line 126: issue: This should be part of the setup. No need to split the setup in two parts, makes the check itself seem more complicated than it is.

```suggestion
    rounded_values <- pwres[, expected_num_param_cols] |>
      apply(2, function(x) all(na.omit(abs(x - round(x, 3))) < 1e-9))
    expect_true(all(rounded_values))
```

m-kolomanski (changes requested):

tests/testthat/test-pivot_wider_pknca_results.R
Gero1999: @m-kolomanski  I refactored to this, hopefully looks more clear. Let me know if not!

```
  it("rounds numeric values to three decimals", {
    expected_num_param_cols <- c(
      "CMAX[ng/mL]", "TMAX[hr]", "TLST[hr]", "LAMZ[1/hr]",
      "R2", "R2ADJ", "LAMZLL[hr]", "LAMZNPT[count]", "LAMZSPN",
      "AUCINT_0-2[hr*ng/mL]", "AUCINT_2-4[hr*ng/mL]"
    )
    actual_values <- pwres[, expected_num_param_cols]
    expected_values <- round(actual_values, 3)
    diff_values <- actual_values - expected_values
    
    all_non_nan_values_are_rounded <- all(is.na(diff_values) | diff_values == 0)
    expect_true(all_non_nan_values_are_rounded)
  })
```

tests/testthat/setup_testing_data.R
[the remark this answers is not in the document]
Gero1999: ok I will work on consolidating this new testing data with the bioavailability tests so we can rely on them for `setup.R`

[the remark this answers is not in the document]
Gero1999: @js3110 @m-kolomanski while doing this I found an issue/bug when using the testing data on the `calculate_F()` for `bioavailability.R`. I solved it in the last commit, all tests keep passing and results are the same. I did it in the last commit so let me know if you think would be worthy to open this in a new branch or solve it directly here!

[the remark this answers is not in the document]
m-kolomanski: _Technically_, I think the changes are significant enough that they should be in separate PR, BUT I understand that the work has already been done and it will take time to revert it and create a separate PR, so as long as the changes are described in the PR I am fine with it.

js3110:
Hi @Gero1999 , what was the bug in bioavailability?
This new version doesn't work and crashes the app when I try with data for bioavailability (eg DummyRO_MRoutes)

I would suggest to just revert the changes, and keep the dummy test data it used as well for now, then it can be modified in a separate issue.

R/pivot_wider_pknca_results.R
```diff
@@ -88,7 +97,7 @@ pivot_wider_pknca_results <- function(myres) {
       filter(type_interval == "manual", startsWith(PPTESTCD, "AUCINT")) %>%
       mutate(
         interval_name = paste0(
-          signif(start_dose), "-", signif(end_dose),
+          signif(start), "-", signif(end),
```
js3110, line 100: I would say we should keep this relevant to ref dose, not from first dose

tests/testthat/setup.R
```diff
@@ -1,18 +1,314 @@
-# Import dataset from testthat/data folder
-
-DUMMY_DATA_FIXTURE <- read.csv(testthat::test_path("data", "adnca_dummy_sm_dataset.csv"))
```
js3110, line 3: I think its still worth keeping this dummy data, as its more representative of real life data rather than the manual made dataset. We just can remove the processing of it using our functions, and process it manually instead. Not necessarily for this specific PR but its good to keep it for future tests.

js3110 (changes requested):
I added some comments :)

R/pivot_wider_pknca_results.R
js3110: otherwise there will be a separate column for every dose, which doesn't look great and is not so useful for the User.

Gero1999: ups I agree this should be by reference, thanks for the catch!

tests/testthat/setup.R
```diff
@@ -1,3 +1,320 @@
...
+# 3) PKNCA results include `PPSTRES` and `PPSTRESU` variables.
+#    - Ensures compatibility with CDISC standards.
+# 4) `start_dose` and `end_dose` columns express the actual start and end times of the dose,
+#    relative to the first dose given to the subject.
+#    - Used for dose interval calculations.
+# 5) CDISC denomination of PK parameters (e.g., LAMZNPT, LAMZLL, LAMZ).
+#    - Temporarily needed to derive LAMZNPT and LAMZMTD.
+
+# Create Testing Concentration Data
+TEST_CONC_DATA <- data.frame(
```
m-kolomanski, line 24: nitpick: I think a better name would be `CONC_DATA_FIXTURE`.

tests/testthat/setup.R
```diff
@@ -1,3 +1,320 @@
...
+  is.excluded.hl = FALSE,
+  is.included.hl = FALSE,
+  exclude_half.life = FALSE,
+  # Units
+  AVALU = "mg/L",
+  RRLTU = "h"
+)
+
+# Create Testing Dose Data
+TEST_DOSE_DATA <- data.frame(
```
m-kolomanski, line 118: nitpick: Same as above, `DOSE_DATA_FIXTURE`

tests/testthat/setup.R
```diff
@@ -1,3 +1,320 @@
...
+      unique(),
+    by = "DOSNO",
+    relationship = "many-to-many"
+  )
+auc_intervals[, all_params] <- FALSE
+auc_intervals <- auc_intervals %>%
+  mutate(
+    aucint.last = TRUE
+  )
+INTERVALS <- rbind(main_intervals, auc_intervals) %>%
```
m-kolomanski, line 229: nitpick: Similarly, `INTERVALS_FIXTURE`

tests/testthat/setup.R
```diff
@@ -1,3 +1,320 @@
...
+    USUBJID == 5 & DOSNO == 1 ~ "start_conc0",
+    USUBJID == 5 & DOSNO == 2 ~ "start_conc0",
+    USUBJID == 6 & DOSNO == 1 ~ "start_conc0",
+    USUBJID == 6 & DOSNO == 2 ~ "start_conc0",
+    USUBJID == 7 & DOSNO == 1 ~ "start_conc0",
+    USUBJID == 7 & DOSNO == 2 ~ "start_conc0",
+    TRUE ~ NA_character_
+  ))
+
+TEST_PKNCA_DATA <- PKNCA::PKNCAdata(
```
m-kolomanski, line 247: nitpick: PKNCA_DATA_FIXTURE.

tests/testthat/setup.R
```diff
@@ -1,3 +1,320 @@
...
+                               time.nominal = "NFRLT"),
+  units = PKNCA::pknca_units_table(
+    concu = "ng/mL", doseu = "mg/kg", amountu = "mg", timeu = "hr"
+  )
+)
+TEST_PKNCA_DATA$intervals <- INTERVALS
+TEST_PKNCA_DATA$options <- list(keep_interval_cols = c("DOSNO", "type_interval"))
+
+# Add start_dose and end_dose columns
+TEST_PKNCA_RES <- withCallingHandlers(
```
m-kolomanski, line 260: nitpick: `PKNCA_RES_FIXTURE`

tests/testthat/setup.R
```diff
@@ -1,3 +1,320 @@
...
+      ifelse(
+        end %in% c(2, 5),
+        1,
+        2
+      )
+    }
+  )
+#####################################################################
+
+TEST_DOSE_DATA_TO_JOIN <- select(
```
m-kolomanski, line 294: nitpick: I am presuming this variable will not be used in actual tests? If so, I would suggest using lowercase name.

tests/testthat/setup.R
```diff
@@ -15,4 +332,12 @@ PKNCA_DATA_FIXTURE$intervals <- format_pkncadata_intervals(
...
-PKNCA_RESULTS_FIXTURE <- PKNCA_calculate_nca(PKNCA_DATA_FIXTURE)
+PKNCA_RESULTS_FIXTURE <- withCallingHandlers(
+  PKNCA_calculate_nca(PKNCA_DATA_FIXTURE),
+  warning = function(w) {
+    # Suppress warnings matching the regex "Too few points for half-life"
+    if (grepl("^Too few points for half-life", conditionMessage(w))) {
+      invokeRestart("muffleWarning")
+    }
+  }
+)
```
m-kolomanski, line 343: question: Would we be able to get rid of this with the implementation of proper fixtures in this PR? I am guessing this part stays in for support of older tests that use it and we will update those test to use new fixtures in separate PRs?

I would love to have a `#TODO` comment added here, explaining why we have two `PKNCA_RESULTS` object fixtures.

tests/testthat/test-create_start_impute.R
```diff
@@ -1,107 +1,73 @@
-# Create concentration, dose and interval datasets with all imputation cases
-conc_data <- data.frame(
-  conc = c(1:3, c(1, 2, 0.5), 1:3, 6:4, 6:4, rep(2, 3), 1:3, 6:4),
-  time = c(0:2, c(1, 2, 2.5), 4:6, c(1, 2, 2.5), 4:6, c(1, 2, 2.5), c(1, 2, 2.5), c(1, 2, 2.5)),
-  route = c(rep("extravascular", 9), rep("intravascular", 15)),
-  param = c(rep("A", 21), rep("B", 3)),
-  USUBJID = c(rep(1, 3), rep(2, 6), rep(3, 6), rep(4, 3), rep(5, 3), rep(6, 3)),
-  DOSNO = c(rep(1, 3), rep(rep(1:2, each = 3), 2), rep(1, 3), rep(1, 3), rep(1, 3))
-)
+pknca_data <- TEST_PKNCA_DATA
```
m-kolomanski, line 1: question: Do we need to _localize_ the fixture, can't we just use it directly in tests?

tests/testthat/test-pivot_wider_pknca_results.R
```diff
@@ -0,0 +1,182 @@
+#' Validate PKNCA Parameters
+#'
+#' This function validates that the reshaped PKNCA results contain the expected values
+#' for each parameter based on the original `TEST_PKNCA_RES` object.
+#'
+#' @param TEST_PKNCA_RES The original PKNCA results object.
+#' @param pwres The pivoted results from `pivot_wider_pknca_results`.
+#' @returns A logical value indicating whether all checks passed.
+.validate_pknca_params <- function(TEST_PKNCA_RES, pwres) {
```
m-kolomanski, line 9: issue: Arguments to functions probably should not be uppercase, since they are not constant by definition.

tests/testthat/test-pivot_wider_pknca_results.R
```diff
@@ -0,0 +1,182 @@
...
+
+    have_similar_vals <- all(round(result_vals, 3) %in% round(original_vals, 3))
+    have_same_length <- length(result_vals) == length(original_vals)
+    if (!(have_similar_vals && have_same_length)) {
+      stop(paste("Mismatch in values for parameter:", param_col))
+    }
+  }
+
+  # If all checks pass, return TRUE
+  TRUE
```
m-kolomanski, line 66: nitpick: No need to return anything in this function.

tests/testthat/test-pivot_wider_pknca_results.R
```diff
@@ -0,0 +1,182 @@
...
+      "CMAX[ng/mL]", "TMAX[hr]", "TLST[hr]", "LAMZ[1/hr]",
+      "R2", "R2ADJ", "LAMZLL[hr]",
+      "LAMZNPT[count]", "CLSTP[ng/mL]", "LAMZHL[hr]",
+      "LAMZSPN", "LAMZIX",
+      "LAMZMTD", "Exclude"
+    )
+    expect_true(all(expected_columns %in% colnames(pwres_only_main)))
+
+    # Check number of rows is the expected (each dose has 1 row in the pivoted data)
+    expect_true(nrow(pwres_only_main) == nrow(res_only_main$data$dose$data))
```
m-kolomanski, line 98: nitpick:
```suggestion
    expect_equal(nrow(pwres_only_main), nrow(res_only_main$data$dose$data))
```

tests/testthat/test-pivot_wider_pknca_results.R
```diff
@@ -0,0 +1,182 @@
...
+    expected_columns <- c(
+      "USUBJID", "start", "end", "DOSNO", "AFRLT", "ARRLT",
+      "NFRLT", "NRRLT", "ADOSE", "ROUTE", "ADOSEDUR",
+      "CMAX[ng/mL]", "TMAX[hr]", "TLST[hr]", "LAMZ[1/hr]",
+      "R2", "R2ADJ", "LAMZLL[hr]",
+      "LAMZNPT[count]", "CLSTP[ng/mL]", "LAMZHL[hr]",
+      "LAMZSPN", "LAMZIX",
+      "LAMZMTD", "Exclude"
+    )
+    expect_true(all(expected_columns %in% colnames(pwres_only_main)))
```
m-kolomanski, line 95: question: Do we expect some more columns to be created in the process? This test would ignore them. Do we want that? Maybe a better test would be
```R
expect_setequal(expected_columns, colnames(pwres_only_main))
```

tests/testthat/test-pivot_wider_pknca_results.R
```diff
@@ -0,0 +1,182 @@
...
+    expected_columns <- c(
+      "USUBJID", "start", "end", "DOSNO", "AFRLT", "ARRLT",
+      "NFRLT", "NRRLT", "ADOSE", "ROUTE", "ADOSEDUR",
+      "CMAX[ng/mL]", "TMAX[hr]", "TLST[hr]", "LAMZ[1/hr]",
+      "R2", "R2ADJ", "LAMZLL[hr]",
+      "LAMZNPT[count]", "CLSTP[ng/mL]", "LAMZHL[hr]",
+      "LAMZSPN", "LAMZIX", "LAMZMTD", "Exclude",
+      "AUCINT_0-2[hr*ng/mL]", "AUCINT_2-4[hr*ng/mL]"
+    )
+    expect_true(all(expected_columns %in% colnames(pwres)))
```
m-kolomanski, line 118: question: Same question as before with the expected columns.
question: Previous test had `nrow()` check, should this be included here as well?

m-kolomanski:
Hey, forgot to write in my review that this overall looks very nice, I only have some minor nitpick and questions regarding some test and the naming conventions for the fixtures - generally, it would be good to have the word `FIXTURE` in the names to signify that this data is fixed for the testing suite, as opposed to `TEST` data that suggest that it might be generated on per-test basis.

m-kolomanski:
Also, I am now thinking that we are creating a lot of intermediate variables when generating the fixtures. I think it might be worth to consider wrapping everything in `base::local({})` to stop those variables that are not needed bleeding into testing environments:
```R
# test-*.R currently
print(TEST_CONC_DATA) # prints the data
print(all_params) # prints all params

# setup.R
base::local({
  # global assignment for fixtures that will be used in tests
  TEST_CONC_DATA <<- data.frame( 
    ...
  )
  # local assignment for variables that are only used for creating fixtures, but not relevant to tests
  all_params <- setdiff(names(PKNCA::get.interval.cols()),
                        c("start", "end"))
})

# test-*.R
print(TEST_CONC_DATA) # prints the data
print(all_params) # Error: object `all_params` not found
```

Gero1999:
> Hey, forgot to write in my review that this overall looks very nice, I only have some minor nitpick and questions regarding some test and the naming conventions for the fixtures - generally, it would be good to have the word `FIXTURE` in the names to signify that this data is fixed for the testing suite, as opposed to `TEST` data that suggest that it might be generated on per-test basis.

Hey @m-kolomanski  maybe it is just me but I always found easier to identify the var in the env if it has a common prefix rather than a suffix. Perhaps would you be ok with using `FIXTURE_*` instead of `*_FIXTURE` ?

js3110 (approved):
lgtm :)

m-kolomanski:
@Gero1999 Prefix is absolutely fine as well.

m-kolomanski (approved):
Thank you for the changes, the setup and the tests look great!

PR #344, opened 2025-04-14
Enhancement: add tests flexible_violinboxplot #287

Gero1999:
## Issue

Contributes to #287 

### Main implementation
* Use of `vdiffr` to test ggplot outputs
* Added tests to `tests/testthat/test-flexible_violinboxplot.R` for 100% coverage (covr). Included checks for simple boxplot generation, boxplots with additional xvars, boxplots with additional colorvars, plots with varvalstofilter, cases with missing data, violin plots & plotly output generation (class testing)

 Made some minor changes in the function:
 * added arg plotly so output can be returned in plotly and ggplot (last one makes easier the testing)
 * refactor fun to consider empty string for PPSTRESU

## Definition of Done
- [x] tests have a 100% coverage for `flexible_violinboxplot.R` (`covr`)
- [x] tests seem logical

## How to test
- check tests in file `tests/testthat/test-flexible_violinboxplot.R`

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer


js3110 (approved):
Hey, looks good and I am happy to approve! Only thing I want to query:
would it be a bit better if we have the code for the RDS FIXTURE_PLOTS, rather than having it just straight uploaded? That way its a lot easier if it ever needs to change, as we can then create/load it directly, and its more transparent how it was created and what was included (which I feel is important for the testing)

Let me know what you think :)

m-kolomanski (commented):
The tests seem a bit convoluted, I see it is not easy to test those plot objects in a meaningful way.

Maybe we could move into [snapshot testing](https://testthat.r-lib.org/articles/snapshotting.html) for this purpose?

Using inbuild `testthat` functions we could snapshot the plot object that the function returns or even save it as an actual .png file. Then on the next test run the output of the function will be compared with the snapshot to see if anything has changed. This might be useful for this case, especially that if we make any deliberate changes to the plot, we can run a single function to accept a new snapshot, instead of having to manually re-generate the `.Rds` file.

Let me know what you think.

Gero1999:
hey @m-kolomanski @js3110 if you want I can try to integrate something to easily save the new plots

@m-kolomanski In my first commits I tried with snapshot but I did not manage to make it work. For that reason I went to this solution which I though might be more accurate than our current tests (class testing) as it is testing the data/layout of the object. If you could give me assistance with snapshot I would be glad to make the change!

m-kolomanski:
@Gero1999 

I have played around with the issue and now I have even more points on my _list I do not like about plotly_ 😜.

Each time the function is run, plotly generates `slightly` different object, meaning making snapshots is indeed difficult. I understand your approach of stripping irrelevant fields 100%. However, I would still implore to use in-build testthat snapshots for the ease of management, there is no need to build our own system.

I have pushed suggested implementation to branch called `enhancement/add_tests-flexible_violinboxplot-287-with-snaps`. It still requires some ironing out (moving helper functions to setup so they can be used all over the testing suites, adding documentation etc). If you like the approach, I can polish the implementation and merge that branch into this one. Let me know what you think!

Alternatively, we can also just use the plotly object and directly check specific, crucial values, like
```R
plotly_list <- plotly::plotly_json(simple_plot, jsonedit = FALSE, pretty = FALSE) |> jsonlite::fromJSON(b)

expect_equal(plotly_list$layout$xaxis$title$text, "DOSEA")
expect_equal(plotly_list$layout$yaxis$title$text, "cmax  [ ng/mL ]")
expect_equal(plotly_list$layout$shapes$type[1], "rect")
expect_equal(plotly_list$layout$yaxis$range, c(0.8, 5.2))
```

etc.

I think both approaches are valid, let me know what you think!

m-kolomanski:
@Gero1999 Following yesterdays discussion, I have implemented a `png` file snapshotting for `ggplot` objects on branch `enhancement/add_tests-flexible_violinboxplot-287-with-snaps2`. Please check it out and run the tests to see if they work for you well.

Again, let me know what you thing about the implementation. I could clean it up a little to reduce code duplication as well.

rabiibouhestine (approved):
LGTM

PR #347, opened 2025-04-15
Enhancement/add tests create start impute

Gero1999:
## Issue

Contributes to #287 

### Main implementation
* Added a test to verify that the function correctly imputes values when the `DRUG` column is missing but the analyte is present. (`tests/testthat/test-create_start_impute.R`,
* Added a test to ensure that the function can infer `DOSNO` when the column is missing from the intervals data. (`tests/testthat/test-create_start_impute.R`,

### Additional modifications
* Changed `group_by` and `arrange` to use `any_of` instead of `all_of` to avoid errors when some columns are missing. 
* Added filtering and deduplication steps after merging `mydata$intervals` to ensure that only relevant and unique rows are processed. 

## Definition of Done
- [x] tests have a 100% coverage for `create_start_impute.R` (`covr`)
- [x] tests seem logical

## How to test
- check tests in file `tests/testthat/test-create_start_impute.R`

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer


js3110 (changes requested):
LGTM, just small nitpick:

In the testthat file, when creating the intervals its currently written as:
```
auc.obs = TRUE,
  auc.pred = TRUE
```
these don't exist as parameters. So should be changed to something like `auclast`, `aucinf.pred` or something. 
This is giving me an error because my PKNCA is updated to dev version, where Bill has introduced more checks.

tests/testthat/test-create_start_impute.R
```diff
@@ -105,4 +105,38 @@ describe("create_start_impute", {
...
+    # No drug and no analyte
+    mydata_no_analyte <- mydata
+    mydata_no_analyte$dose$data$DRUG <- NULL
+    mydata_no_analyte$conc$data$param <- NULL
+    result_no_analyte <- create_start_impute(mydata_no_analyte)
+    # For last subject now analyte and drug are matching, should be start_logslope
+    expect_equal(result_no_analyte$intervals %>%
+                   dplyr::filter(USUBJID == 6, DOSNO == 1) %>%
+                   dplyr::pull(impute),
+                 "start_logslope")
```
m-kolomanski, line 128: nitpick: Since there is some logic to extract the result, I would assign it to a variable instead of directly inside `expect_equal` call.
```suggestion
    result_with_analyte_impute <- result_with_analyte$intervals %>%
                   dplyr::filter(USUBJID == 6, DOSNO == 1) %>%
                   dplyr::pull(impute)
    expect_equal(result_with_analyte_impute, "start_logslope")

    # No drug and no analyte
    mydata_no_analyte <- mydata
    mydata_no_analyte$dose$data$DRUG <- NULL
    mydata_no_analyte$conc$data$param <- NULL
    result_no_analyte <- create_start_impute(mydata_no_analyte)
    # For last subject now analyte and drug are matching, should be start_logslope
    result_no_analyte_impute <- result_no_analyte$intervals %>%
                   dplyr::filter(USUBJID == 6, DOSNO == 1) %>%
                   dplyr::pull(impute)
    expect_equal(result_no_analyte_impute, "start_logslope")
```

m-kolomanski (approved):
✅ LGTM, minor nitpick to consider.

js3110 (approved):
LGTM

PR #349, opened 2025-04-15
Enhancement: add tests for lambda slope plot

js3110:
## Issue

Contributes to #287 

## Description

Added testing for `lambda_slope_plot.R` to reach 100% coverage.
Refined the function where needed to improve the tests.

## Definition of Done

- [x] Tests implemented for lambda_slope_plot.R
- [x] 100% coverage achieved

## How to test

`covr::package_coverage`

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

I had to fix the additional tests to be able to check coverage, so those are also done in here. Will probably create conflicts since I think theyve been changed in a couple other PRs, but that can be dealt with when merging


tests/testthat/test-lambda_slope_plot.R
```diff
@@ -0,0 +1,115 @@
...
+        )
+      )
+    
+    plotly_output <- lambda_slope_plot(
+      conc_pknca_df = conc_pknca_df,
+      row_values = row_values,
+      myres = myres_mod
+    )
+    
+    expect_s3_class(plotly_output, "plotly")
```
m-kolomanski, line 47: question: If this is about handling missing points gracefully, would `expect_no_error` make sense here? Can we somehow check if the points were handled correctly? As far as we know, the function could return an empty or distorted plot.

tests/testthat/test-lambda_slope_plot.R
```diff
@@ -0,0 +1,115 @@
...
+    conc_modified <- conc_pknca_df
+    conc_modified$AVAL <- -1
+    
+    expect_warning({
+      empty_plot <- lambda_slope_plot(
+        conc_pknca_df = conc_modified,
+        row_values = row_values,
+        myres = myres
+      )
+      expect_s3_class(empty_plot, "plotly")
```
m-kolomanski, line 60: question: Can we somehow check if the plot is truly empty? Could we possibly run `.plotly_empty_plot` and compare the results directly?

I am myself unsure if this is needed, since we are catching the warning anyway, and it only can appear in one place - before returning empty plot, so might be redundant.

m-kolomanski (commented):
Hey, looks fine, but the checks are failing. I will approve once the checks are passing.

tests/testthat/test-lambda_slope_plot.R
js3110: i think since I catch the warning anyway, we don't need it.

js3110:
@m-kolomanski Ive updated the tests, they are passing on my local computer, however in the check(oldrel) they are failing. I think this is because locally I have the new PKNCA (where the assert_intervals function is implemented) whereas in this PR its for the old PKNCA.

So either we ignore this fail, or I remove the changes I made to keep_interval cols and just accept that it will fail for me locally. Thoughts?

m-kolomanski (approved):
LGTM, thanks!

Gero1999 (approved):
Maybe would be good as a suggestion to think about testing something else apart from being a plotly object? (just to check that the plot is the one we expect)  But this covers 100% so feel free to merge, good job!

PR #350, opened 2025-04-15
Bug: NCA can run before data is updated

### LINKED ISSUE #341: Bug: NCA can run before data updated
### Description
The settings updates in the data are set to a debounce of 2500, however the Run NCA button is always available to click. This means that if you update the settings but do not way 2.5 secs before clicking Run, the NCA will run with the old data, not the new settings.

### To reproduce
Update settings and click run immediately after. See that your settings have not been included

### Expected behaviour
Run NCA button should be disabled / also have a debounce of 2500 so it is only clickable when settings have been updated.

m-kolomanski:
## Issue

Closes #341

## Description
The settings updates in the data are set to a debounce of 2500, however the Run NCA button is always available to click. This means that if you update the settings but do not way 2.5 secs before clicking Run, the NCA will run with the old data, not the new settings.

This PR introduces JS function that disables a button for a given period of time. This is then applied to the Run NCA button to disable it when any setup changes are detected, for a time equal to settings debounce + 250ms.

## Definition of Done

Run NCA button should be disabled / also have a debounce of 2500 so it is only clickable when settings have been updated.

## How to test

Run the application, map the data, go to NCA tab. Change anything in the settings (add Analyte, add Dose Number, add Specimen). See if the button is disabled and placeholder text is displayed. Check if the button is re-enabled after ~3 seconds.

## Contributor checklist
- [x] New logic is documented


rabiibouhestine (approved):
LGTM

PR #356, opened 2025-04-22
Bug: Manual slope for single ANALYTE/PCSPEC not working

### LINKED ISSUE #355: Bug: manual slopes doesn't work for single unique ANALYTE/PCSPEC
### Description
If you use the DummyROADNCA data set, then add an exclusion and try to run NCA, it gives an error message. The console says:
```
ERROR [2025-04-22 16:36:08] Error calculating NCA results:
Can't select columns that don't exist.
✖ Column `PCSPEC` doesn't exist.
```

Caused by a select that names the variables when it should be using slope_groups.
### To reproduce
use the DummyROADNCA data set, then add an exclusion and try to run NCA

### Expected behaviour
No crash

js3110:
## Issue

Closes #355 

## Description

Removed hard coding of variables where the manual slopes are filtered to check for REASON. As the hard coded variables are not always in the data object.

## Definition of Done

- [x] No bug when running manual exclusions
- [x] Error message still appears if there is no reason added

## How to test

DummyRO_ADNCA data -> add manual exclusion but no reason -> run NCA

## Contributor checklist
- [x] New logic is documented


m-kolomanski (approved):
LGTM

Gero1999 (approved):
lgtm!

PR #359, opened 2025-04-23
Bug: nca intervals based on dosno

### LINKED ISSUE #161: Bug: NCA intervals are based DOSNO
### Description
The current implementation is using a set and estimated TAU for multiple dose studies. After #138 intervals are customized based on the actual cycle (DOSNO; dose number). However, we are planning to change DOSNO to planned (nominal) DOSNO, which means for the intervals to remain accurate, DOSNO should be removed from intervals calculation.

At some point, TAU will also become an optional column, so it would be good to also set this manually in the app (eg if TAU column not available, TAU is set as (end - start).
TAU is calculated as next dose actual time - current dose actual time using EX domain variables in SDTM, so we can do the same in the app (end time is next dose, start time is current dose time).

### To reproduce
format data intervals function in format_data.R uses DOSNO as a grouping variable. This can be removed, and time_dose used instead. Time_Dose is the time of dose, as calculated using AFRLT and ARRLT, so is the actual time of actual doses.

### Expected behaviour
Calculate intervals based on actual times, not using DOSNO.

js3110:
## Issue

Closes #161 , redo of #345 

## Description

Realised there was a bug in the duplicate creation so there was no IX created if the data only had one dose. Also an issue when running on multiple PCSPEC

## Definition of Done

## How to test

## Contributor checklist
- [ ] New logic is documented

## Notes to reviewer

### DIFF
--- a/R/dose_profile_duplicates.R
+++ b/R/dose_profile_duplicates.R
@@ -45,7 +45,11 @@ dose_profile_duplicates <- function(conc_data,
 
   #If only one dose, return the original data
   if (n_distinct(conc_data[[dosno]]) == 1) {
-    return(conc_data)
+    return(conc_data %>%
+             group_by(across(all_of(groups))) %>%
+             arrange(across(all_of(c(groups, arrlt)))) %>%
+             mutate(IX = seq_len(n())) %>%
+             ungroup())
   }
   # Step 1: Identify the dosing times (ARRLT == 0)
   dose_times <- conc_data %>%


m-kolomanski (approved):
LGTM

m-kolomanski (approved):
Hey, looks fine!

js3110:
@Gero1999 PTAL, same as before but minor modifications to fix the bugs

R/PKNCA.R
```diff
@@ -108,7 +106,7 @@ PKNCA_create_data_object <- function(adnca_data) { # nolint: object_name_linter
 
   pknca_dose <- PKNCA::PKNCAdose(
     data = df_dose,
-    formula = DOSEA ~ TIME | STUDYID + PCSPEC + DRUG + USUBJID,
+    formula = DOSEA ~ TIME_DOSE | STUDYID + PCSPEC + DRUG + USUBJID,
```
Gero1999, line 109: quest: why not removing `PCSPEC` already?

Gero1999 (approved):
lgtm! only one quest

R/PKNCA.R
js3110: hey! I did this in a separate PR #374 since in this one I hadn't yet noticed the issue. I'll probs change the scope of that one anyway since EVID not an ADNCA column, and make it just about refining the dosing data creation!

PR #361, opened 2025-04-24
Bug: Summary stats crash with bioavailability

### LINKED ISSUE #357: Bug: summary stats crash when bioavailability calculated
### Description
If you run on preclinical data, the summary stats crash as the bioavailability parameter causes an error in units conversion.

### To reproduce
Run Dummy preclinical data (currently in testing setup folder i think) and add bioavailability- app crashes.

### Expected behaviour
No crash.

js3110:
## Issue

Closes #357 

## Description

Summary statistics units conversion was crashing when bioavailability calculated, as the results were only in PPSTRES and not PPORRES. Fixed by adding results to both.

## Definition of Done

- [x] Bioavailability calculation runs fine
- [x] Summary statistics generated

## How to test

MRoutes data set for bioavailability

## Contributor checklist
- [x] New logic is documented


m-kolomanski (approved):
LGMT!

PR #366, opened 2025-04-25
Chore: Run main workflow on merge with main

m-kolomanski:
I have noticed that we skip the main workflow on merges with `main` - the checks are only run on the feature branch during PR, but skipped on merge commit. This should make no substantial difference - we are usually merging main into feature branch before merging into main, so the checks run before the merge are performed on identical code as after merging. However, this means checks on merge commits are skipped and lead to a red cross with `failure` label on main branch. With this change, the checks will run again after merge, giving the main branch a green checkmark and additionally confirming that everything is in order after merge.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,7 +43,6 @@ jobs:
   spellcheck:
     name: Spelling
     uses: ./.github/workflows/spellcheck.yml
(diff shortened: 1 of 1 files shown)


rabiibouhestine (approved):
LGTM

Gero1999 (approved):
lgtm

PR #377, opened 2025-04-30
Fix: Page counters in `slope_selector` going past desired range

### LINKED ISSUE #212: Bug: Page Navigation Counter Continues Incrementing Beyond Last Plot
### Description
When navigating through the plots in the `Slope plots` section, if a patient search is set and the "Next Page" button is pressed after reaching the last available plot, the counter continues incrementing beyond the last plot. As a result, when trying to go back using the "Previous Page" button, the user must press it multiple times—equal to the extra increments—before returning to the previous plot.

### To reproduce
1) Search for a patient.
2) Navigate forward using the "Next Page" button.
3) Continue pressing "Next Page" until reaching the last plot.
4) Press "Next Page" again – the counter keeps increasing even though no new plot is displayed.
5) Attempt to navigate back using "Previous Page" – multiple presses are required to return to the last visible plot.

### Expected behaviour
- [ ] When reaching the last plot, pressing "Next Page" should not increment the counter beyond the available plots.
- [ ] The "Previous Page" button should always navigate correctly without requiring excessive presses.

> Originally identified in #177

m-kolomanski:
## Issue

Closes #212 

## Description

Buttons for selecting page for slope selector plots were not working correctly. Clicking *Previous* on the first page would crash the app, just as pressing *Next* on the last page.

Proposed solution is not ideal and the pagination system could use a rework to make the end user experience better. However, it does solve the issue at hand for now - let me know if this is good enough at the moment.

## Definition of Done

- [x] When reaching the last plot, pressing "Next Page" should not increment the counter beyond the available plots.
- [x] The "Previous Page" button should always navigate correctly without requiring excessive presses.

## How to test

Map the data, go to NCA -> Setup -> Slope selector. Try pressing the *Previous* button on the first page. Navigate to the last page and try pressing the *Next* button.

## Contributor checklist

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.0.0.9067
+Version: 0.0.0.9068
 Authors@R: c(
     person("Ercan", "Suekuer", , "ercan.suekuer@roche.com", role = c("aut", "cre"),
            comment = c(ORCID = "0009-0001-1626-1526")),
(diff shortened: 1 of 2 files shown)


rabiibouhestine (approved):
LGTM

PR #386, opened 2025-05-06
test: add tests for PKNCA_add_F

Gero1999:
Contributes to #287 

### Main implementation
- Added tests for PKNCA_add_F

## Definition of Done
- [x] tests have a 100% coverage for `bioavailability.R` (`covr`)
- [x] tests seem logical

## How to test
- check tests in file `tests/testthat/test-bioavailability.R

## Contributor checklist
- [x] New logic is documented
- ~Package version is incremented~

### DIFF
--- a/tests/testthat/test-bioavailability.R
+++ b/tests/testthat/test-bioavailability.R
@@ -28,3 +28,26 @@ describe("calculate_F", {
     expect_true(any(is.na(result$f_auclast)) || any(result$f_auclast > 0))
   })
 })
+
+describe("PKNCA_add_F", {
+  it("returns the original object when bioavailability is NULL", {
+    result <- PKNCA_add_F(PKNCA_RESULTS_FIXTURE, NULL)
+    expect_equal(result, PKNCA_RESULTS_FIXTURE)
+  })
+
+  it("adds bioavailability data to the result slot when provided", {
+    bioavailability <- calculate_F(PKNCA_RESULTS_FIXTURE, c("f_auclast", "f_aucinf.obs"))
+
+    result <- PKNCA_add_F(PKNCA_RESULTS_FIXTURE, bioavailability)
+
+    expect_true(all(c("f_auclast","f_aucinf.obs") %in% result$result$PPTESTCD))
+
+    expected_res <- result$result %>%
+      filter(PCSPEC == "Plasma",
+             PPTESTCD %in% c("f_auclast", "f_aucinf.obs"))
+    expect_equal(result$result$PPSTRES[result$result$PPTESTCD == "f_auclast"][1:5],
+                 c(NA, NA, NA, 26.899, 19.98260), tolerance = 1e-4)
+    expect_equal(result$result$PPSTRES[result$result$PPTESTCD == "f_aucinf.obs"][1:5],
+                 as.double(c(NA, NA, NA, NA, NA)), tolerance = 1e-4)
+  })
+})


Gero1999:
I am not sure why the GitHub workflow fails, it does not for me locally. Let me know however if you find the issue @m-kolomanski @js3110

js3110 (changes requested):
Hi, 

I am unable to test because I get an error:

```
Warning: Values from `Dose` and `vals` are not uniquely identified; output will contain list-cols.
• Use `values_fn = list` to suppress this warning.
• Use `values_fn = {summary_fun}` to summarise duplicates.
• Use the following dplyr code to identify duplicates.
  {data} |>
  dplyr::summarise(n = dplyr::n(), .by = c(STUDYID, PCSPEC, DRUG, USUBJID, PARAM, start, end, DOSNO, type_interval, PPTESTCD, exclude, PPORRESU, PPSTRESU, PPSTRES, start_dose, end_dose, conv_factor, std_route)) |>
  dplyr::filter(n > 1L)
Warning: Error in mutate: ℹ In argument: `AUCdn_IV = PKNCA::pk.calc.dn(vals_intravascular, Dose_intravascular)`.
Caused by error in `parameter / dose`:
! non-numeric argument to binary operator
```

Using DummyMRoutes dataset, and selected all F auc options

js3110:
Hi @Gero1999 , the initial reason for having two separate functions (`calculate_f`, and `PKNCA_add_f`) was because I thought it was more useful to have a function that worked as a standalone bioavailability calculation, rather than only working within our aNCA workflow with PKNCA. Maybe @m-kolomanski could give your opinion on whether this is useful or not?

Also could you explain the reason for adding the dose normalisation? When `pk.calc.f` already does that (thats why Dose is one of the parameters). To me it seems like an unnecessary additional step.

m-kolomanski:
As far as my opinion goes, I prefer keeping those functions separate (one for doing the calculation, one for adding the result to the data object), as it adheres more to the functional programing principles - keeps functions more specialized and granular, makes it easier to maintain and test IMHO.

For the check, seems like the old version is failing. @Gero1999 if you have introduced any new packages / functions to the workflow, please do check what versions release and oldrel use, see if the behavior of those functions changed over the versions.

Gero1999:
Hey hey @m-kolomanski  @js3110 . For me personally keeping the CDISC/PKNCA standards for the output would make more sense for both the user and the App. It seems much consistent, as bioavailability should be like any other parameter when is calculated (and at some point, it will also be like this in PKNCA). However, I am open to change it back, although I really think is an unnecessary confusing step for the user.

I will check if I can solve that issue, I honestly did not test that particular dataset. Good catch, thanks!

Here the dose normalization is calculated previous to doing the mean, instead of the other way around. I did this because David (scientist in the team) pointed out to me that this is the correct procedure when grouping data.

m-kolomanski:
@Gero1999 @js3110 We can also have separate functions:
a) `calculate_F` that returns an arbitrary table
b) and `PKNCA_calculate_F` that calls the `calculate_F` function, but also does some transformations to return a CDISC/PKNCA compatible object.

That way
a) users that want bioavailability results will call one function (`calculate_F`)
b) and users that want bioavailability integrated into their results will call one function (`PKNCA_calculate_F`).

Feels like very easily we can have both here.

Gero1999:
@js3110 @m-kolomanski would it be ok if I add an argument in `calculate_F()` so the user can choose the format in which they want the output? (`PKNCA results` or  `pivoted PKNCA results`)

m-kolomanski:
I mean, from my perspective, this is also _fine_, but sounds like it should be more like `append_results = TRUE/FALSE`, if `FALSE` we return just bioavailability table, and if `TRUE` we append this to results object. Adding pivoting to the mix seems like this single function is doing to many different things. It is _fine_, as in it will work, but then the structure is more rigid, less flexible.

R/bioavailability.R
```diff
@@ -31,144 +31,156 @@
...
+      exclude = case_when(
+        is.na(vals_extravascular) & is.na(vals_intravascular) ~ paste0(
+          gsub("f_", "", PPTESTCD), " not available"
+        ),
+        is.na(vals_extravascular) ~ "Bioavailability is not calculated for IV records",
+        !is.na(AUCdn_IV) ~ "",
+        !is.na(Mean_AUCdn_IV_subj) ~ "Mean AUC.dn IV for the subject was used",
+        !is.na(Mean_AUCdn_IV_coh) ~ "Mean AUC.dn IV for the cohort was used",
+        TRUE ~ "Bioavailability: No individual, subject or cohort IV records to compare with"
+      ),
```
js3110, line 134: imo this is slightly overkill- i think its better to only specify if and how its been calculated, otherwise just leave it blank

js3110 (changes requested):
Hi, it looks good thanks! Just a note re the exclude comments,

firstly, I think its not needed to say if its not calculated.
Secondly, its currently not telling me in the exclude column how it was calculated, only for the IV records that its not available.

R/bioavailability.R
Gero1999: Indeed she confirmed me would like to keep all but IV records message. Let's do it like that then :)

tests/testthat/test-format_data.R
m-kolomanski, line 1: I am guessing this line addition was by mistake?

tests/testthat/test-bioavailability.R
```diff
@@ -1,30 +1,114 @@
...
+  filter(USUBJID %in% 1:6)
+
+describe("pknca_calculate_f", {
+  it("returns NULL for missing or unavailable AUCs, last case with a warning", {
+    expect_null(pknca_calculate_f(pknca_res, NULL))
+    expect_warning(
+      expect_null(pknca_calculate_f(pknca_res, c("f_AUCXXX", "f_AUCYYY"))),
+      "No AUC extracted from f_aucs available in res_nca \\(PPTESTCD\\): AUCXXX, AUCYYY"
+    )
   })
```
m-kolomanski, line 13: nitpick: Please move this test down and start the testing suite with a working example, instead of the function handling an exception.

tests/testthat/test-bioavailability.R
```diff
@@ -1,30 +1,114 @@
...
 
-  it("returns a data.frame", {
-    result <- calculate_F(PKNCA_RESULTS_FIXTURE, c("f_auclast"))
+  result <- pknca_calculate_f(pknca_res, c("f_AUCLST"))
```
m-kolomanski, line 15: suggestion: When tested function is run outside of `it()` statement, please wrap it in something like `expect_no_error`.

When run like this, if the function throws an error, it will be a generic `(code run outside of `test_that()`) ` error and it will stop execution.

When wrapped in a test, the code will fail and the rest of the suite will still run, giving you a possible better understanding of what exactly happened and what was affected.

R/bioavailability.R
m-kolomanski, line 1: issue: Please clean up `@importFrom` statements. For example, in `pknca_calculate_f` `bind_rows`, `full_join` and `reduce` are not used, but stated in documentation. On the opposite, `syms`, `ungroup`, `rename` are used, but not stated. There are asimilar issues in the other functions in this file.

m-kolomanski (approved):
Hey, the code itself is good, so I am approving the PR. I have found some minor issues with documentation and tests though, I am leaving a couple of comments, feel free to merge when those are addressed.

PR #387, opened 2025-05-07
Bug: summary stats download not filtered

### LINKED ISSUE #381: Bug: Summary stats download not filtered
### Description
Summary stats download will download all the parameters, not just the ones filtered by the user in the table selection.

### To reproduce
Descriptive statistics -> download after only selecting a couple of parameters/statistics

### Expected behaviour
Download should include only what is shown in the table.

m-kolomanski:
## Issue

Closes #381 

## Description

Summary stats download will download all the parameters, not just the ones filtered by the user in the table selection. This change filters the summary data table and selects appropriate columns before downloading a file.

## Definition of Done

- [x] Download summary stats only downloads parameters selected by the user.

## How to test

Run the application -> map the data -> run NCA with default parameters. Go into Results -> Descriptive statistics. Go to _Parameter to display_ and _Statistic to display_ widgets, play around with the selections - you can i.e. deselect all and select only the first one. The table should only display relevant columns with parameters and rows with relevant Statistic. This should also be true for the .csv file downloaded using button under the table. 

## Contributor checklist

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.0.0.9072
+Version: 0.0.0.9073
 Authors@R: c(
     person("Ercan", "Suekuer", , "ercan.suekuer@roche.com", role = c("aut", "cre"),
            comment = c(ORCID = "0009-0001-1626-1526")),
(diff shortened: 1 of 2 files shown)


js3110 (approved):
Looks good, thanks @m-kolomanski ! 
Non-blocking but maybe just a request, since were here, it would be good if the file download had a less generic name, maybe including the STUDYID or a timestamp (similar to main nca results download?)

rabiibouhestine (approved):
LGTM

PR #399, opened 2025-05-08
Enhancement: allow DOSNO to be character variable

js3110:
## Issue

Closes #

## Description

Change description.

## Definition of Done

## How to test

## Contributor checklist
- [ ] New logic is documented

## Notes to reviewer


inst/shiny/modules/tab_nca/nca_setup.R
```diff
@@ -224,7 +226,7 @@ nca_setup_server <- function(id, data, adnca_data) { # nolint : TODO: complexity
       updateSelectInput(
         session,
         inputId = "select_dosno",
-        label = "Choose the Dose Number:",
+        label = "Choose the profiles for the NCA:",
         choices = rows_for_selected_analytes$DOSNO,
```
m-kolomanski, line 230: question: Shouldn't `NCA_PROFILE` be referenced here as well?
```suggestion
        choices = rows_for_selected_analytes$NCA_PROFILE,
```

inst/shiny/modules/tab_nca/nca_setup.R
```diff
@@ -324,8 +326,8 @@ nca_setup_server <- function(id, data, adnca_data) { # nolint : TODO: complexity
       updateSelectInput(
         session,
         inputId = "select_dosno",
```
m-kolomanski, line 328: suggestion: I think it would be good to rename the widget as well.
```suggestion
        inputId = "select_nca_profile",
```

m-kolomanski (commented):
Hey, I have tested the app and everything seems to work fine, even with intentional weird selections, so good job!

I have found some typos and have some questions and suggestions to improve the code further - not strictly change requests, so I am leaving this as a comment, let me know what you think about the suggestions!

R/general_lineplot.R
```diff
@@ -76,41 +78,58 @@ general_lineplot <- function(
...
+        dose_time = dose_times$dose_time[match(new_NCA_PROFILE, dose_times$NCA_PROFILE)],
+        ARRLT = ARRLT + dose_time
+      ) %>%
+      filter(!is.na(new_NCA_PROFILE)) %>%
+      mutate(NCA_PROFILE = new_NCA_PROFILE) %>%
+      select(-new_NCA_PROFILE, -dose_time)
 
-    preprocessed_data <- rbind(predose_records, preprocessed_data)
+    # Combine the original and duplicated rows
+    preprocessed_data <- bind_rows(preprocessed_data, predose_dup)
```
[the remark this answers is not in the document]
js3110: I agree- and its actually very similar to dose_profile_duplicates, just with slightly different input data (no DOSNOA). So maybe I should rework that function to work for both? set an extra parameter for example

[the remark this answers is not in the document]
Gero1999: I think that is the best @js3110 

you can leave the column DOSNOA as default argument, that way the tests for dose_profile_duplicates will still work. Just add in test-dose_profile_duplicates a test to check that changing the argument works as expected!

If you use dose_profile_duplicates() here, this part of the code will already be covered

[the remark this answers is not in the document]
m-kolomanski: @js3110 Completely agree, great idea.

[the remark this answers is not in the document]
js3110: done :)

m-kolomanski (approved):
At the moment looks good to me! 🚀

The PR is quite complex, if any of the other reviews leads to significant changes, please re-request the review from me.

Gero1999 (approved):
I did not find any issue related with this branch, although I could not test completely `NCA_PROFILES` because `main` seems to already have a big issue with any `select_DOSNO > 5`. I created an issue #423 

With that said I think would be nice another deep review from someone else to prevent potential misses when multiple files are renamed @Shaakon35

Shaakon35 (changes requested):
image

I think NCA_Profile should be a factor by default

Shaakon35 (commented):
There are still some DOSNO variables in nca_setup.R
I wonder if that is correct because once you change data(), it does not contain DOSNO anymore.

If I rename my variable DOSNO in NCA, the NCA tab does not work anymore

js3110:
> There are still some DOSNO variables in nca_setup.R I wonder if that is correct because once you change data(), it does not contain DOSNO anymore.
> 
> If I rename my variable DOSNO in NCA, the NCA tab does not work anymore

good spot!! found it, was a leftover DOSNO that slipped through the cracks :) thank you!

R/format_data.R
```diff
@@ -213,19 +216,12 @@ format_pkncadata_intervals <- function(pknca_conc,
...
     mutate(!!!setNames(rep(FALSE, length(all_pknca_params)), all_pknca_params)) %>%
     mutate(across(any_of(params), ~ TRUE, .names = "{.col}")) %>%
 
-    # Prevent any potential attributes associated to the column names
-    mutate(across(everything(), ~ {
-      column <- .
-      attributes(column) <- NULL
-      column
-    })) %>%
-
```
js3110, line 228: @Gero1999 I have removed this code because @Shaakon35 suggested to set NCA_PROFILE as factor, and it doesn't work if this code stays as then there are downstream issues. I'm not sure what the reason was to strip the attributes in the first place, so idk if it is needed or not. Everything seems to work fine but maybe you had in mind special cases that would mean this code is required? lmk :) then we can just leave NCA_PROFILE as character or numeric. thanks!

Gero1999: @js3110 are you sure it is 100% needed to remove the code to make the column a factor? This code is to prevent issues with PKNCA, as PKNCA does not expect attributes internally and I already reported to Bill some issues when this happens. Perhaps give it a look please if you can do the column a factor without taking off the code!

js3110: what issues are there? it seems to be working fine for me are you sure they weren't solved in the latest update?
otherwise I can just refactorise it later on after the pknca calculations, but I want to make sure first that its still an issue, otherwise theres no point! :)

Gero1999: was 4 months ago, the issue was when running `pk.nca()`. if you are keen on it and the `pk.nca()` still works without it feel free to take it off :)

Issue #159, opened 2025-01-09
Enhancement: Incorporate Labels app-wide


js3110:
### Description
For a better user experience, we want any areas in the app that contain column names, to also include the column labels.

~In all data tables (as a subheader)~

In all selectInputs:
Visualisation section:
-  Variable Selection in all plots
- Boxplots grouping variables
~Optional (see what looks best):~
- ~Drag and drop in Descriptive Statistics (Visualisation)~
- 
Use virtualSelectInput, see example in https://github.com/pharmaverse/aNCA/pull/795

### Definition of Done
The requirements for the feature to be complete:
- [ ] Labels added to selectInput column names
- ~[ ] Labels implemented in all data tables~
- [x] Labels extracted from attr of dataset if RDS file uploaded

Example:
![Image](https://github.com/user-attachments/assets/311947b8-7b72-4d4f-b9d5-0564b3af42a6)

js3110:
Blocked until #158 merged

PR #400, opened 2025-05-09
Enhancement: alternative file types

### LINKED ISSUE #383: Enhancement: Upload of alternative file types
### Description
Currently the app only accepts .csv files. We would like to expand the possible file type uploads, to include:
- .rds
- .RData 
- SAS files (read_sas, haven package)
- xlsx
- parquet (read_parquet, arrow package)

The data should then be process from these inputs, and be able to continue as usual.

### Definition of Done
- [x] FileInput accepts multiple file types
- [x] Data created from each file type 
- [x] Data processed as normal
- [x] Check if the .RData still contains labels when created, e.g. attr(adpc[["AVAL"]], "label") should be equal to "Analysis Value" before uploading, and after uploading

m-kolomanski:
## Issue
Closes #383 

### Description
Currently the app only accepts .csv files. We would like to expand the possible file type uploads. The data should then be process from these inputs, and be able to continue as usual.

### Definition of Done
Formats to support:
- [x] .rds
- [x] SAS files (read_sas, haven package)
- [x] .xlsx
- [x] parquet (read_parquet, arrow package)

The requirements for each format handling to be complete:
- FileInput accepts multiple file types.
- Data created correctly.
- Data processed as normal.
- Check if the .rds still contains labels when created, e.g. attr(adpc[["AVAL"]], "label") should be equal to "Analysis Value" before uploading, and after uploading

## How to test
Run the application and try uploading files with different file formats. Make sure to proceed past mapping the data to make ensure `PKNCA` data object is created correctly.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer
If you find any issues with new file formats, please do attach the mock/test file to your comment - I have used already existing test data to create other formats for unit tests, original files that were generated using other software (SAS, Parquet) might be slightly different and cause issues.


rabiibouhestine (approved):
LGTM, I suggest adding `openxlsx2`, `haven`, `arrow` in `DESCRIPTION` as suggests

js3110:
@m-kolomanski is there any dummy data that we can test for this? maybe nice to at least add a Dummy.rds file, as then we can also test having labels

m-kolomanski:
@rabiibouhestine The packages are added to `Suggests`. The order is a bit confusing, it basically worked out to:
1. Packages used for testing and development
2. Suggested packages used within the package itself (the new libraries have been added here)
3. Packages used in the Shiny application.
In their respective groups the packages are sorted alphabetically.

@js3110 You can see the units tests on how to create a testing dataset for each file format - you can do the same, just saving normally instead of using temp file. Then you get files you can upload into the app. Would be cool if we could get our hands on some files generated by the appropriate software to test more thoroughly (I do not have access to those). Other than that I do not see a reason to duplicate the same testing data, just in different formats.

I do not see the reason to test the labels? If we add labels -> save rds -> read rds -> check labels, we are not really testing any part of our package, just the base functions, and it is not our responsibility to do so.

js3110 (approved):
LGTM! 
I had to install the xlsxl package, but it let me know so I think thats fine!

js3110:
> I do not see the reason to test the labels? If we add labels -> save rds -> read rds -> check labels, we are not really testing any part of our package, just the base functions, and it is not our responsibility to do so.

Hi, sorry, i should have worded it different! I don't mean test as in testthat testing, but rather that we later want to implement the reading and implementation of labels across the entire app where possible (#159), so thats why it will be useful in the future to have an RDS file with labels included. But can be done when we cross that bridge!

js3110:
@m-kolomanski im sorry I just pushed to your branch instead of mine, I am tired and trying to multitask ://// feel free to revert it

PR #407, opened 2025-05-13
rds file for dummy complex data

Shaakon35:
rds file for dummary complex data
I added all the labels, except NRRLT that has a label = NA

## Issue

Closes #

## Description

Change description.

## Definition of Done

## How to test

## Contributor checklist
- [ ] New logic is documented

## Notes to reviewer


m-kolomanski:
Hey, I don't quite understand why we need this? Seems like csv files cover more ground in terms of testing, as they lack the labels and the app needs to assign them itself. Why do we need a duplicate of data with labels specifically?

Shaakon35:
@m-kolomanski 
In case you have a variable that does not have a standard name (with CDISC labels), the variable will not have any pretty name in the whole app. For example, BALB, label = "Albumin at Baseline", this is a non- standard variable in ADPC, and therefore we would like to have the label name kept in the app. If you insert a csv, if the column is named BALB, there is no label associated to it. While with a rds file, it keeps the labels of the dataset.

Shaakon35:
Currently, this works perfectly as I want:
STUDYID didn't have a label, it is populated (because it's standard)
Name did have a label (not a standard variable), it kept the label
TEST didn't have a label, (not a standard variable), label have been set to the variable name
image

m-kolomanski:
@Shaakon35 
I understand what we want to achieve. This has nothing to do with .rds files or how they are loaded. .rds files will preserve attributes - they are just serializations of R data objects.

In #410 I have added a change that makes `apply_labels` function preserve any existing labels. It does not matter the source (although in the context of the application, .rds files are pretty much the only plausible source that would preserve the labels).

Unit tests implemented in #410 cover that behavior. I do not see any relation between that feature and loading .rds files.

Sorry, maybe I am missing something.

Shaakon35:
@m-kolomanski Perfect, before lunch I found out that I had to modify the function apply_labels and you just did it over lunch, thank you!
All good now. Then we can close this issue then :) the #410 is enough!

PR #415, opened 2025-05-15
bug: remove AUCINT from default parameters

### LINKED ISSUE #404: bug: AUCINT false in main parameters by default
### Description
Unless selected by the user AUCINT is not computed for each profile... It is rather more expected just to be computed for partial AUCs. 

To reproduce
Start the App > Mapping > NCA setup > Check parameters by default selected

Expected behaviour
AUCINT is not selected by default in the tab NCA setup

Gero1999:
## Issue

Closes #404

### Description
Unless selected by the user AUCINT is not computed for each profile... It is rather more expected just to be computed for partial AUCs. 

Definition of Done
AUCINT is not selected by default in the tab NCA setup

## How to test
Start the App > Mapping > NCA setup > Check parameters by default selected

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.0.0.9080
+Version: 0.0.0.9081
 Authors@R: c(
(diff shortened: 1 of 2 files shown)


Shaakon35:
@Gero1999 All good, do we still need start and end columns? It might be confusing for the T1 and T2 of the partial AUC.
image

PR #416, opened 2025-05-20
fix: default units table adapted to multiple units per analyte (e.g, PCSPEC) #382

### LINKED ISSUE #382: Bug: Default units issue for multiple analytes
### Description
When multiple analytes are considered in the App with different PCSPEC their units can change. This is something the App was not considering and can confuse in case the scientists want the parameters in those units. So we need to put them by default in those.

### To reproduce
Change in the currernt dataset the units of 1 PARAM in only 1 PCSPEC. See that the units will remain based on the first appearing row

### Expected behaviour
Default unit assignments should also consider PCSPEC

Gero1999:
## Issue

Closes #382

## Description
When multiple analytes are considered in the App with different PCSPEC their units can change. This is something the App was not considering and can confuse in case the scientists want the parameters in those units. So we need to put them by default in those.

## Definition of Done
- [ ] Default unit assignments should also consider PCSPEC (or any other conc group var)
- [ ] Filtering system for the table is not only for analytes

## How to test
- Check a dataset with multiple units for the same analyte (e.g, `Dummy_complex_data.csv`)
- Play with the units table (NCA setup > Select PCSPEC / Analyte > Units table)

## Contributor checklist
- [ ] New logic is documented

## Notes to reviewer

### DIFF
--- a/R/PKNCA.R
+++ b/R/PKNCA.R
@@ -144,21 +144,6 @@ PKNCA_create_data_object <- function(adnca_data) { # nolint: object_name_linter
     intervals = intervals, #TODO: should be default
     units = PKNCA_build_units_table(pknca_conc, pknca_dose)
   )
-browser()
-  # Update units
-  concu_vars <- setdiff(names(getGroups(pknca_conc)), names(getGroups(pknca_dose)))
-  units_table <- pknca_conc$data %>%
-    select(any_of(c(names(getGroups(pknca_conc)), "AVALU", "DOSEU", "RRLTU"))) %>%
-    unique() %>%
-    select_relevant_columns(o_conc$columns$concu)
-  unique_analytes <- unique(
-    pknca_data_object$conc$data[[pknca_data_object$conc$columns$groups$group_analyte]]
-  )
-  analyte_column <- pknca_data_object$conc$columns$groups$group_analyte
-  pknca_data_object$units <- tidyr::crossing(
-    pknca_data_object$units, !!sym(analyte_column) := unique_analytes
-  ) %>%
-    mutate(PPSTRESU = PPORRESU, conversion_factor = 1)
 
   pknca_data_object
 }
@@ -449,8 +434,8 @@ PKNCA_impute_method_start_c1 <- function(conc, time, start, end, ..., options = 
 
 PKNCA_build_units_table <- function(o_conc, o_dose) {
   # Extract relevant columns from o_conc and o_dose
-  group_dose_cols <- names(getGroups(o_dose))
-  group_conc_cols <- names(getGroups(o_conc))
+  group_dose_cols <- names(PKNCA::getGroups(o_dose))
+  group_conc_cols <- names(PKNCA::getGroups(o_conc))
   uconc_col <- o_conc$columns$concu
   utime_col <- o_conc$columns$timeu
   udose_col <- o_dose$columns$doseu
--- a/R/PKNCA.R
+++ b/R/PKNCA.R
@@ -113,7 +113,9 @@ PKNCA_create_data_object <- function(adnca_data) { # nolint: object_name_linter
     formula = AVAL ~ TIME | STUDYID + PCSPEC + DRUG + USUBJID / PARAM,
     exclude_half.life = "exclude_half.life",
     include_half.life = "include_half.life",
-    time.nominal = "NFRLT"
+    time.nominal = "NFRLT",
+    concu = "AVALU",
+    timeu = "RRLTU"
   )
 
   pknca_dose <- PKNCA::PKNCAdose(
@@ -121,7 +123,8 @@ PKNCA_create_data_object <- function(adnca_data) { # nolint: object_name_linter
     formula = DOSEA ~ TIME_DOSE | STUDYID + DRUG + USUBJID,
     route = std_route_column,
     time.nominal = "NFRLT",
-    duration = "ADOSEDUR"
+    duration = "ADOSEDUR",
+    doseu = "DOSEU"
   )
 
   # create basic intervals so that PKNCAdata can be created
@@ -139,15 +142,15 @@ PKNCA_create_data_object <- function(adnca_data) { # nolint: object_name_linter
     data.conc = pknca_conc,
     data.dose = pknca_dose,
     intervals = intervals, #TODO: should be default
-    units = PKNCA::pknca_units_table(
-      concu = pknca_conc$data$AVALU[1],
-      doseu = pknca_conc$data$DOSEU[1],
-      amountu = pknca_conc$data$AVALU[1],
-      timeu = pknca_conc$data$RRLTU[1]
-    )
+    units = PKNCA_build_units_table(pknca_conc, pknca_dose)
   )
-
+browser()
   # Update units
+  concu_vars <- setdiff(names(getGroups(pknca_conc)), names(getGroups(pknca_dose)))
+  units_table <- pknca_conc$data %>%
+    select(any_of(c(names(getGroups(pknca_conc)), "AVALU", "DOSEU", "RRLTU"))) %>%
+    unique() %>%
+    select_relevant_columns(o_conc$columns$concu)
   unique_analytes <- unique(
     pknca_data_object$conc$data[[pknca_data_object$conc$columns$groups$group_analyte]]
   )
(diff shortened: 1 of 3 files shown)


R/PKNCA.R
```diff
@@ -443,3 +424,215 @@ PKNCA_impute_method_start_c1 <- function(conc, time, start, end, ..., options =
...
+
+  # Join concentration and dose data
+  groups_units_tbl <- pknca_full_join_conc_dose(o_conc, o_dose) %>%
+    select(any_of(c(group_conc_cols, group_dose_cols,
+                    concu_col, amountu_col, timeu_col, doseu_col))
+    ) %>%
+    mutate(across(everything(), ~ as.character(.))) %>%
+    select(-any_of(c(o_conc$columns$subject, o_dose$columns$subject))) %>%
+    unique() %>%
+    select_relevant_columns(c(concu_col, amountu_col, timeu_col, doseu_col))
```
m-kolomanski, line 499: question: Why do we need to run `select()` three times? Can't we just select relevant columns once?

R/PKNCA.R
```diff
@@ -443,3 +424,215 @@ PKNCA_impute_method_start_c1 <- function(conc, time, start, end, ..., options =
...
+
+    # Combine all PKNCA units tables into one
+    unnest(cols = c(pknca_units_tbl)) %>%
+    select(-any_of(c(concu_col, amountu_col, timeu_col, doseu_col))) %>%
+    mutate(
+      PPSTRESU = PPORRESU,
+      conversion_factor = 1
+    ) %>%
+    select(any_of(c(group_conc_cols, group_dose_cols)),
+           PPTESTCD, PPORRESU, PPSTRESU, conversion_factor) %>%
```
m-kolomanski, line 523: question: Again, do we need two `select()` statements?

R/PKNCA.R
```diff
@@ -443,3 +424,215 @@ PKNCA_impute_method_start_c1 <- function(conc, time, start, end, ..., options =
...
+#' @details
+#' The function performs the following steps:
+#' 1. Extracts necessary columns from the concentration and dose objects.
+#' 2. Combines the data using `bind_rows`, adding a helper column to distinguish dose rows.
+#' 3. Fills values in the combined data.
+#' 4. Filters out dose rows and returns the final combined data frame.
+#'s
+#' @examples
+#' # Assuming `o_conc` and `o_dose` are valid PKNCA objects and present common groups:
+#' library(PKNCA)
```
m-kolomanski, line 545: issue: please don't load entire package in the `@examples`, namespace function calls instead.

R/PKNCA.R
```diff
@@ -443,3 +424,215 @@ PKNCA_impute_method_start_c1 <- function(conc, time, start, end, ..., options =
...
+#'   dose = 100,
+#'   doseu = "mg",
+#'   time = 0,
+#'   timeu = "h"
+#' )
+#' o_conc <- PKNCAconc(d_conc, conc ~ time | subj / analyte, concu = "concu", timeu = "timeu")
+#' o_dose <- PKNCAdose(d_dose, dose ~ time | subj, doseu = "doseu")
+#' combined_data <- pknca_full_join_conc_dose(o_conc, o_dose)
+#'
+#' @importFrom dplyr bind_rows mutate group_by arrange ungroup filter select
```
m-kolomanski, line 565: issue: some imports are missing: across, any_of, everything, technically also `%>%`

R/PKNCA.R
```diff
@@ -443,3 +424,215 @@ PKNCA_impute_method_start_c1 <- function(conc, time, start, end, ..., options =
...
+#' 3. Generates a PKNCA units table for each group, including conversion factors and custom units.
+#' 4. Returns a unique table with relevant columns for PKNCA analysis.
+#'
+#' @examples
+#' # Assuming `o_conc` and `o_dose` are valid PKNCA objects:
+#' # 1) Sharing group variables in their formulas
+#' # 2) Time units are the same within dose groups
+#' # 3) Units are the same for subjects within the same concentration group
+#'
+#' library(PKNCA)
```
m-kolomanski, line 451: issue: please do not import entire library in @examples, namespace function calls instead.

R/PKNCA.R
```diff
@@ -443,3 +424,215 @@ PKNCA_impute_method_start_c1 <- function(conc, time, start, end, ..., options =
...
+#'   doseu = "mg",
+#'   time = 0,
+#'   timeu = "h"
+#' )
+#' o_conc <- PKNCAconc(d_conc, conc ~ time | subj / analyte, concu = "concu", timeu = "timeu")
+#' o_dose <- PKNCAdose(d_dose, dose ~ time | subj, doseu = "doseu")
+#' units_table <- PKNCA_build_units_table(o_conc, o_dose)
+#' print(units_table)
+#'
+#' @importFrom dplyr select mutate rowwise
```
m-kolomanski, line 472: issue: some imports are missing: any_of, everything, technically `%>%`

R/PKNCA.R
```diff
@@ -443,3 +424,215 @@ PKNCA_impute_method_start_c1 <- function(conc, time, start, end, ..., options =
...
+      pknca_obj$data[[unit_colname]] <- pknca_obj$units[[unit_name]]
+    } else {
+      pknca_obj$data[[unit_colname]] <- NA_character_
+    }
+  }
+  pknca_obj
+}
+
+# Selects only the relevant columns from the data based on changes in the target column.
+select_relevant_columns <- function(data, target_col) {
```
m-kolomanski, line 612: question: Do we need to have this code wrapped in a separate function? Seems to me like unnecessary abstraction.

question: What makes the column relevant? Should either be included in the docstrings, or the function name could be more specific - what kind of columns are we selecting?

Also, sorry, but I do not quite understand the purpose of this function - would you mind explaining _what_ we are doing and _why_?

R/PKNCA.R
```diff
@@ -443,3 +424,215 @@ PKNCA_impute_method_start_c1 <- function(conc, time, start, end, ..., options =
...
+
+  # If there is no target_col specified, simply return the original data
+  if (is.null(target_col)) return(data)
+
+  # Convert all columns to numeric factors for comparison
+  data_binary <- data %>%
```
m-kolomanski, line 618: question: What makes the data _binary_? Does not look like binary data to me...?

tests/testthat/setup.R
```diff
@@ -339,10 +415,7 @@ base::local({
 
 # Dummy data
 # Import dataset from testthat/data folder
-
-# ToDo (Gerardo): These fixtures are supporting still test-bioavailability.R
-# We need to substitute them with the previous ones for consistency
```
m-kolomanski, line 344: question: Why was this `# TODO` removed? Seems like nothing has changed in that regard?

m-kolomanski (changes requested):

tests/testthat/setup.R
m-kolomanski: Ah, right, sure, but in that case we have missed removing the `PKNCA_DATA_FIXTURE` creation - we are not using them, so there is no reason to have them in the setup.

But this is another issue.

Gero1999:
@m-kolomanski improved and documented more the function. Hopefully now seems more clear, but feel free to lmk if not

R/PKNCA.R
```diff
@@ -445,3 +426,185 @@ PKNCA_impute_method_start_c1 <- function(conc, time, start, end, ..., options =
...
+    mutate(
+      PPSTRESU = PPORRESU,
+      conversion_factor = 1
+    ) %>%
+    # Order the columns to have them in a clean display
+    select(any_of(c(group_conc_cols, group_dose_cols)),
+           PPTESTCD, PPORRESU, PPSTRESU, conversion_factor)
+}
+
+# Ensures that specified unit columns exist in the PKNCA object, creating them if necessary.
```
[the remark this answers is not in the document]
Gero1999: added them, let me know if they feel enough :)

m-kolomanski:
Hey,

I have noticed that unit conversion factor is weirdly off when the target units are smaller than the source units (see screenshot).

Screenshot from 2025-06-03 13-34-17

Gero1999:
@m-kolomanski I used a DT function to change the diplaying format, hopefully now is fine, thanks a lot for the catch!

PR #425, opened 2025-05-26
fix: set PARAM as default for boxplot colors

### LINKED ISSUE #424: Bug: Boxplots not displaying by default
### Description
The user always needs to introduce a colorvar, which I feel is very confusing for most. Let's solve this small issue so the demo or first user experience of the App is not catches them out of guard

### To reproduce
Map data > Visualization > Boxplots

### Expected behaviour
When you go the the tab the results should already be displaying something by default

Gero1999:
## Issue

Closes #424

## Description
Allows boxplot to show by default. This will eliminate some confusion for the users when testing the App ( specially tomorrow! ;] )

## Definition of Done

Boxplots show by default once NCA is run.

## How to test

Map columns > Run NCA > Visualization > Boxplots
A default display should be there

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.0.0.9083
+Version: 0.0.0.9085
 Authors@R: c(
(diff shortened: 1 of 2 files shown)


js3110 (approved):
LGTM

PR #440, opened 2025-05-28
Data tab stepper nav

### LINKED ISSUE #439: Enhancement: Add stepper navigation in Data tab.
### Description
According to the figma files of the redesign, the Data tab features a stepper navigation highlighting the steps in the process of the data preparation.

### Definition of Done
- [ ] there should be a stepper at the top of the page
- [ ] there should be navigation buttons at the bottom of the page
- [ ] the stepper number should turn to blue at each corresponding step
- [ ] navigation buttons should work
- [ ] all functionality should be preserved

rabiibouhestine:
## Issue

Closes #439 

## Description

According to the figma files of the redesign, the Data tab features a stepper navigation highlighting the steps in the process of the data preparation.

## Definition of Done

- [x] there should be a stepper at the top of the page
- [x] there should be navigation buttons at the bottom of the page
- [x] the stepper number should turn to blue at each corresponding step
- [x] navigation buttons should work
- [x] all functionality should be preserved

## How to test

1. Open the app
2. Press ctrl/cmd + shift + R to hard refresh styles.
3. Check that navigation buttons work, and the stepper reacts to it (the circle color should change).
4. Check that all functionality in the data tab is preserved, and nothing is broken.

## Contributor checklist

## Notes to reviewer

- This PR aims to implement the stepper navigation without affecting any functionality.
- All functionality is preserved at this point, and will be handled in different PRs.
- The styling (and design) of the actual content inside each step will be handled in different PRs.
- The expected behaviour of the stepper component:
    - the circle color changes to blue at the corresponding step
- The expected behaviour of the navigation buttons at the bottom of the page:
    - Next button moves to next step
    - Previous button moves to previous step
    - Cancel button moves to first step
    - Buttons don't affect any logic (yet?)


rabiibouhestine:
@Gero1999 Thanks for the feedback! That's currently the expected behaviour of the 'cancel' button.

I currently set the buttons to only navigate and not interfere with logic until I get to the redesign of the content itself and see if there is any needed change in logic (for example, I believe the 'Next' button will play the role of 'Submit Mapping' in the new design of the mapping section).

For now I will rename the 'Cancel' button to 'Restart'.

m-kolomanski (approved):
Hey, looks awesome! I have two minor suggestions to consider:

suggestion: Should we disable previous / next button when the user is on the first / last step?

suggestion: @Gero1999 Should the first step be labeled as something like  "Upload" or "Import" or "Upload/Import data" instead of just "Data"?

Gero1999:
> suggestion: Should we disable previous / next button
I personally think it is still ok as a user. Functionally speaking it does not trigger any action, so I personally don´t see it as a problem. 

> Should the first step be labeled as something like "Upload" or "Import" 
Yes this is a good catch. `Import` or `Upload` could be more accurate, as the general tab in the left-sidebaar is already called `Data`

@rabiibouhestine @m-kolomanski

rabiibouhestine:
From a UX perspective, I actually think that the first step needs to have a preview of the 'default' dataset, perhaps below the Upload widget. It would let users know that they can already go to the next step without having to upload a dataset. It can also work as a confirmation the dataset has been successfully uploaded by previewing it there immediately.

For this PR, I will rename the steps as follows:

1. Data Import
2. Data Filtering
3. Data Mapping
4. Data Preview

And I will also implement disabling the buttons when they don't make sense to be enabled.


In later PRs, I think we should look at:

- Previewing the data in the first step
- Improving the startup experience by either or both:
  - Loading screen until default dataset is loaded and everything is initialised
  - Optimising the loading time

@m-kolomanski @Gero1999 WDYT?

Gero1999:
Hey @rabiibouhestine good observations as well!

> For this PR, I will rename the steps as follows:
> 1. Data Import
> 2. Data Filtering
> 3. Data Mapping
> 4. Data Preview

I would still suggest not to repeat the use of `Data`, should be implied by the general tab/actions. For 1. you can either use `Upload` or if you prefer `Import`. But if you use the last one, please also change the title that says `Upload` when you are at the section

> And I will also implement disabling the buttons when they don't make sense to be enabled.

Sounds good to me!

> Previewing the data in the first step

I agree, the same way as it is currently in main. If it is not problematic, perhaps could be changed back now before closing this pr (?). If it is more difficult, feel free to open a new issue as a `SHOULD`

> Improving the startup experience by either or both:
Loading screen until default dataset is loaded and everything is initialised
Optimising the loading time

Improving things sounds always bettter! 💪🏽 
However, I believe this is an independent `WISH` issue. The section never received user complains in the loading.

m-kolomanski:
@rabiibouhestine For me, all the suggestions sound awesome, it is just a matter of prioritization - which I believe @Gero1999 has explained.

PR #447, opened 2025-05-30
442 bug/error parameter datasets

### LINKED ISSUE #442: Bug: Error in Parameter datasets if Exclusions added
### Description
After adding manual exclusions or inclusions to the data, Parameter datasets section does not work

```
Error:
! `SUBJID` must be size 8 or 1, not 2152.
  136: export_cdisc [C:/aNCA/R/export_cdisc.R#122]
```

### To reproduce
In slope selector, add an exclusion for a subject, then Run NCA and view parameter datasets

### Expected behaviour
Should show results.

Gero1999:
## Issue

Closes #442 

## Description
- Solves issue in slope_selector action when exclusions or selections are performed. Data will stay ungrouped after it.
- Solves and independent potential issue found in derivation of PCSTRESU 
- Change the `dummyData.csv` to remove PCSTRESU column so we no longer assume it exists

## Definition of Done
- [x] Parameter datasets after running NCA with Exclusions/Selection in `Slope Selector`
- [x] PCSTRESU is present in ADPC

## How to test
In slope selector, add an exclusion for a subject, then Run NCA and view parameter datasets

## Contributor checklist
- [x] New logic is documented


js3110 (approved):
lgtm!

PR #448, opened 2025-06-02
428 enhancement/output folder

### LINKED ISSUE #428: Enhancement: Output folder with ADPP, ADPC, PP and other results
### Description
Allow a export button for all the regulatory submission objects

It will be advantageous that the App offers the possibility to export all result objects directly to a folder `/output`. As discussed, for following some standards the files should have a clever name style, and be organized in subfolders with different dates. In case of the current date already having a previous folder, version number is added:

### Definition of Done
- [ ] Export button produces a subfolder in the folder output with `STUDYID_DATE_version`
- [ ] ADPP.csv, ADPC.csv, PP.csv, Summary-Table.csv are exported in this subfolder
- [ ] TLGs are exported in this subfolder within another sub-subfolder called `TLGs`
- [ ] ❗New requirement: Output formats for data.frames include CSV, RDS, XPT

Gero1999:
## Issue

Closes #428

## Description
This pull request introduces functionality to save analysis results from various modules (NCA results, descriptive stats, parameter datasets) in the Shiny application to an output directory once the user presses `Run NCA`.

I also left an example on how the output folder will be structured, not only for the review but also for potential curious users that check our GitHub and want to know how our expected outputs look like 

### New utility function for saving outputs:

* Added the `save_output` function to save outputs (data frames, plots, or lists) to a specified directory. The function handles creating directories if they don't exist and supports saving lists of plots or data frames with appropriate naming conventions.

  - Added a reactive `results_dir` to dynamically generate the output directory path based on study ID and timestamp.
  - Updated the `nca_results_server` function to save pivoted NCA results to a CSV file using the `save_output` function.
- Updated the `parameter_datasets_server` function to save CDISC datasets (`pp`, `adpp`, and `adpc`) to the output directory using the `save_output` function. The same with the summary statistics table and the pivoted results.

## Definition of Done
- [x] Export button produces a subfolder in the folder output with `STUDYID_DATE_version`
- [x] ADPP.csv, ADPC.csv, PP.csv, Summary-Table.csv are exported in this subfolder (and pivoted results)

- [x] ❗**New requirement**: Output formats for data.frames include CSV, RDS, XPT

> Note: TLG exportation will be handled in a different issue #

## How to test

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer


Gero1999:
hey @KOBANAK I think you might have a very similar perspective to @Shaakon35 in how a Data Scientist would like the results to be named / organized for clear submissions. Let me know how this implementation looks to you. Once `run NCA` you will see all output files (except TLGs) automatically saved in your `inst/shiny/output` directory. Feedback always appreciated :)

m-kolomanski (commented):
I don't think this is the right approach:

- Without any system for removing old files, the output folder will become bloated quickly.
- In a server deployment, users will not be able to access the results at all (as they do not have access to the machine the app is run on, the `/inst/shiny/output` folder).
- In a server deployment, this can lead to unexpected loss of data (say, two users created unnamed projects at the same time).
- This is especially bad considering that in production settings (when user installs the package from github or CRAN, as opposed to cloning the repo), because then the folder with package system files will be bloated.
- As mentioned, the output folder will be located alongside system files for the package - which, from users perspective, will not be easy to find and/or manage at all.
- Seems like current rules for when the app overwrites and creates new output folder are a bit arbitrary - ~I am only allowed to create one output per project per day? Why?~ Sorry, I have just noticed that actually I am allowed one output folder per hour - seems even more arbitrary, if I click `Run NCA` at 13:58, 13:59 and 14, I will get two results, but those run with the settings at 13:58 will be lost.

Most importantly: why can't we just return a `.zip` file and let the user save it wherever they want and manage it however they want?

If we want the app to organize all project files itself, then we should ask the user for the output folder. Then use some kind of config file to save that. But then again, we would have to be clever when thinking about server deployment.

Furthermore, I think it would be much better for the users to provide a vignette with a description of the project structure, as opposed to outright example. This vignette could have some steps with functions / script to reproduce / create an example project output.

KOBANAK:
Hi,
I think that is a great feature for the shiny app in Improve. 
It would be nice if the subfolders could be organized under a main folder named using study name +"_aNCA" and structured like this :
- Inputs folder (adpc_datetime.csv)
- NCA folder (with results and every file related to the analysis, with the date and time as well)
- PK Parameters folder (with pp_datetime.csv, adpp_datetime.csv)
- and later a TLG folder

js3110:
i agree with @m-kolomanski , I don't think this is the best approach. Would be better to have a button for the user to click when they want to create the outputs. I do like the idea of having everything in one output though.

Especially if they are doing exploration and constantly tweaking and changing parameters, they might not want to output until the very end. Furthermore if we later include the non-nca analysis, then just clicking Run NCA is not enough to include those.

Gero1999:
@m-kolomanski you are right a `zip` downloadable seems a much better idea for this and simplifies the task. I will implement the changes!

Gero1999:
@js3110 @m-kolomanski As discussed with Rabii, the UI changes will be done later in another issue that will arrange everything as we have planned. Let me know though if the features of the implementation seem fine now!

js3110 (commented):
Looks amazing @Gero1999 thanks! I tested it and everything downloads really well :)
Other than UI comments (which I'll leave since its going to change anyway) I just have one request if its possible:
can you also add another folder in the output for the additional analysis section? That then downloads the ratios table if its been created?

Gero1999:
Now should include the matrix results & also as post-requested by @Shaakon35 the other formats (`rds`, `xpt`, `csv`)

js3110:
Looks nice! I'm not sure about exporting all 3 file types together, can be a lot if someone only wants a csv, I would suggest giving the user an option to select which outputs they want? Feel free to disagree @Shaakon35  but I think otherwise it may overwhelm the user if they dont want all of them.

js3110 (approved):

m-kolomanski (changes requested):
issue: 
I do not like the fact that the application just saves files as we go through the analysis - this is not a robust or scalable solution to this problem, and can lead to many unexpected consequences, depending on the exact deployment architecture. The application should only generate / save and export any files **when the user asks for it**, so when they click the download button.

The results should be gathered in memory in some kind of data structure, for example in `reactiveValues` (but that would need to be passed down to every sub-module) or simply via using `session$userData`. We could have the modules save the results to `session$userData$results` or something like this, and then only when the user clicks the download button, the elements from that object are run through `save_output()`, `.zip` is generated and forwarded to the user.

This also allows us to clean up the temp folder with results immediately, so that way we are tidy about any files we leave on the user device. 

Also, I do not see any reason for the results directory to be a `reactive()` - it does not really change during the session, and with the solution above, the temporary folder can be established when the files are generated.

inst/shiny/modules/tab_nca/nca_results.R
```diff
@@ -102,9 +103,31 @@ nca_results_server <- function(id, pknca_data, res_nca, settings, grouping_vars)
...
 
+    # Provide the zip file for download
+    output$download_zip <- downloadHandler(
+      filename = function() {
+        project <- session$userData$project_name()
+        datetime <- attr(res_nca(), "provenance")$datetime
+        paste0(project, "_", format(datetime, "%d-%m-%Y"), ".zip")
+      },
+      content = function(fname) {
+        output_tmpdir <- paste0(tempdir(), "/output")
```
m-kolomanski, line 114: nitpick: The recommended way to construct a folder path would be
```suggestion
        output_tmpdir <- file.path(tempdir(), "output")
```

inst/shiny/modules/tab_nca/nca_results.R
```diff
@@ -102,9 +103,31 @@ nca_results_server <- function(id, pknca_data, res_nca, settings, grouping_vars)
...
+
+        save_output(output = session$userData$results, output_path = output_tmpdir)
+        files <- list.files(
+          output_tmpdir, full.names = TRUE, pattern = ".[(csv)|(rds)|(xpt)]$", recursive = TRUE
+        )
+        files <- gsub(paste0(output_tmpdir, "/"), "", fixed = TRUE, files)
```
m-kolomanski, line 120: issue: We can receive the same result by just settting `full.names` parameter to `FALSE`.
```suggestion
        files <- list.files(output_tmpdir, pattern = ".[(csv)|(rds)|(xpt)]$", recursive = TRUE)
```

inst/shiny/modules/tab_nca/nca_results.R
```diff
@@ -102,9 +103,31 @@ nca_results_server <- function(id, pknca_data, res_nca, settings, grouping_vars)
...
+        setwd(output_tmpdir)
```
m-kolomanski, line 121: issue: Meddling with working directory on a running app is dangerous, but I understand that we want to get the correct folder structure within the `zip` file - this has always been annoying to me, how it works. We need to reset the working directory, otherwise we will cause unexpected behavior or crashes.
```suggestion
        wd <- getwd()
        on.exit(setwd(wd), add = TRUE) # this will reset the wd when download handler function is done
        setwd(output_tmpdir)
```

inst/shiny/app.R
```diff
@@ -43,6 +43,12 @@ ui <- function() {
     id = "sidebar",
     title = "aNCA",
 
+    div(
+      class = "project-name-container",
+      textInput("project_name", label = NULL, placeholder = "Project Name"),
+      icon("file", class = "project-name-icon")
+    ),
```
m-kolomanski, line 50: suggestion: My proposition would be to add the project name input to the top / title bar, I think it looks nice.
```suggestion
    title = tagList(
      span("aNCA"),
      div(
        class = "project-name-container",
        textInput("project_name", label = NULL, placeholder = "Project Name"),
        icon("file", class = "project-name-icon")
      )
    ),
```

m-kolomanski (changes requested):
issue: The feature does not work for me in its current form, it just produces a bunch of folders, see comments for details
Screenshot from 2025-07-04 07-44-18

inst/shiny/functions/save_output.R
```diff
@@ -0,0 +1,42 @@
...
+#' @returns Invisibly returns the file path written.
+save_output <- function(output, output_path) {
+
+  # Create output directory if it doesn't exist
+  dir.create(output_path, showWarnings = FALSE, recursive = TRUE)
+
+  for (name in names(output)) {
+
+    if (length(output[[name]]) > 1) {
+      save_output(output = output[[name]], output_path = paste0(output_path, "/", name))
```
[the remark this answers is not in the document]
Gero1999: ohhh this is why it was not working for you. I remember solving this precisely by checking the class but I may have forgotten to commit

PR #451, opened 2025-06-02
Bug: Logslope imputation error

### LINKED ISSUE #423: Bug: DummyRO with DOSNO > 6 does not allow pk.nca()
### Description
When running NCA with a selected DOSNO >5 (i.e, 6) the App won't run NCA providing the next error coming from `pk.nca`

```
Error during wrapup: ℹ In index: 14.
Caused by error in `assert_conc()`:
! Please report a bug.
: Error with interval start=840, end=1008: Assertion on 'conc' failed: Must be finite.
```

### To reproduce
Map data > NCA setup > Select DOSNO: 6 > Run NCA 

### Expected behaviour
App should run the NCA without an error message and producing the expected results

js3110:
## Issue

Closes #423 

## Description

Identified a small bug, when running DOSNO > 5 on dummy data, due to subj 11104 which was a BOLUS subject. Identified that the error was coming from the logslope imputation. Due to the function missing the specification of time_dose, the default was 0, and for a later dose this caused the back extrapolation of c0 to be Inf, which then caused the error message.

## Definition of Done

- [x] No error when running DOSNO > 6 on dummy data'
- [x] Logslope imputation correct

## How to test

Run NCA for DOSNO > 6 on dummy data

## Contributor checklist
- [x] New logic is documented


Gero1999 (approved):
LGTM as well as you know, good catch! :P

PR #464, opened 2025-06-10
Enhancement: set end = Inf for single dose studies

### LINKED ISSUE #462: Enhancement: End = Inf for single dose studies
### Description
For single dose studies, to ensure the appropriate parameters are calculated, end of interval should be set to Inf.
Since there is no end time to the dose profile.

### Definition of Done
- [ ] End = Inf if study is single dose
- [ ] NCA calculated smoothly
- [ ] No errors with default parameters

js3110:
## Issue

Closes #462 

## Description

For single dose studies, end should = Inf
Going by the assumption that if TAU column is not present in the data, and length(unique(data$DOSNOA)) == 1, then it is a single dose study.
If so, set end to = Inf

## Definition of Done

- [x] End = Inf for single dose studies
- [x] End = TAU for MD studies
- [x]  End = last AFRLT for MD studies with no TAU

## How to test

Filter for DOSNO == 1 in dummy data, then check Summary of NCA Setup.

## Contributor checklist
- [x] New logic is documented


Gero1999 (changes requested):
hey good idea, only left some considerations that might be worthy to take into account!

js3110:
i've updated @Gero1999 thanks for your comments. regarding the violin plot, i have no idea what changed, but the version now is correct (Cmax is max 5 for both doses- 15 was not possible with the data) Maybe it was using the old setup data?

R/format_data.R
```diff
@@ -220,11 +229,14 @@ format_pkncadata_intervals <- function(pknca_conc,
...
       case_when(
         !is.na(lead(TIME_DOSE)) ~ lead(TIME_DOSE),
+        is.na(TAU) & one_dose ~ Inf,
+        is.na(TAU) ~ max_end,
         TRUE ~ start + TAU
       )
     } else {
       case_when(
         !is.na(lead(TIME_DOSE)) ~ lead(TIME_DOSE),
+        one_dose ~ Inf,
```
Gero1999, line 239: issue: `one_dose` (or `is_single_dose`) is a transitory variable. In other words, is a column we create (and not PKNCA) to do a particular operation. These type of variables I would suggest taking them off once we use them, to keep the output object of the function "clean"

Gero1999 (approved):
@js3110  LGTM, good job! Only two remarks that I think can improve the code

PR #466, opened 2025-06-11
Fix: get conversion factor correctly

### LINKED ISSUE #465: Bug: incorrect units conversion for summary statistics
### Description
Dataset provided internally, wrong unit conversions. Not sure why

### To reproduce
Unable to reproduce open source

### Expected behaviour
If applicable, what do you think should happen instead?

js3110:
## Issue

Closes #465 

## Description

For a dataset tested internally at roche, the summary statistics were not applying the unit conversion to one group. On investigation, I found it was due to the handling of NA values in the unit conversion of the summary statistics. I fixed by using our `get_unit_conversion `function instead.

## Definition of Done

- [x] summary statistic units correct

## How to test

Unfortunately I was unable to recreate the bug in the dummy data, so hard to test without the dataset. If you are testing from Roche I can give you the data, otherwise just try to see if you think anything could be incorrect, sorry!

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

Let me know if you want the dataset to test.


m-kolomanski (approved):
LGTM!

Gero1999 (approved):
hey @js3110 thanks this makes sense to me but I still think we should also fix the problem that we discussed when we talked about this. The computation of `ModeUnit `& `ModeConvFactor` should be parallel and not independent, to prevent future issues.

However, this imp looks good, so If you want we can open a different issue for it!

Gero1999:
The problem is here:

```
  ModeUnit = names(sort(table(PPSTRESU), decreasing = TRUE, useNA = "ifany"))[1],
      ModeConv_factor = as.numeric(names(sort(table(conv_factor),
                                              decreasing = TRUE, useNA = "ifany"))[1]),
```

We need that `ModeConv_factor` is the corresponding `conv_factor` value for the `ModeUnit`, not the mode of the `conv_factor`. This can cause problems when we have multiple units working if, for example, the most common conversion factor is NA because we have different unrecognised transformations, but actually the most common unit is X (which has a recognised transformation)

Gero1999:
But actually it might be better to open a different issue. Because this problem is indeed not solving the root issue, which is the fact that we are doing 1 summary table for variables with different units (i.e `Analyte`). There should be automatically one table for each. So we can leave this weird issue for now and instead solve this later by changing a bit the visualization features.

PR #478, opened 2025-06-17
redesign data filtering

### LINKED ISSUE #477: Enhancement: Redesign Data Filtering
### Description
Redesign Data Filtering Step in Data Tab based on figma designs.

### Definition of Done
- [ ] All functionalities preserved
- [ ] Styles follow design pattern in the provided figma files

rabiibouhestine:
## Issue

Closes #477 

## Description

Redesign Data Filtering Step in Data Tab based on figma designs.

## Definition of Done

- [x] All functionalities preserved
- [x] Styles follow design pattern in the provided figma files

## How to test

1. Open App
2. Go to Data Filtering Step
3. Check that filters work (correctly filter the table)
4. Go to the next steps and check that filtered table is preserved

## Contributor checklist


rabiibouhestine:
@js3110 Thanks for the review! I actually realized that the numeric filters I implemented did not have an easy way for the user to :

1. find which filters were used (unless they remember)
2. reset the filters to the default values.

Sliders can solve this issue but very hard to use for float variables, and even for some integer variables.

For now, I opted for using the old filters, which solves these issues and also answers your concern about the length of the filters box.

Let me know if this is better now.

js3110:
@rabiibouhestine This is nice, much cleaner and easier to track now!
However, I did like the features you added in the first version where the character variables provide dropdown options, and for numeric you are able to set a min and max. With this version, I guess its possible to have min and max but would not be as easy...
Let me know if its possible to have these included, or if you think it would be too complex. In my head you could add some toggles that change the filter input box depending whether its character or numeric column, but maybe it sounds easier than it is lol.

Also, for this tab I would suggest making the reactable have `filterable = TRUE` so that the user can type filters for each individual column to check before applying the filters.

m-kolomanski (changes requested):
Hey, the UI rework looks great!

The PR seems to have a lot of commented out code - please remove it if it is not needed Also, I have noticed that the new functions introduced in `data_filtering_helpers.R` are used only in the code that is commented out. I guess they are also left over from dev work?

m-kolomanski:
@js3110 @Gero1999 I have picked this PR up, let me know what you thing of the current implementation - tried to take as much feedback from the thread in as possible.

Shaakon35 (commented):
Looks really great,

Can't we modify the numeric values so that we have a list like for the characters? (we can easily as.character the numeric variables, show the values available, and then as.numeric back fo the filtering. 
image

Shaakon35:
> Hey @Shaakon35, but I presume we do not want all numeric values to be like this, just specific ones? We can make a workaround and apply that, but we would need a way to determine which numeric values should have pick-list style input?

Hello @m-kolomanski , I see what you mean now.
If there are more than 100+ distinct numeric values, the list could be giant.
Let's set a limit to a maximum of 20 (if that is fine for you) distinct numeric values to show the available values; otherwise, leave the list empty. What do you think?

Shaakon35:
> @m-kolomanski I really like it, and I think it looks great! I agree with @Shaakon35 it would be nice to have an option to filter %in% for the numeric- for example DOSNO 1 and 5. But I don't think having a pickerInput the same as the character variables is good, but rather still text input (maybe if they put multiple values like 1,5,10 it filters as %in%)?

i think the picker input is a nice option, but we have to restrict it to a number of unique values (e.g. 20).
I we don't go for picker input, we need additional information on how to separate numeric values, because "," might be decimal point and that could be confusing

m-kolomanski:
@Shaakon35 @js3110 I have made it so that numeric values with less than 20 unique vals will appear as `pickerInput`, please let me know what you think.

Shaakon35 (approved):

PR #491, opened 2025-06-26
Add BLQ imputation options (#139)

### LINKED ISSUE #139: Enhancement: BLQ imputation Options
### Description
Option in the Data tab to allow the user to select which BLQ method should be applied to the dataset. The processed data is then used for NCA and for Plotting.

Suggested BLQ calculation options:

- Set BLQs before tmax to 0, and after tmax to Missing
- set the BLQ value to 1/2 of the lloq value.
-  1/3 Imputation rule

Make use of clean.conc.blq  function in PKNCA

### Definition of Done
- [ ] Option for user to select BLQ method, including option for None (ie BLQs already processed in data)
- [ ] If BLQ method chosen, Input options for User to define which column is LLOQ col, or to manually input LLOQ for each combination of PARAM, ASMED
- [ ] Data processed to mutate AVAL so that BLQs are processed in the chosen method.

Gero1999:
## Issue

Closes #139 

## Description

This PR allows users to flexibly specify BLQ imputation rules from the UI and propagate these rules through the NCA data processing pipeline. The changes include both backend logic for applying BLQ imputation and frontend UI updates to let users select their preferred BLQ handling strategy.

**BLQ Imputation Feature**

* Added support for user-defined BLQ imputation rules in NCA setup, enabling  PKNCA flexible strategies (e.g., positional, tmax-based, value assignment, or no handling) via the new `data_imputation_ui` and `data_imputation_server` modules,
* Extended the NCA pipeline to accept and apply BLQ imputation rules, updating relevant functions (`update_main_intervals`, `rm_impute_obs_params`)
* Updated `NEWS.md` and function documentation to announce and describe the new BLQ imputation capability. 
* Registered new global variables and helper functions required for BLQ imputation in the package metadata. 
- Imputations only affect non-observational parameters, by just affecting those intervals

## Definition of Done
Note: Considering our latest conversations this feat is only in order to incorporate the options that PKNCA support. But we are not doing it with actual intention of internal use. That said, I would suggest against overcomplicating the feature options far from what PKNCA supports

- [x] Allows PKNCA imputation based on `tmax`, `BLQ position` or allows to introduce a particular BLQ value (i.e `LLOQ/2`)
- [x] The process affects the NCA results when relevant (missing concentrations!)

## How to test
Map data > NCA settings > Test the different BLQ imputations in `Data Imputation` section

Use extreme value imputations (i.e, `10000`) and check if the `auclast` is changed for profiles with BLQ values. I made modifications in data to include BLQ points (AVAL = 0), one case is the first profile (USUBJID = 11101, PCSPEC = "CSF", NCA_PROFILE = 1)

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

~Blocked until we figure out what to do with the potential bug in PKNCA!~ No need

### DIFF
--- a/R/PKNCA.R
+++ b/R/PKNCA.R
@@ -211,6 +211,8 @@ PKNCA_create_data_object <- function(adnca_data, nca_exclude_reason_columns = NU
 #' concentration data. Each item in the list should have:
 #' - reason: character string with the exclusion reason (e.g., "Vomiting")
 #' - rows: integer vector of row indices to apply the exclusion to
+#' @param keep_interval_cols Optional character vector of additional columns
+#' to keep in the intervals data frame and when the NCA is run (pk.nca) also in the results
 #'
 #' @returns A fully configured `PKNCAdata` object.
 #'
@@ -227,7 +229,8 @@ PKNCA_update_data_object <- function( # nolint: object_name_linter
   selected_profile,
   selected_pcspec,
   should_impute_c0 = TRUE,
-  exclusion_list = NULL
+  exclusion_list = NULL,
+  keep_interval_cols = NULL
 ) {
 
   data <- adnca_data
@@ -239,7 +242,8 @@ PKNCA_update_data_object <- function( # nolint: object_name_linter
     progress = FALSE,
     keep_interval_cols = c(
       "ATPTREF", "DOSNOA", "type_interval",
-      adnca_data$dose$columns$route, "ROUTE"
+      adnca_data$dose$columns$route, "ROUTE",
+      keep_interval_cols
     ),
     min.hl.r.squared = 0.01
   )
@@ -251,9 +255,11 @@ PKNCA_update_data_object <- function( # nolint: object_name_linter
   data$intervals <- format_pkncadata_intervals(
     pknca_conc = data$conc,
     pknca_dose = data$dose,
-    start_from_last_dose = should_impute_c0
+    start_from_last_dose = should_impute_c0,
+    keep_interval_cols = keep_interval_cols
   ) %>%
     # Join route information
+    # TODO (Gerardo): Add ROUTE to keep_interval_cols
     left_join(
       select(
         adnca_data$dose$data,
--- a/R/intervals.R
+++ b/R/intervals.R
@@ -5,6 +5,9 @@
 #' @param pknca_conc A PKNCAdose object containing the concentration data.
 #' @param pknca_dose A PKNCAdose object containing the dose data.
 #' @param start_from_last_dose Logical defining if start is at time of last dose or C1.
+#' @param keep_interval_cols Optional character vector of additional columns
+#'  to keep in the intervals
+#' data frame.
 #'
 #' @returns A data frame containing the dose intervals and specified pharmacokinetic parameters.
 #'
@@ -33,7 +36,8 @@
 #' @export
 format_pkncadata_intervals <- function(pknca_conc,
                                        pknca_dose,
-                                       start_from_last_dose = TRUE) {
+                                       start_from_last_dose = TRUE,
+                                       keep_interval_cols = NULL) {
   if (!inherits(pknca_conc, "PKNCAconc")) {
     stop("Input pknca_conc must be a PKNCAconc object from the PKNCA package.")
   }
@@ -60,7 +64,10 @@ format_pkncadata_intervals <- function(pknca_conc,
 
   # Select conc data and for time column give priority to non-predose samples
   sub_pknca_conc <- pknca_conc$data %>%
-    select(any_of(c(conc_groups, "ARRLT", "ATPTREF", "DOSNOA", "TRTRINT", "VOLUME")))
+    select(any_of(c(
+      conc_groups, "ARRLT", "ATPTREF", "DOSNOA",
+      "TRTRINT", "VOLUME", keep_interval_cols
+    )))
 
   has_trtrint <- "TRTRINT" %in% names(sub_pknca_conc)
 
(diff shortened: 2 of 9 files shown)


inst/shiny/modules/tab_nca/setup/data_imputation.R
```diff
@@ -0,0 +1,124 @@
...
+        condition = sprintf(
+          "input['%s'] == 'Positional BLQ imputation'", ns("select_blq_strategy")
+        ),
+        blq_selectize(ns("before_first_non_blq"), "Before first non-BLQ", selected = "drop"),
+        blq_selectize(ns("in_between_non_blqs"), "In between non-BLQs", selected = "keep"),
+        blq_selectize(ns("after_last_non_blq"), "After last non-BLQ", selected = "drop")
+      ),
+      conditionalPanel(
+        condition = sprintf("input['%s'] == 'Set value for all BLQ'", ns("select_blq_strategy")),
+        blq_selectize(ns("blq_value"), "Value for BLQ", selected = "0.05"),
```
js3110, line 49: For this options, would it be better to have a numeric input option? why 0.05?

inst/shiny/modules/tab_nca/setup/data_imputation.R
```diff
@@ -0,0 +1,124 @@
...
+    ))),
+    br(),
+    br(),
+    selectInput(
+      ns("select_blq_strategy"),
+      "Select BLQ Imputation Strategy",
+      choices = c(
+        "tmax based imputation",
+        "Positional BLQ imputation",
+        "Set value for all BLQ"
```
js3110, line 27: can we also add an option for No BLQ handling?

js3110, line 27: Nitpick: formatting.
```suggestion
        "Tmax based imputation",
        "Positional BLQ imputation",
        "Set value for all BLQ"
```

js3110 (changes requested):
Hi,

I added some comments, and also I have noticed an issue in the handling of the BLQs, it seems to be dropping the last sample in all the options.

For example, subjid 11101 in the `main` branch has a slope with points 5, 6, 7, where the last point is time = 383.7, conc = 0.0063.
In this branch, that time point is gone and only 5 and 6 are there, which is not enough for slope calculations,

It seems to be the same trend for all the subjects.
This is not only an issue in the plots but also in the main results.

inst/shiny/modules/tab_nca/setup/data_imputation.R
Gero1999: 1. `numericInput` won't allow "drop" or "keep", just in case they want to remove (or not do anything) to all BLQ
2. Just to put a low number, no special reason. Lmk if you think something else could be better

inst/shiny/modules/tab_nca/setup/data_imputation.R
Gero1999: Maybe the last option could be called `Set all BLQ` instead of `Set Value for all BLQ`

Do you think that would make clearer the fact that can be used to also not do anyhting?

js3110: does "keep" mean do nothing? in that case yes maybe we can jsut make that more clear

Gero1999: nice catch

Gero1999:
@js3110 @m-kolomanski Not sure what is the best to do here. Due to the potential discovered bug already reported in PKNCA ([here](https://github.com/humanpred/pknca/discussions/447)), using a BLQ rule with `tmax.before = "drop"` or `first = "drop"` seems to cause the drop of the `start_conc0` imputation in extravascular studies. 

I opened a discussion in PKNCA to figure out if this is indeed the situation and needs a fix (https://github.com/humanpred/pknca/discussions/447) But maybe @billdenney can confirm if this is the issue or if he knows of a workaround for it

billdenney:
For the potential bug in PKNCA, is it a real use case where you would say that you want to drop early BLQ values but also you want to impute a BLQ value? It seems like the user is giving discordant instructions at that point.

If it is a real use case (rather than option interaction testing which is useful), can you please describe the scenario for this to be intentional?

R/PKNCA.R
```diff
@@ -222,7 +228,9 @@ PKNCA_update_data_object <- function( # nolint: object_name_linter
   selected_profile,
   selected_pcspec,
   params,
-  should_impute_c0 = TRUE
+  should_impute_c0 = TRUE,
+  # PKNCA default BLQ imputation rule
```
m-kolomanski: style: comment is redundant
```suggestion
```

inst/shiny/modules/tab_nca/setup/data_imputation.R
```diff
@@ -0,0 +1,134 @@
...
+}
+
+data_imputation_server <- function(id) {
+  moduleServer(id, function(input, output, session) {
+    ns <- session$ns
+
+    should_impute_c0 <- reactive({
+      input$should_impute_c0
+    })
+
```
m-kolomanski: suggestion: I would not assign this simple reactive, instead return it outright.
```suggestion
```

and then

```R
    list(
      should_impute_c0 = reactive(input$should_impute_c0),
      blq_imputation_rule = blq_rule
    )
```

R/intervals.R
```diff
@@ -189,9 +199,32 @@ update_main_intervals <- function(data, parameter_selections,
...
+  }
+
+  ############################################
+  # Define a BLQ imputation method for PKNCA
+  # and apply it only for non-observational parameters
+
+  if (!is.null(blq_imputation_rule)) {
+    PKNCA_impute_method_blq <<- function(conc.group, time.group, ...) { # nolint
+      PKNCA::clean.conc.blq(conc = conc.group, time = time.group, conc.blq = blq_imputation_rule)
+    }
```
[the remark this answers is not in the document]
Gero1999: @m-kolomanski will this be good enough?

js3110 (commented):
In DoD it says " allows to introduce a particular BLQ value (i.e LLOQ/2)" I don't see where this is possible?

inst/shiny/modules/tab_nca/setup/data_imputation.R
```diff
@@ -0,0 +1,131 @@
...
+    selectInput(
+      ns("select_blq_strategy"),
+      "Select BLQ Imputation Strategy",
+      choices = c(
+        "Tmax based imputation",
+        "Positional BLQ imputation",
+        "Set value for all BLQ",
+        "No BLQ handling"
+      ),
+      selected = "Tmax based imputation"
```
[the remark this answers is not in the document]
Gero1999: hmm you are right that for now we are preprocessing. I am sure we will eventually change, but until then yes I set it to by default keeping the BLQs

PR #492, opened 2025-06-26
Enhancement: add summary stats to ratios

### LINKED ISSUE #479: Enhancement: Add summary stats to tissue-plasma ratios
### Description
Tissue ratios should also include an option to calculate the mean/summary stats, and define how to group it.

### Definition of Done
- [ ] mean calculations for ratios
- [ ] grouping options

js3110:
## Issue

Closes #479 

## Description

Add mean and N summary stats to tissue-plasma ratios section

## Definition of Done

- [x] Summary stats for ratios

## How to test

Try creating ratios and group by different options

## Contributor checklist
- [x] New logic is documented


Gero1999 (approved):
slay! functionally speaking LGTM, but what would be amazing is to integrate the module and/or functions that we already have for the other summary stats. That way we only have to maintain and keep track of one :) 

Anyway, I am talking without getting into it, so if it feels actually tricky just ignore that, the way it is seems fine!

PR #498, opened 2025-06-27
Enhancement/auto upload of csvrds files

### LINKED ISSUE #480: Enhancement: auto upload of csv/rds files
### Description
Would be nice to have an optional parameter:
aNCA::run_app(file = "adpc.csv")
 so that when you open the app, your dataset is already loaded.

### Definition of Done
- [x] Make sure dataset file = "adpc.csv" is already loaded in the app once you open the app.
- [x] Different format checks also need to be done
- [x] Make sure it works for any path, and that the wd can be changed for ease
- [x] file argument is NULL by default, if NULL, nothing happens
- [ ] Make sure the new option works when the app is published on Rsconnect (with file = NULL, the default option)

lucyaspridis:
## Issue

Closes #480  

## Description

Function run_app() now has an argument called datapath that can be used if the user wants the dataset to be pre-loaded when the app launches.
The user needs to call aNCA::run_app(datapath = "path_to_dir/adpc.csv"). The function sets this path as an internal option and attemps to run the app having automatically loaded this data. If no argument is specified, the app will launch as before, without any pre-loaded data.

## Definition of Done
- [x] Make sure dataset file = "adpc.csv" is already loaded in the app once you open the app.
- [x] Different format checks also need to be done
- [x] Make sure it works for any path, and that the wd can be changed for ease
- [x] file argument is NULL by default, if NULL, nothing happens
- [x] Make sure that a data file uploaded in the app overwrites a datapath specified by the user in run_app().
- [x] New logic is documented

## How to test
- Make sure paths from outside and inside the working directory can be used and check that the data is loaded correctly. 
-  It can also be checked that everything works fine when no argument is passed and that a dataset can still be uploaded from inside the app in that case.
- Check that if a data path has been specified in the console, but the user decides to change their mind and upload a dataset through the widget, this dataset gets "priority" and is correctly uploaded.

## Notes to reviewer
Path passed as an argument to run_app() has to be an absolute path, I could not get it to work otherwise.


m-kolomanski (changes requested):
Hey,

I would advise against using global env and setting any variables there, especially since the application / package is designed for explorative and interactive work.

With this implementation, the variable will persist outside of shiny session:
```R
r$> devtools::load_all(); aNCA::run_app(datapath = "test.csv")
ℹ Loading aNCA

Listening on http://127.0.0.1:4385
Browsing http://127.0.0.1:4385
INFO [2025-06-27 17:59:04] Startup
TRACE [2025-06-27 17:59:04] slope_selector: Attaching server
TRACE [2025-06-27 17:59:04] tlg-tlg: Attaching server.
INFO [2025-06-27 17:59:04] Data upload module initialized with datapath: test.csv
ERROR [2025-06-27 17:59:04] Error loading user data: File does not exist: test.csv
TRACE [2025-06-27 17:59:04] Rendering TLG table.

^C # CTRL + C to exit the app

r$> datapath
[1] "test.csv"
```

I think a good practice here would be to use `options()`. That was intended to handle configuration values that are supposed to persist in one session only:

```R
run_app <- function(datapath = NULL, ...) {
  if (!is.null(datapath)) {
    opt <- options(aNCA.datapath = datapath)
    on.exit(options(opt), add = TRUE)
  }

  check_app_dependencies()
  shiny::runApp(system.file("shiny", package = "aNCA"), ...)
}
```

Then, in the application, you can do:
```R
datapath <- getOption("aNCA.datapath", NULL)
```

and proceed from there.

R/run_app.R
```diff
@@ -1,7 +1,13 @@
 #' Run the Shiny app
+#' @param datapath Path to the directory containing the data files.
 #' @param ... Arguments passed to `shiny::runApp()`
 #' @export
```
js3110, line 4: Would be good to add a more detailed description about what happens in the different scenarios eg no datapath, datapaht specified, or datapathspecified but data uploaded using widget.

Also good to include an example with  `@examples` and` \dontrun` and a fake datapath just to help show how it should look!

Issue #449, opened 2025-06-02
Enhancement: QC plots for dose events


Gero1999:
### Description
We would like that at some point at the review section we could incorporate QC plots on the dose events, preferably including both actual and nominal times. The purpose of this is to allow the user review if their data seems to be what they expected.

### Definition of Done
For now we don't need to implement the UI, just create the function that makes this:

- [ ] Integrate a server function that can be used with a `PKNCAdata` object
- [ ] Plot should represent:
- Vertical axis: Each patients, seggregated by cohorts (PKNCAdata$columns$groups)
- Horizontal axis: Time since first dose (AFRLT)
- Cells: Colour bars, points or whatever you find easier for reading the plot
- Optional: Add lines or similar to indicate nominal dose times (NFRLT)

For axis labels: /_Projects/Training_5152/CPDS/NCA/aNCA_Clinical/Pillar3/functions/g_options.R and see filter_breaks.R function

PR #501, opened 2025-07-01
Enhancement: Add Dosing Events to Individual Plots

### LINKED ISSUE #500: Enhancement: add dosing events to individual plots
### Description
Add a vertical dashed line to represent the dosing events for individual plots, so the user can see which points are within which dosing event.
Use `TIME_DOSE `created in `format_data.R`

### Definition of Done
- [ ] Option to add dosing events to ind plot
- [ ] Vertical lines displaying time of dose

js3110:
## Issue

Closes #500 

## Description

Add a vertical dashed line to represent the dosing events for individual plots, so the user can see which points are within which dosing event.
Use `TIME_DOSE `created in `format_data.R`

### Definition of Done
- [x] Option to add dosing events to ind plot
- [x] Vertical lines displaying time of dose

## How to test

Go to Visualisation section -> individual plots -> select add dosing times

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

I set it to only add the lines if there are < 5 subjects on display, otherwise it becomes quite chaotic. Lmk if you think theres further ways to improve this!


js3110:
@Shaakon35 I had a go with the colours but its difficult to do because if you set the colours for the vline then it overrides the colours of the line plot and then theyre all grey. 
I have instead just set it to green.

IMO its not really necessary to see different colours because thats not the purpose of this plot- they will have the dose plot #449 if they want to look at dosing, and can also colour the lines by dose if they want to see that.
but lmk what you think and you can have a go to do it yourself if you want !

m-kolomanski (approved):
Looks great!

The only nitpick that I would have is that the green vline blurs with the blue line of the data and is barely visible. I would probably change it to gray, unless the current look is intended.

Screenshot from 2025-07-10 17-03-04

PR #521, opened 2025-07-14
390 enhancement/ux refine data table layouts

Gero1999:
## Issue

Closes #

## Description
Data tables not always easy to view, especially when using small screen. Want to ensure the following:

Data table does not take up more than the full page
Scrolling horizontally and vertically can be done in one page
Text is a bit smaller so its easier to see
No wrap on text in rows- and rows fully expanded if possible
Possibility to select 'all" for page pagination options, so all data can be seen on the same page

## Definition of Done
- [ ] All data tables formatted the same
- [ ] Text small and easy to read, with limited scrolling
- [ ] Information all present

## How to test
Check all tables through the App (datatables and/or reactables) to seee that they have
- pagination for all rows possibility
- smaller text 
- more compact columns to see more columns than before

## Contributor checklist
- [ ] New logic is documented

## Notes to reviewer

### DIFF
--- a/inst/shiny/modules/tab_data.R
+++ b/inst/shiny/modules/tab_data.R
@@ -165,10 +165,13 @@ tab_data_server <- function(id) {
         highlight = TRUE,
         wrap = TRUE,
         resizable = TRUE,
-        defaultPageSize = 25,
+        defaultPageSize = 10,
         showPageSizeOptions = TRUE,
+        pageSizeOptions = c(10, 25, 50, 100, nrow(processed_data())),
         striped = TRUE,
         bordered = TRUE,
+        compact = TRUE,
+        style = list(fontSize = "0.75em"),
         height = "98vh"
       )
     })
--- a/inst/shiny/modules/tab_data/data_mapping.R
+++ b/inst/shiny/modules/tab_data/data_mapping.R
@@ -337,7 +337,11 @@ data_mapping_server <- function(id, adnca_data) {
         onClick = "select",
         compact = TRUE,
         wrap = FALSE,
-        resizable = TRUE
+        resizable = TRUE,
+        showPageSizeOptions = TRUE,
+        pageSizeOptions = c(10, 25, 50, 100, nrow(df_duplicates())),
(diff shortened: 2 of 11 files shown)


Gero1999:
@m-kolomanski This is a very good call, all tables could be changed except the `units_table` which should stay with DT for the factorial filtering

However, I am afraid this standardization won't be so straightforward. I think we can separate this in two new issues:

1) Create a function to allow in datatables `Excel`, `CSV` and perhaps `XPT` download buttons (with a nice style)
2) Substitute all `DT` tables with `Reactable` tables. Make a common table standard for all (as a function or CSS), so the style can be easily replicated for any new created table

Let me know if this sounds good to you, that way we can also diversify the work. Otherwise, if you think is actually not needed to split I can redefine the issue of this PR and do it here

Shaakon35:
In Data, I still need to slide down to get the page 1 to 216, can we manage to to be able to not have the slider on the right?

Also, the table is too small, make it a bit bigger?

image

When this is done:
Apply the same as in the upload section in:
> Filtering
> NCA > Setup > Summary

Gero1999:
@Shaakon35 I made them smaller, lmk if that fits well

Gero1999:
sorry I think I misunderstood, the problem was with needing to use the scroller because there was a whitespace between the default table number of rows (10) and the window size... So I just reduced the table sizes so they will exactly fit with the rows

I don't see another way to do it unless we modify the whole other parts of the UX so the table can take over more place!

inst/shiny/modules/tab_nca/setup/summary.R
```diff
@@ -48,7 +48,7 @@ summary_server <- function(id, processed_pknca_data) {
         showPageSizeOptions = TRUE,
         striped = TRUE,
         bordered = TRUE,
-        height = "98vh"
+        height = "68vh"
```
Shaakon35, line 51: @Gero1999 Why 68 but not 48vh?

m-kolomanski (approved):

PR #522, opened 2025-07-14
Enhancement: faceting and grouping options for individual plots

### LINKED ISSUE #519: Enhancement: Faceting options for individual plots
### Description
Option to facet_wrap by chosen variables for the individual plots

### Definition of Done
- [ ] Picker input for faceting options
- [ ] Option to facet by > 1 variable
- [ ] Plots displayed in grid

js3110:
## Issue

Closes #519 

## Description

Add faceting and additional grouping options for individual plots.

I have also introduce a helper function `get_persistent_palette` which creates a colour palette based on the grouping variables selected, and then sets the colours so that they do not change every time the user changes the filters.

This update also ensures that USUBJID is always included in the grouping (since theyre individual plots)

New features:
- Grouping for all variables optional
-  Faceting optional
- Faceting filters the dose lines
- Specified colour palette and theme options.

## Definition of Done

- [x] Faceting options
- [x] Additional grouping options

## How to test

Visualisation section individual plots, have a go with different combinations!

## Contributor checklist
- [x] New logic is documented


inst/shiny/functions/get_persistent_palette.R
```diff
@@ -0,0 +1,38 @@
...
+    return(NULL)
+  }
+
+  # Create a temporary column in the data representing the interaction of the variables
+  interaction_data <- data %>%
+    mutate(interaction_col = interaction(!!!syms(colorby_vars), sep = ", "))
+
+  # Get all unique levels of this new interaction column
+  all_levels <- sort(unique(na.omit(interaction_data$interaction_col)))
+  n <- length(all_levels)
```
m-kolomanski, line 19: suggestion: Could be solved with a single pipe
```suggestion
  # Create a temporary column in the data representing the interaction of the variables
  # and get all unique levels
  all_levels <- data %>%
    mutate(interaction_col = interaction(!!!syms(colorby_vars), sep = ", ")) %>%
    pull(interaction_col) %>%
    na.omit() %>%
    unique() %>%
    sort()
  n <- length(all_levels)
```

inst/shiny/functions/get_persistent_palette.R
```diff
@@ -0,0 +1,38 @@
...
+      grDevices::colorRampPalette(RColorBrewer::brewer.pal(11, "Spectral"))(n)
+    } else {
+      RColorBrewer::brewer.pal(max(n, 3), "Spectral")[1:n]
+    }
+  } else {
+    scales::hue_pal()(n)
+  }
+
+  # Return the final named palette
+  setNames(colors, all_levels)
```
m-kolomanski, line 37: suggestion: Looks like a `switch` type of case
```suggestion
   switch(
    palette_name,
    "default" = scales::hue_pal()(n),
    "viridis" = viridisLite::viridis(n),
    "spectral" = if (n > 11) {
      grDevices::colorRampPalette(RColorBrewer::brewer.pal(11, "Spectral"))(n)
    } else {
      RColorBrewer::brewer.pal(max(n, 3), "Spectral")[1:n]
    },
    scales::hue_pal()(n)  # Fallback to default hue palette
  ) %>%
    setNames(all_levels)
```

inst/shiny/modules/tab_visuals.R
```diff
@@ -297,30 +336,41 @@ tab_visuals_server <- function(id, data, grouping_vars, res_nca) {
...
+        p <- p %>%
+          layout(
+            xaxis = list(
+              rangeslider = list(type = "time")
+            )
           )
-        )
-
+      }
+      return(p)
```
m-kolomanski, line 373: nitpick:
```suggestion
      p
```

m-kolomanski (commented):
discussion:
Hey, I must say I am not a fan of the solution for consistent color palettes. I get where that is coming from, but with large amount of unique values and colors being generated sequentially, this means that neighboring values will get very similar, possibly indistinguishable.

I personally would prefer to have inconsistent colors between selections, but distinct enough that I can differentiate between close values. But I am not an end-user of the app, so you tell me what are the expectations, I just wanna raise a point that maybe we could figure out something better.

One trick that we could do to try and achieve the best of both words is to use `sample()` in place of `sort()` when creating the `all_levels` variable. Then, the values will get _scrambled_, so neighboring ones will have a greater change to get assigned colors that are more spread out.

<details>
<summary>Screenshot with `sort()`</summary>
Screenshot from 2025-07-14 18-31-30
</details>

<details>
<summary>Screenshot with `sample()`</summary>
Screenshot from 2025-07-14 18-38-05
</details>

The flaw is that with this approach, the colors will be _semi-randomized_ again every time the user e.g. switches between palettes. But there is a way around it, we could run the palette only once for given data and colorvar and then _cache_ it within the application, so that unless the user reloads the data or the entire app, the colors will stay consistent. Or there might be completely different approach that I am not thinking of that could be even better.

Let me know what you think and if this is something worth spending a little more time on - the current solution might be enough, LMK.

Shaakon35 (approved):
LGTM

js3110:
@m-kolomanski thanks for the feedback, I've gone with sample as I think this fits our needs. It is fine if the palette changes when you change theme, the important thing is that the colours stay the same if the filters change (e.g you add more subjects or remove some) Thats all the consistency required!
PTAL :)

PR #530, opened 2025-07-15
Enhancement/qc plots for dose events

### LINKED ISSUE #449: Enhancement: QC plots for dose events

lucyaspridis:
## Issue

Closes #449 

## Description

2 functions were created. The first one, `faceted_qc_plot`, generates a ggplot faceted by the groups specified in the `grouping_vars` argument. The x variable, the y variable and a variable to colour the points by, as well as a title for the plot, are to be specified, so that the plot can be used in other settings if needed. (PK sample plot for instance). 
A labels data frame can be specified if the plot is to be passed to `ggplotly`, for the labels to be used on the tooltip text, instead of the variable names. This tooltip text is created in the second function, `generate_tooltip_text`, in `label_operators.R`.
I also modified the `get_label` function so that if `labels_df` did not contain a label for the variable passed as an argument, the variable name would be returned.

The plot is not implemented in the app yet, this will be done in a separate issue. The idea is to have the user select the grouping variables from a selectize widget. 

## Definition of Done
- [x] Integrate a server function that can be used with a PKNCAdata object
- [x] Make a ggplot object that can be passed to ggplotly
- [x] Have the labels be used instead of the variable names on the tooltip texts

## Contributor checklist
- [x] New logic is documented

## How to test

To use the plot as a dose QC plot, here is an example with the Dummy dataset.
```
df = read.csv("inst/shiny/data/Dummy_complex_data.csv")
p = faceted_qc_plot(data = df,
                x_var = "AFRLT",
                y_var = "USUBJID",
                colour_var = "DOSEA",
                grouping_vars = c("TRT01A", "SEX"),
                labels_df = LABELS,
                title = "Dose Plot") 

p %>% ggplotly(tooltip = "text") %>% 
  layout(title = list(text = paste0(p$labels$title, "<br><sup>", p$labels$subtitle, "</sup>")))
```

## Notes to reviewer

- Should there be an option to set `colour_var ` to `NULL`, i.e., not colour the plot?
- I did not set the x axis scale to only display the nominal times in the data, as Valentin had suggested, because it was tedious. Feel free to either make it a separate issue if you think it is needed, or suggest a way to have nominal times appear on the plot somehow.
- I wasn't sure which unit tests to set up, feel free to suggest other ones if needed


R/faceted_qc_plot.R
```diff
@@ -0,0 +1,90 @@
...
+#' faceted_qc_plot(
+#'   data = qc_data,
+#'   x_var = "TIME",
+#'   y_var = "RESULT",
+#'   colour_var = "DOSE",
+#'   grouping_vars = "ARM",
+#'   labels_df = label_data,
+#'   title = "Sample QC Plot"
+#'   )
+#' }
```
m-kolomanski, line 45: issue: There is no reason to wrap this example in `\dontrun{}`, especially that you took extra effort to setup dummy functions
issue: but I do not see the reason why the dummy functions would be needed here, the examples seem to run fine with them removed?
```suggestion
#' @examples
#' # Sample data
#' qc_data <- data.frame(
#'   TIME = 1:6,
#'   RESULT = c(5, 6, 8, 9, 12, 11),
#'   DOSE = as.factor(c(10, 10, 20, 20, 30, 30)),
#'   ARM = rep(c("A", "B", "C"), each = 2)
#'   )
#' label_data <- data.frame() # Dummy labels object
#'
#' # Generate the plot
#' faceted_qc_plot(
#'   data = qc_data,
#'   x_var = "TIME",
#'   y_var = "RESULT",
#'   colour_var = "DOSE",
#'   grouping_vars = "ARM",
#'   labels_df = label_data,
#'   title = "Sample QC Plot"
#'   )
```

R/faceted_qc_plot.R
```diff
@@ -0,0 +1,90 @@
...
+#'
+#' @export
+faceted_qc_plot <- function(data,
+                            x_var,
+                            y_var,
+                            colour_var,
+                            grouping_vars,
+                            labels_df,
```
m-kolomanski, line 53: suggestion: `labels_df` could have  default value in case the user does not want to provide labels and wants to rely on variable names.
```suggestion
                            labels_df = data.frame(),
```

m-kolomanski (changes requested):
Hey, code quality is good and works well, I have some issues with the documentation and the tests.

m-kolomanski (approved):
Looks great, awesome job 🚀

Issue #446, opened 2025-05-30
Bug: link units_table modifications with PKNCAdata & PKNCAresults


Gero1999:
### Description
Currently the App will build correctly a default units table and allow to the user to modify the `final_results` through the `session$userData$units_table()`. However, this won´t affect the original PKNCAdata$units nor will affect the `res_nca` (hence the `Parameter Datasets`).

**Potential solution:** What we need is to integrate either `session$userData$units_table()` or directly the `modal_units_table` from the module to `PKNCA_update_object`

### To reproduce
Map columns > NCA setup > Units Table > Do modifications (i.e TMAX, Custom Unit: ms) > Check `Parameter Datasets`, TMAX is still in `h` (default)

### Expected behaviour
Customizations in the units table should replicate in ALL results, at the bare minimum after `Run NCA` is pressed:
- [ ] Pivoted results show the updated units
- [ ] Parameter Datasets show the updated units

> Note: The preferred solution will always be the one most ´PKNCA´ integrated. This issue is blocked until merge of #416

PR #551, opened 2025-07-22
fix: problems in summary table with get_conversion_factor  & link better units table (#446, #549)

### LINKED ISSUE #549: Bug: get_conversion_factor error in if (initial_unit == target_unit)
### Description
A change at some previous merge in how the summary table is generated is producing a problem due to the assumptions of `get_conversion_factor`.  This is supposed to be an easy fix by just reducing the strength of its assumptions.

``` 
Warning: Error in mutate: ℹ In argument: conv_factor = get_conversion_factor(PPORRESU, PPSTRESU).
Caused by error in if (initial_unit == target_unit) ...:
! missing value where TRUE/FALSE needed
  2: shiny::runApp
  1: run_app [C:/Users/spinnerj/Documents/aNCA/R/run_app.R#37]
```

### To reproduce
Study with multiple analytes >> Change unit manually (i.e, clearance) >> run NCA

### Expected behaviour
No error & obtention of results

Gero1999:
## Issue

Closes #549 and hopefully also #446

## Description
- Fixed function `get_conversion_factor` to better accept NA inptus
- PKNCA results (`res_nca`) will consider changes in the units table before running NCA (pk.nca) instead of post-processing
- Summary statistics table will take the Mode unit and calculate the conversion factor, instead of taking the mode of the conversion factor (which potentially can be wrong and becomes messy with NAs!)

## Definition of Done
When making changes in the units table:
- [x] Pivoted results and also Parameter Datasets show the updated units
- [x] Running NCA produces no error and generates the summary table

## How to test
Study with multiple analytes >> Change unit manually for 1 analyte (i.e, clearance) >> run NCA >> Check all results

## Contributor checklist
- [x] New logic is documented


js3110 (approved):
LGTM! thanks

m-kolomanski (approved):

js3110 (approved):
lgtm

PR #556, opened 2025-07-23
Enhancement: settings upload to include units

### LINKED ISSUE #472: Enhancement: Settings upload to include units
### Description
When user uploads settings, as well as updating general settings it should also update any manual unit conversions that were made. This means the units need to be exported in settings download, and updated with settings upload.

### Definition of Done
- [ ] unit conversions exported with download of settings (rds ?)
- [ ] unit conversions added to table with settings upload
- [ ] unit conversions applied to data after settings upload.

js3110:
## Issue

Closes #472 

## Description

Updated the settings object to include the units table, so that when its uploaded the units are automatically updated as well.

## Definition of Done

- [x] Settings upload works and includes units

## How to test

Go through normal NCA workflow -> change units and other settings -> save settings
Refresh page -> upload settings -> see if unit changes are uploaded

## Contributor checklist
- [x] New logic is documented

## Notes for the Reviewer

I have also updated the pickerinputs for dosno and pcspec so that the settings correctly update them.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.0.0.9123
+Version: 0.0.0.9124
 Authors@R: c(
     person("Ercan", "Suekuer", , "ercan.suekuer@roche.com", role = c("aut", "cre"),
            comment = c(ORCID = "0009-0001-1626-1526")),


m-kolomanski (commented):
question: Do we want to attach the units specification of the user leaves it at default? Currently, `session$userData$units_table()` object is `NULL` untill user changes and saves units. Can there be a situation, where user runs an analysis without ever opening the units table, but then loads some data with somehow different units, then loads the settings and wants to have the same units as with the first analysis run?

Could this be an issue and lead to some unexpected behavior? Or are we safe with this?

js3110:
> question: Do we want to attach the units specification of the user leaves it at default? Currently, `session$userData$units_table()` object is `NULL` untill user changes and saves units. Can there be a situation, where user runs an analysis without ever opening the units table, but then loads some data with somehow different units, then loads the settings and wants to have the same units as with the first analysis run?

@m-kolomanski this is a good point, to be honest I'm don't think so, but I'm not 100% sure. If we automatically save then yes as you said the user can upload and have the units back to the original run. But honestly I think its highly unlikely that they will upload the same data with different units.  If anything I would say its more likely that, in the rare case the same data has different units, they would want to keep those units rather than the former ones. Because if the units of the data have changed its likely for a reason.

In terms of unexpected errors, I can't think but maybe you have a suggestion of what may occur?

Maybe @Shaakon35 and @Gero1999 can weigh in on what they think is best: only save units if they were changed, or always save units?

m-kolomanski:
@js3110 Not really errors, but it is more about being intuitive to users.

Because sometimes we will be saving units along with the settings, and sometimes not.

The user might expect units to be overwritten by imported settings, but they will not (because they were not included in the settings export).

Would **always** overwritting units be annoying to the user? At least it would lead to consistent behavior. But it is possible that we are discussing unrealistic or very rare cases that will not make a real difference to the UX.

Shaakon35:
> question: Do we want to attach the units specification of the user leaves it at default? Currently, `session$userData$units_table()` object is `NULL` untill user changes and saves units. Can there be a situation, where user runs an analysis without ever opening the units table, but then loads some data with somehow different units, then loads the settings and wants to have the same units as with the first analysis run?
> 
> Could this be an issue and lead to some unexpected behavior? Or are we safe with this?

I don't think this scenario can or will happen, so if it's too complex to code it, you can forget about this scenario.

m-kolomanski (approved):
LGTM!

lucyaspridis (approved):
works perfectly!

PR #562, opened 2025-07-23
decimal fix

### LINKED ISSUE #561: Bug: PPSTRESN != PPSTRESC
### Description
PPSTRESN != PPSTRESC

Shaakon35:
## Issue

Closes #561 

## Definition of Done

Always
PPORRES = PPSTRESC =PPSTRESN  

## How to test

Test with dummy data

## Contributor checklist
- [ ] New logic is documented

### DIFF
--- a/R/export_cdisc.R
+++ b/R/export_cdisc.R
@@ -146,7 +146,6 @@ export_cdisc <- function(res_nca) {
         TRUE ~ PPTESTCD
       )
     ) %>%
-
     ungroup() %>%
     #  Recode PPTESTCD PKNCA names to CDISC abbreviations
     add_derived_pp_vars(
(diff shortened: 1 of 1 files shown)


Shaakon35:
@Gero1999 Basically what was going wrong was that you were reformatting PPSTRESN based on the specifications, where lenght=8.

You could not know, but length = 8 is something that is coming from SAS, In SAS, you can define a maximum length of the numeric value so that the software only shows a certain number of characters (like a View in R).
In SAS, the format for the numeric is most of the time best12. which means that you want to round to 12 digits, and that's what e already did in add_derived_pp_vars

Now, you can try to submit PP in Pinnacle 21 and it should work!

Gero1999 (changes requested):
I still need to test in P21, but from the code I am afraid that if it passes it could just be a solutions for datasets like the dummy one, that have a certain number of decimals. I tried to point out potential problems I found within the code.

Shaakon35:
Current branch: https://github.com/pharmaverse/aNCA/issues/561

PR #563, opened 2025-07-24
enhancement: Refine ratios table handling

### LINKED ISSUE #538: Enhancement: Refine parameter ratios options table
### Description
There are some ideas that might help improving the interface and preventing confusing situations for the user.

### Definition of Done
- [x] Clarify instructions for `Other levels` adding an example
- [x] Create a warning message (or use the ones with no-groups) if a dataset has different analytes and differents drugs and thus is not comparable...
- [ ] Introduce a delay for updating the PPTESTCD when done so it does not refresh the reactable for every letter we type
- [x] Add PPANMETH to the output to indicate reference
- [x] Ratio parameters should be included in `Parameter Datasets`
- [x] Differentiate `Parameter_Test` and `Parameter_Reference` so operations can be done with different parameters within the same profile

Gero1999:
## Issue

Closes #538 

## Description
Implemented some ideas that improve the interface and also expands the functionality (allowing now operations within the same profile). 

Change description.
## Definition of Done
- [x] Clarify instructions for `Other levels` adding an example
- [x] Create a warning message (or use the ones with no-groups) if a dataset has different analytes and differents drugs and thus is not comparable...
- [x] ~Introduce a delay for updating the PPTESTCD when done so it does not refresh the reactable for every letter we type~
- [x] Add column to the output to indicate reference (PPANMETH)
- [x] Ratio parameters should be included in `Parameter Datasets`
- [x] Differentiate `Parameter_Test` and `Parameter_Reference` so operations can be done with different parameters within the same profile

## How to test
Map NCA > Setup > Section `Parameter Ratios`

## Contributor checklist
- [x] New logic is documented


js3110 (commented):
@Gero1999 I am getting different results when I add ratios vs not:
eg For Analyte01 and Metabolite, doses 1 and 2 Serum, if I add a ratio all of the metabolite parameters are set to missing, but usually there are normal results

js3110 (approved):
LGTM! nice one 🚀

PR #565, opened 2025-07-24
564 enhancement/unify variable labelling

### LINKED ISSUE #564: Enhancement: Unify label addition of variables in the App
### Description
Now that we have a good object with all information for NCA variables `metadata_nca_variables` we can consistently use it through the App for all needed labeling. In combination with `translate_terms` it will standardize the way we use metadata

### Definition of Done
- [ ] labels are properly called 
- [ ] All recognised CDISC variables within the App are labelled with tooltips in the corresponding dataframes

Gero1999:
## Issue

Closes #564 

## Description
Now that we have a good object with all information for NCA variables `metadata_nca_variables` we can consistently use it through the App for all needed labeling. In combination with `translate_terms` it will standardize the way we use metadata

## Definition of Done
- [x] We unify the use of `metadata_nca_variables` instead of `LABELS`
- [x] labels are properly defined for CDISC variables
- [x] All reactable tables, when having CDISC columns, show labels

## How to test
Check the labels and compare `main` with this branch 

## Contributor checklist
- [x] New logic is documented


inst/shiny/modules/tab_data/data_mapping.R
```diff
@@ -216,7 +216,9 @@ data_mapping_server <- function(id, adnca_data) {
 
     # Loop through each label and create the renderText outputs
     purrr::walk(MAPPING_DESIRED_ORDER, \(label) {
-      output[[paste0("label_", label)]] <- renderText(get_label(LABELS, label, "ADPC"))
+      output[[paste0("label_", label)]] <- renderText(
+        get_label(labels_df = metadata_nca_variables, label, "ADPC")
```
m-kolomanski, line 220: nitpick: No need to provide the default variable. Also, if function has argument names specified, please do try to keep the order the same as in the function definition (`labels_df` is the third argument, here you provide it as first).
```suggestion
        get_label(label, "ADPC")
```

inst/shiny/modules/tab_nca/excretion_analysis.R
```diff
@@ -186,6 +186,7 @@ excretion_server <- function(id, input_pknca_data) {
     reactable_server(
       "results_table",
       results_output,
+      columns = generate_col_defs,
```
m-kolomanski, line 189: suggestion: If we are adding column definitions to all reactables, how about we add it to the defaults in the `reactable_server` module itself?

m-kolomanski (changes requested):
Good change in general, I have some suggestions though

R/label_operators.R
```diff
@@ -22,111 +21,26 @@
...
-    if (is.factor(data[[col]])) {
-      data[[col]] <- as_factor_preserve_label(data[[col]])
-    }
-  }
+apply_labels <- function(data, labels_df = metadata_nca_variables, type = "ADPC") {
+  formatters::var_labels(data) <- ifelse(
+    is.na(formatters::var_labels(data)),
+    get_label(names(data), type = type, labels_df = labels_df),
+    formatters::var_labels(data)
+  )
```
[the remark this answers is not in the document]
Gero1999: @m-kolomanski this is actually with the intention of only applying labels if they are not already provided (so if the user has some original ones we don't change them with our own). Maybe I am missing something but I would say this is correct

PR #580, opened 2025-07-30
Enhancement: refresh when restart button clicked

### LINKED ISSUE #569: Enhancement: restart button should refresh the app
### Description
Restart button of data tab currently just sends you back to the first page of data tab. It would be better if the restart button actually just refreshes the page so that everything starts from scratch. Otherwise sometimes there are crashes in the app if you upload new data after already submitting mapping on a dataset.

### Definition of Done
- [ ] Restart button refreshes app
- [ ] Everything else works fine

js3110:
## Issue

Closes #569 

## Description

When restart button is clicked in the data tab workflow, rather than just taking you back to page 1 it refreshes the app for a real restart. This will prevent crashes caused by uploading new data after mapping previously.

## Definition of Done

- [ ] Restart button refreshes the app

## How to test
 
Run through various steps of data tab workflow and click restart.

## Contributor checklist
- [ ] New logic is documented

### DIFF
--- a/inst/shiny/modules/tab_data.R
+++ b/inst/shiny/modules/tab_data.R
@@ -90,8 +90,7 @@ tab_data_server <- function(id) {
     })
 
     observeEvent(input$restart, {
-      data_step(steps[1])
-      updateTabsetPanel(session, "data_navset", selected = step_labels[1])
+      shinyjs::refresh()
     })
 
     observeEvent(input$next_step, {


m-kolomanski:
I don't know. If we have this approach, I do not know why we need a restart button at all then, if we can just refresh the page.

One case where this might save two clicks is when user runs the app locally within their IDE window - then you might not have a refresh button (I don't, in VSCode) and you have to close the window and call `run_app()`. This takes a (literal) second, maybe 1.5, so is barely an inconvenience, especially that I am usually runnning with `shiny::devmode(T)`, which does that automatically for me.

But then, in the past, I have had users request the app to have a bespoke **Log out** button, even though what it did was just to call `session$reload()` and with the deployment setup pressing the reload button in the browser (or F5) would do exactly the same. Still, they wanted a *semantic* button for performing specific action, and the mechanics did not matter. So maybe there is an UX issue we are solving, when the user does not feel like they are *hacking the system* or *going the long route* by refreshing the page I guess, and instead restart the app *the intended way*.

As a developer, I dont really care, we probably shouldn't dwell on this very long, pick an option, do it quickly and forget about the case.

js3110:
> As a developer, I dont really care, we probably shouldn't dwell on this very long, pick an option, do it quickly and forget about the case.

ok, imo its fine like this, and we haven't had any requests to restart in general, it just works nicely with the workflow of the data tab. So lets leave it how it is and we can come back to it if the Users have any requests.

PR #586, opened 2025-08-05
Nitpick: lint all files

### LINKED ISSUE #585: Nitpick: Lint
There are some lint warnings that come up but not official lint fails so I would like to fix them

js3110:
## Issue

Closes #585 

## Description

Lint returned 20 warnings, mostly just because of `incomplete final line` for functions and files. I have added these so lint runs perfectly now with no warnings.

## Definition of Done

- [x] No lints found

## How to test

`devtools::lint`

## Contributor checklist
- [x] New logic is documented

This was just bugging me so I wanted to fix it :)


Shaakon35 (approved):

m-kolomanski (approved):
suggestion: Can you also remove the following line from `.lintr` file:
```R
trailing_blank_lines_linter = NULL
```

That is the reason these issues are thrown as warnings as opposed to errors, and why our CI passes even with those.

I presume this is a legacy thing that was set up when `lintr` was first set up for the project, I dont remember anymore ;p

Nevertheless, if the files are cleaned with this PR, we can switch the lint on and make sure it pops up on the CI in the future.

PR #591, opened 2025-08-14
length 12 for PPSTRESN

### LINKED ISSUE #561: Bug: PPSTRESN != PPSTRESC

Shaakon35:
## Issue

Closes #561 

## How to test

Now, you can try to submit PP in Pinnacle 21 and it should work!

## Contributor checklist
- [ ] New logic is documented

### DIFF
--- a/R/data.R
+++ b/R/data.R
@@ -51,6 +51,7 @@
 #'   \item{is.used}{Logical. TRUE if the variable is meant to be included.}
 #'   \item{Values}{Character. Possible values for the variable (if applicable).}
 #'   \item{is.mapped}{Logical. TRUE if the variable is mapped in ADNCA (App's input).}
+#'   \item{mapping_tooltip}{Character. A descriptive tooltip for the mapping tab.}
 #' }
 #' @source Used for PP and ADPP mapping rules and checks in the export_cdisc function
 "metadata_nca_variables"
--- a/inst/WORDLIST
+++ b/inst/WORDLIST
@@ -14,6 +14,7 @@ DOSEA
 DOSEU
 DOSNOA
 EVID
+Extravascular
 Kezia
 Kobana
 LAMZ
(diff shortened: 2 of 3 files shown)


R/export_cdisc.R
```diff
@@ -401,7 +401,7 @@ add_derived_pp_vars <- function(df, conc_group_sp_cols, conc_timeu_col, dose_tim
       # Parameter Variables
       PPORRES = as.character(round(as.numeric(PPORRES), 12)),
       PPSTRESN = round(as.numeric(PPSTRES), 12),
-      PPSTRESC = as.character(format(PPSTRESN, scientific = FALSE, trim = TRUE)),
```
Gero1999, line 404: Consider not doing this

issues:
- PPSTRESN does not still match PPSTRESC in the conflictive row (`XX01-25103, AUCPEP, Analyte01`)
- doing this there is a problem with the scientific nomenclature if you do not force it to FALSE...

R/export_cdisc.R
```diff
@@ -357,7 +357,7 @@ adjust_class_and_length <- function(df, metadata) {
       df[[var]] <- substr(as.character(df[[var]]), 0, var_specs$Length)
     } else if (var_specs$Type %in% c("Num", "integer", "float") &&
                  !endsWith(var, "DTM")) {
-      df[[var]] <- round(as.numeric(df[[var]]), var_specs$Length)
+      df[[var]] <- round(as.numeric(df[[var]]), 12)
```
Gero1999, line 360: as mentioned I think the problem is the round() operation, not the number itself, no need to do it, it can be the reason of the differences PPSTRESN, PPSTRESC

Gero1999 (changes requested):
@Shaakon35 I run on P21 and found still the issues

Gero1999:
I think that still is not a problem in the XPT submission file, but as this is not priority I will let this PR breathe and think about it

PR #607, opened 2025-08-22
Enhancement: Better loading widget when running nca

### LINKED ISSUE #523: Enhancement: UX loading widget when running NCA
### Description
Improve the loading widget that currently appears when pressing `Run NCA`. Some improvements include:

- Centralize the widget to the middle of the screen, to make more clear to the user what is happening
- Align better the timing of appearance of the widget. Right now the widget seems to disappear a bit before the results have been processed and the user moves to another tab. 

### Definition of Done
- [ ] centralize loading widget (or just make more visible)
- [ ] no significant lag between widget disappearance and loading of results!

m-kolomanski:
## Issue

Closes #523

## Description

- Implements full loading popup with a spinner when running NCA.
- Popup is visible until the results table is ready to be displayed.
- Refactors `stepper_ui` to be reusable.
- ~Implements `stepper_ui` in the NCA tab.~
- ~Removes old `Run NCA` button.~

~When trying to work out this issue with loading notification closing too fast, I noticed the reactivity with changing the tabs works a bit weird. I decided to migrate the stepper input from `data` tab into `nca` tab, since it seems like nice fit for the flow (setup -> results -> analysis). In order to run the NCA, you now need to click the `Next` button in the bottom right. you can then go back and forth between the tabs without recalculating. Recalculation will be triggered when any settings change - then, upon clicking the `Next` button, analysis will be run again.~

Stepper implementation in NCA tab was reverted. Since the job was already done, I am keeping the improvements to `stepper_ui` - might be useful in the future.

## Definition of Done

- [x] centralize loading widget (or just make more visible)
- [x] no significant lag between widget disappearance and loading of results!

## How to test

Run the application, go `Next` on default data until the preview. Go to `NCA` tab. Click `Next`. Check if the modal is closed at the right time (after results are displayed). Go back and forth between the tabs. Try changing some settings in order to trigger recalculation.

## Contributor checklist
- [x] New logic is documented

### DIFF
--- a/inst/shiny/modules/tab_nca.R
+++ b/inst/shiny/modules/tab_nca.R
@@ -205,7 +205,7 @@ tab_nca_server <- function(id, adnca_data, grouping_vars) {
           showNotification(w_message, type = "warning", duration = 5)
         })
 
-        updateTabsetPanel(session, "ncapanel", selected = "Results")
+        updateTabsetPanel(session, "nca_navset", selected = "Results")
 
         log_success("NCA results calculated.")
 


js3110 (commented):
Hi @m-kolomanski thank you for taking this issue, it looks pretty good. However I have a couple of concerns with this reformatted workflow when it comes to user experience:

1. Additional Analysis is not necessarily step 3 in the process, its kinda isolated as a step. You could do it before running the NCA if you wanted as it doesn't require the results. (eg some people may only want excretion or ratios and not actualy NCA). Currently with the stepper UI its impossible to get to that page without running NCA.

2. The 'next' aka run NCA button is not available on the slope selector tab (this could be fixed I think since currently its just because its next page for the plots instead). For the user, after selecting the initial settings they are likely to then go to the slope selector page and then run the NCA, so it would be good if it was possible for them to run the NCA on this page.


Point 2 I think could be fixed in this quite easily, but I am not sure the best way to deal with point 1 using this new layout. Maybe you have some ideas!

osenan (changes requested):
I test it and it works as expected.
However, is strange that the next buttons appears or disappear depending on the context. Is not appearing in the exploratory plots, then appears again. We should define if it should appear always or why not. Then implement it. Or then next is only to navigate the four first changes and not more.

Also, the css of the modal I think should be in scss file, more details in the comments. I would try a little bit small style of the font.

m-kolomanski:
@js3110 The only idea I have for the additional analysis step is to move it to **Exploration** tab, but I don't know if that is a good fit either. Otherwise, I can just revert and scrap the stepper, keep it only for the data tab, and just keep the changes for the loading widget in this PR.

inst/shiny/www/styles/partials/_loading_spinner.scss
```diff
@@ -0,0 +1,45 @@
...
+  }
+  40% {
+    box-shadow: 0 -2em;
+    height: 5em;
+  }
+}
+
+h4.modal-title:has(#loading-title) {
+  width: 100%;
+}
```
osenan, line 45: please add a trailing newline at EOF

inst/shiny/www/styles/partials/_loading_spinner.scss
```diff
@@ -0,0 +1,45 @@
+.loading-spinner-container {
+  height: 4em;
+}
+
+#loading-spinner {
+  color: #3f8abf;
+  position: relative;
+  font-size: 0.5em;
+  background: #3f8abf;
```
osenan, line 9: I recommend either trying the `$anca-blue` variable or if you think it is better use this blue please  create a color sass variable in the file [_colors.scss](https://github.com/pharmaverse/aNCA/blob/main/inst/shiny/www/styles/modules/_colors.scss)

osenan (approved):
Good changes regarding the styling! I only have one comment to create a color variable instead of repeating the hex definition.

Gero1999 (approved):
LGTM, great job! And I agree with the comments, they are just WISH issues, so no relevant at all

Issue #606, opened 2025-08-22
Bug: Parameters not selected shown in results if required for calculations


js3110:
### Description
If you only want to see one parameter and so select it, but then that parameter requires other parameters for calculation (e.g `R2adj` requires `halflife` which requires `tmax`and `tlast`) then all parameters will appear in results.

### To reproduce
Deselect all parameters except for R2ADJ, see that there are more parameters shown.

### Expected behaviour
Filter final results so that only the parameters given in parameter selection are visible (can use TRUE intervals in `results$data$intervals` to filter rather than passing on parameter selection from a separate module)

### Definition of Done

- [ ] Results filtered for selected parameters
- [ ] ADPP and PP filtered for selected parameters
- [ ] Exports filtered for selected parameters

Issue #588, opened 2025-08-07
Enhancement: Automatically detect study type(s) in data


js3110:
### Description
Create a select input for study type(s) that is automatically populated based on data. Types of study:
- Intravascular (if ADOSEDUR present and > 0)
- Extravascular
- Single dose (only 1 dose and Tau not mapped)
- Multiple dose (Tau mapped or > 1 dose)
-  Excretion (VOLUME col mapped and has > 0 values)

Could be multiple study types per dataset so if >1 type detected, mutliple are selected.

### Definition of Done
The requirements for the feature to be complete:
- [ ] Select input with study type (multiple = TRUE)
- [ ] Automatically filled with app-detected study types
- [ ] In Data preview section?

Issue #532, opened 2025-07-15
Refactor: Main PKNCA workflow functions


Gero1999:
### Description
The purpose is to simplify as much as possible the main PKNCA workflow by:

- Removing conditionals when not really needed
- Reusing as much as possible PKNCA functions instead of custom ones
- Using PKNCA assumptions 
- Not self-creating column names that are not in the input or PKNCA generated

For that functions in the file `R/format_data.R`, `R/PKNCA.R` & `R/create_start_impute` should be rechecked and simplify as much as possible.

### Definition of Done
The requirements for the feature to be complete:
- [x] `format_pkncaconc_data` `format_pkncadose_data` might be considered to be simplified and even phused with PKNCA.R functions. Try also to make them in better contact with the columns mapped, using that object as an argument
- [ ] `format_pknca_data_intervals` has to be simplified, use if possible more PKNCA functions, prevent creating self-made columns.
- [ ] try only making and using one of `PKNCA_create_data_object`  & `PKNCA_update_data_object`. Most of their actions are redundant

Issue #436, opened 2025-05-27
Enhancement: set default parameter selections based on study type


js3110:
### Description
At the moment, if you select a parameter then it will be calculated, which does not take into account whether its scientifically relevant. 
Create some workflows that ensure the following:

**Excretion**
- For urine and feces PCSPEC, only urine parameters selected.

**Extravascular**
- For extravascular data, no IV parameters should be selected

**Single Dose**
- For single dose studies (or DOSNOA == 1), only single dose parameters should be calculated, so no steady state parameters.
- End should be set to Inf

**Steady State**
- For steady state (assume for DOSNOA > 1), steady state parameters calculated instead of single dose- Clss, Vss

Create a select input for study type(s) that is automatically populated based on data. Types of study:
- Intravascular (if ADOSEDUR present and > 0)
- Extravascular
- Single dose (only 1 dose and Tau not mapped)
- Multiple dose (Tau mapped or > 1 dose)
-  Excretion (VOLUME col mapped and has > 0 values)

Refactor to move parameter selection to Summary tab, and have one column per study type detected, with parameter selections for each column.

Eg:

Study Type Detected: | Type 1 (eg MD Infusion) | Type 2 (eg Excretion)
-- | -- | --
Parameters |   |  
AUCINF | TRUE | FALSE
AUCLAST | TRUE | FALSE
CMAX | TRUE | FALSE
TMAX | TRUE | FALSE
RCAMNT | FALSE | TRUE

### Definition of Done
The requirements for the feature to be complete:
- [x] Checks for different types of studies
- [ ] Select input for study type that user can change?
- [x] Default parameter selections based on study types
- [x] Allow manual editing of these selections
- [x] NCA runs fine with those parameters removed

Gero1999:
For whoever implements this in the future, please do not affect the intervals that produces the results, as discussed in #433. 

This must be more a post-processing action that filters the displayed results for the `NCA results`, ``Parameter Datasets`, `Slope Table` ...

Otherwise, if you are keen in modifying intervals, take into consideration at least the parameters needed for renal calculations and / or parameter ratios (check pr #487). Consider other potential parameter needs as well!

PR #608, opened 2025-08-25
Enhancement: rework parameter selection for multiple study types (593)

### LINKED ISSUE #593: Enhancement: Rework parameter selection for multiple study types
### Description
Currently you can only select each parameter you want, but if you want different parameters for different parts of the study (eg urine for only urine and SD for Plasma), this is not possible. Based on description in #436 and after completion of #588 we would move the parameter selection to the summary tab, and allow for different parameter selections based on different study types detected.

### Definition of Done
- [ ] Parameter selection divided into study types
- [ ] Default selections based on PK parameter table defaults (ask @js3110 for more info) 
- [ ] Ability to select different parameters that will then be calculated for different data.

js3110:
## Issue

Closes #593 as a part of #436 
Merging blocked until #588 merged.

## Description

Parameter selection reworked so that different parameters can be selected for different study types.

Summary of the changes made:
- Parameter selection moved to Summary tab.
- Parameter table updated so there is one column per study type (detected using `detect_study_types` function.
- Remove verify_parameters function (the next step will be to have the default parameters for each study type automatically selected, and then the user can decide to change that (any errors will stop them from running NCA so its fine)
- Separation of setting intervals (start and end times) and then setting parameters, so slopes and main parameters are created separately. (`update_PKNCA_data_object` function is more primed to be updated in #532 with these changes, as it could easily be combined with `create_PKNCA_data_object `imo)

Things that still need doing (in separate issues?)
- Default selection of parameters for the different workflows
- Update settings download and upload
- Ensure only selected parameters are in results table (even if required to calculate params: #606 )
- Add select all buttons for categories in parameter selection table

## Definition of Done

- [x] Parameter selection divided into study types
- [x] ~Default selections based on PK parameter table defaults~ (todo in spearate issue)
- [x] Ability to select different parameters that will then be calculated for different data.

## How to test

Try main dummy data or adnca_sm_dummy_dataset, and try selecting different parameters for different studies (the best ones to test are observed params eg CMAX, CLAST, TMAX as the others have dependencies so will still show results (to be fixed in #606 )

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

This is quite a large PR so I plan to tackle other changes like updating the settings separately. Let me know if you think its worth keeping this branch open and then merging settings and default selections into here before ultimately merging to main, or if you think we should merge to main at each step. I could also create a branch in #436 that we use as the stepping stone before merging to main...


Shaakon35 (commented):
The buton Run NCA is delayed by 5 to 10 sec, when adding a new pk parameter: C0 for single IV dose for example.
Should we open a separate issue to otpimize this?

Shaakon35 (commented):
If I add a new PCSPEC in the setup, the summary of study type is very long to appear, especially with the dummy dataset. We should do something there

Gero1999 (commented):
Hey good job, LRPTM! A few comments to consider that might help to nail it:

### Theorical aspects
1. **Other study types tbc**. Maybe `Parent/Metabolite` (I think it was a distinguishing factor in the guidelines, but double check) and for sure a big issue might be not distinguishing `IV bolus / Infusion`, which will make the calculation wrong for many parameters.
2. **Functionality**. Is it currently working? When I choose R2ADJ to be only in IV single dose, I can still see CSF.
3. **Single Dose Study**. Just looking at the App I feel that `SD: start = 0, end = Inf`. This is correct, but you should double-check is not needed to calculate SD parameters for also `NCA_PROFILE = 1` in MD studies, which for me makes scientific sense.

 ### UX aspects
4. I am not sure the initial table for the study types is useful per ce. I think a scientist would be more relief if they see how each cohort was assigned as
5. I understand for simplification you put it in summary. For me it is still a bit unintuitive to find the feature. I would rename it to `Parameter Selection` or something similar.

### General code view aspects
I did not review the code, just have a general view:
6. `create_ & update_ funs` I am not happy separating these type of funs. Something similar to what I commented in #532. PKNCA has already a good integration for updating the object in an easier way for each specific scenario, rather than re-making everything every time one particular thing is changed reactively.

Gero1999 (approved):
Good job, I think it works fine! 💪🏽 

I am only concern if we should somehow keep deploying this part parallelly before merging. I feel that now that we have so many potential new collaborators going through the App, this section can be tricky for them to understand if we merge it incompleted...

PR #609, opened 2025-08-25
feat: simplifying volume units (#547)

### LINKED ISSUE #547: Enhancement: inappropriate units for volume parameters
### Description
Units for volume parameters are currently set as `clearance unit / lamdaz unit` without any conversions, eg `mg/(ng/mL)`. However for volume the units should be volume units only eg `mL`. Which in this case would require a conversion factor of  `1e+6`.

PPTEST = "Vol Dist Steady State Obs".

So units have to be simplified among the same unit, mg and ng are both grams.

### Definition of Done
- [x] Volume units in volume format only, no conversion required. --> At least needed for Vss, Vz

Gero1999:
## Issue

Closes #547 

## Description
Simplifies the units for the parameters with `metadata_nca_parameters$unit_type == volume`

## Definition of Done
- [X] Volume units in volume format only, no conversion required. --> At least needed for Vss, Vz

## How to test
Use `adnca_dummy_sm_dataset.csv` which has excretion-related samples
Map data
 `Parameter Units` table in NCA > Settings: Check VZ, VZO and other volume parameter unit simplifications. Try changing also at least one of them.
Run NCA
Check if by default volume units are simplified. Check if units modified are also modified

## Contributor checklist
- [x] New logic is documented


js3110 (changes requested):
Hi, nice work.
In terms of the dummy clinical data, everything looks really good and the conversion is perfect to mL, which is what we want!

However for the dummy preclinical dataset, its converting the `Vz `and `Vss `to `mL ng -1` with a conversion factor of smth like 0.0000001 which is not at all ideal. Since preclinical has units relative to body weight, this simplification is not correct. It should be simplified to `ml/kg` which has a conversion factor of 1000000 instead.

However I don't think that there is a uniform way to get it to do both of these conversions, so I would instead recommend that you try to just get it to convert to mL (or whatever the colume unit is) and if it can't do that then you leave it as the basic version. For preclinical we can just add it to their units template. (unless of course you are able to figure out a way to do this!)

m-kolomanski (approved):
LGTM.

suggestion: I would appreciate having an example in the docstrings with clear `input` -> `output` units. _simplify_ seems like quite a broad term and would be great to know what the function does to the units exactly.

js3110 (approved):
LGTM! Thanks this is much better, units are correct for clinical and fine for preclinical now :)

PR #639, opened 2025-09-12
fix: meanplot, individualplot, boxplot in zip even without rendering

### LINKED ISSUE #631: Bug: Zip download without meanplot
### Description
Zip download does not download the meanplot. The reason is if you don't go to the tab meanplot or individual plot of the exploration part, the plot is never rendered, and therefore the plot is never downloaded

The same problem also happens to individualplots & boxplots

### To reproduce
Do you NCA setup
Run NCA
Download the zip file

### TODO
- [ ] Mean plots are included in zip without rendering
- [ ] Boxplots are included in zip without rendering
- [ ] Individual plots are included in zip without rendering

Gero1999:
## Issue

Closes #631 

## Description
Includes meanplot for the download zip even if the user does not visit the tab in the UI

## Definition of Done
- [x] Direct download of meanplot even without visiting Explore tab
- [x] Boxplots are included in zip without rendering
- [x] Individual plots are included in zip without rendering

## How to test
Map data > Run NCA > Download ZIP > Check if meanplot is there

## Contributor checklist
- [x] New logic is documented


Gero1999:
hey for the default subjects selected I remember a PR done by Valentin! For mean plots we can open a new one, should be an easy (perhaps a `first issue`?)

Issue #333, opened 2025-04-08
Enhancement: add faceting option for slopes plots


js3110:
### Description
Sometimes it is nicer for the user to be able to view all the slopes plots for certain groups, eg COHORT, TRTO1 etc.

Lets try to add an option for the slopes plots so that the user can select a grouping option, and all plots from each group are on the same page.

### Definition of Done
The requirements for the feature to be complete:
- [ ] Grouping option for slopes plots
- [ ] Faceting so groups are all on one page

PR #641, opened 2025-09-12
Simplify the logic in slope plots, add order option & allow download in ZIP

### LINKED ISSUE #553: Enhancement: Download of Visualization & Individual slope plots
### Description
Generate the TLGs folder in  the downloaded `zip` with HTMLs (and if possible in PNGs or SVGs or similar). The zip downloaded is the one that the user can get in NCA results, is available after running NCA.

### Definition of Done
- [ ] Make a folder `Visualization` including its plots 
- [ ] Make a folder `Slope Individuals`

Gero1999:
## Description

This changes simplify the logic that makes the slope plots in order to make easier future implementations and also make able the download of all plots in a zip file easily. The advantages would be:

### Features
- Slope order input. Users can group their plots in whatever way they want them to appear. 

### Enhancements
 - All plots are generated, instead of only the rendered ones.
 - The new function that generates plots is faster and more aligned with PKNCA (`get_halflife_plot`)
 - Only modified plots are affected by changes, reducing unnecessary plot redraws
 - Upstream, the PKNCA object is updated with the slope rules directly in `setup.R`, instead of in the post-processing of results. We no longer need to use `filter_slopes`
 - We no longer use the custom `is.included.hl` or `is.excluded.hl` that then are used to build `exclude_half.life`. Instead we directly use the `exclude_half.life` & `include_half.life` columns from PKNCA
 - Added more documentation and comments for the files
- The new reactivity flow to create the plots that allows a simpler use of settings. We just simply need to udpate the dataset and/or manual slopes and everything will be auto-managed:

```
#' processed_pknca_data (input, from parent)
#'   └─> slope_selector_server
#'        ├──> plot_outputs (reactive, updated by processed_pknca_data changes)
#'        └──> ├> handle_plotly_click (updates manual_slopes on plot click)
#'                   └> handle_table_edits (updates manual_slopes on user edits & setting overrides)
#'                          └─> manual_slopes (output, used by parent to update processed_pknca_data)
```

### Bugs fixed
- Pagination controls are less likely to break (unless you are a clicking psycho 💀)
- Exclusion/Inclusion of points or table edits won't reset subject search
- Units appear for the Y axis

### Tests
- Setup `FIXTURE_PKNCA_DATA` obj will now contain the columns `include_half.life`, `exclude_half.life`

## Definition of Done

Closes or contributes to #333 (I think with this the client interest is already satisfied)
- [ ] Grouping option for slopes plots
- ~Faceting so groups are all on one page~

Closes #553

- [ ] Logic was simplified and code refactored
- [ ] All `Slope Selector` features work well without issues
- [ ] Make a folder `Slope selector` including its plots

## How to test

## Contributor checklist
- [ ] New logic is documented

## Notes to reviewer
Compare thoughtfully the feats of the plots to make sure I did not miss anything relevant. 
Don't hesitate also to leave suggestions for refactoring or code comments that makes the code easier to read.

### DIFF
--- a/inst/shiny/modules/tab_nca/setup/slope_selector.R
+++ b/inst/shiny/modules/tab_nca/setup/slope_selector.R
@@ -31,7 +31,7 @@ slope_selector_ui <- function(id) {
 
   div(
     class = "slope-selector-module",
-    slopes_table_ui(ns("manual_slopes")),
+    manual_slopes_table_ui(ns("manual_slopes")),
     # Help widget #
     dropdown(
       div(
@@ -192,7 +192,7 @@ slope_selector_server <- function( # nolint
 
     # Creates an initial version of the manual slope adjustments table with pknca_data
     # and handles the addition and deletion of rows through the UI
-    slopes_table <- slopes_table_server("manual_slopes", pknca_data, manual_slopes_override)
+    slopes_table <- manual_slopes_table_server("manual_slopes", pknca_data, manual_slopes_override)
     manual_slopes <- slopes_table$manual_slopes
     refresh_reactable <- slopes_table$refresh_reactable
 
--- a/inst/shiny/modules/tab_nca/setup/manual_slopes_table.R
+++ b/inst/shiny/modules/tab_nca/setup/manual_slopes_table.R
@@ -6,7 +6,7 @@
 #'
 #' @param id Shiny module id
 #' @return Shiny UI element (fluidRow)
-slopes_table_ui <- function(id) {
+manual_slopes_table_ui <- function(id) {
   ns <- NS(id)
 
   fluidRow(
@@ -26,8 +26,6 @@ slopes_table_ui <- function(id) {
   )
 }
 
-
-
 #' Manual Slopes Table Server for Slope Selection
 #'
 #' Server module for managing the manual slopes table (inclusion/exclusion rules).
@@ -39,7 +37,7 @@ slopes_table_ui <- function(id) {
 #' @return List with:
 #'   - manual_slopes: reactiveVal containing the current manual slopes table
 #'   - refresh_reactable: reactiveVal for triggering table re-render
-slopes_table_server <- function(
+manual_slopes_table_server <- function(
   id, pknca_data, manual_slopes_override = NULL
 ) {
   moduleServer(id, function(input, output, session) {
@@ -206,7 +204,6 @@ slopes_table_server <- function(
         manual_slopes(manual_slopes_override())
       })
     }
-
     # Output: manual_slopes (reactiveVal) and refresh_reactable (for UI updates)
     list(
       manual_slopes = manual_slopes,


js3110:
can you link/create the issue that this is closing, just for clarity please @Gero1999 thanks!
Its a very large PR so I will take my time with it, I may have a few questions along the way

js3110 (changes requested):
Hi,

Here is my first round of feedback based on the user perspective only, I haven't fully checked   out the code yet!

First of all, well done! There is lots going on here (might have been better to break it down but ik its not always possible) and the changes look really good! I really like the look of the new plots, and I like the fact that the range now uses the time rather than the index. 

Some issues:

- if I add slope selections via table and via clicking then it crashes:
`Error in rbind: numbers of columns of arguments do not match`

- manual additions table doesn't update based on settings (eg if i deselect CSF, its still an option in the add exclusions table)

- reason for exclusion used to be required but this check has been removed here (this feature was implemented due to request from stakeholders)
- I can't see the x-axis on the plots due to the bar at the bottom when using just my laptop screen (maybe you need to add a buffer/margin in the styling?)
- The plots have a log y axis but its not labelled as such (maybe we can even give them a toggle option for log /linear but first lets just ensure its labelled correct)

This is what I can see from my first few run throughs. But in terms of results and PKNCA handling, everything looks good so far!

Gero1999:
> if I add slope selections via table and via clicking then it crashes:
Error in rbind: numbers of columns of arguments do not match

Should be fixed now, good catch!

> manual additions table doesn't update based on settings (eg if i deselect CSF, its still an option in the add exclusions table)

I added a simple fix. The table will update, but it will need first an action from the user after the change was done. If needed we can open another issue for a more complicated fix

> reason for exclusion used to be required but this check has been removed here (this feature was implemented due to request from stakeholders)

Should be fixed now

>I can't see the x-axis on the plots due to the bar at the bottom when using just my laptop screen (maybe you need to add a buffer/margin in the styling?)

For this UI/UX aspects I will consider it in a new issue, once we also receive the feedback from Artur

> The plots have a log y axis but its not labelled as such (maybe we can even give them a toggle option for log /linear but first lets just ensure its labelled correct)

This should be fine. The scale is log, but the values represented are raw/correct. This in my opinion is better for reviewing the plots than what we had.

Gero1999:
> The error message is given but then run NCA is still triggered and the page gets stuck in the loading NCA pop up...

This problem comes from `main` but I guess I can give it a try and if not open a new issue

> Yes fine for now, I meant the axis label should say Log AVAL not just AVAL. For accuracy. 

The scale is log, but the values represented are raw. Please, believe me when I say is correct/accurate 😅 

> I noticed you have removed NCA Profile (...)

In our last meeting the scientists told me that they did not like the `NCA_PROFILE` for this section. They were not sure what other variable would be better... Valentin proposed VISIT, but again nothing seemed conclusive. We need to really simplify as much as possible. Right now this variable is producing conflicts also with the dev part (PKNCA) and the new users (which never understand what this variable means). But I will see what I can come up to in the next days

> Therefore if possible could you go back to ARRLT

This I am not sure is really necessary. For these plots the users only want to inspect that all the half.life were well estimated (`r^2`, `lambda.z.n.points`, `span.ratio`) so the use of relative or absolute won't matter at all.

js3110:
maybe instead of NCAProfile we use the original column name that was mapped to NCA profile? then its whatever variable they are using

Gero1999:
> The plots look beautiful! And everything seems to be working as expected. I cannot figure out why, but when a new row is added with the Exclusion/Selection button, the RANGE is now written with a "." instead of a ":", e.g. 1.5 instead of 1:5. Not sure if this was on purpose. Writing the range with a ":" still works though. A suggestion for another issue maybe: rewriting the tooltips so that they clearly say AVAL:.., AFRLT:... ?

@lucyaspridis Notice that the default is a `Exclusion` not a `Selection`. The default is just excluding the second point of the profile

Regarding the tooltips I am still thinking about it. I just want something nice for PKNCA as well

lucyaspridis:
Right sorry. This new definition of RANGE makes a lot of sense.
Not sure it is working exactly as expected then (unless I'm not understanding how it works):
Screenshot 2025-10-22 164904

js3110 (changes requested):
Hi,

I've noticed another couple of bugs:

1. If I add an exclusion, run NCA, then go back to setup and remove the exclusion, I get this error:
`Warning: Error in if: missing value where TRUE/FALSE needed`

This happened when I added exclusion with double click, but removed with selecting from the table. I haven't tried other combinations but might be also worth checking.

2. If I run through the workflow quite fast, and add an exclusion, I clikc the run NCA buton then nothing happens. I wait a few seconds and click again then nothing happens, repeat. Then eventually it runs and it Runs NCA like 3 times. Has happened to me twice now so I know its a real bug- but hard to recreate unorganically...

R/get_halflife_plots.R
```diff
@@ -0,0 +1,275 @@
...
+  if (nrow(pknca_data$conc$data) == 0 || nrow(pknca_data$intervals) == 0) {
+    return(list(plots = list(), data = list()))
+  }
+
+  # Identify column names
+  time_col <- pknca_data$conc$columns$time
+  conc_col <- pknca_data$conc$columns$concentration
+  timeu_col <- pknca_data$conc$columns$timeu
+  concu_col <- pknca_data$conc$columns$concu
+  exclude_hl_col <- pknca_data$conc$columns$exclude_half.life
```
m-kolomanski: style: Maybe instead of spelling out all the sub-values for each of the variables, it would be neater to assign just

```R
data_cols <- pknca_data$conc$columns
```

and then use it to call

```R
if (is.null(data_cols$exclude_half.life)) {
 # ...
}
```

R/get_halflife_plots.R
```diff
@@ -0,0 +1,275 @@
...
+                     ifelse(is_halflife_used, "green", "red")),
+      symbol = ifelse(.[[exclude_hl_col]], "x", "circle")
+    ) %>%
+    group_by(!!!syms(c(group_vars(pknca_data), "start", "end"))) %>%
+    group_split()
+
+  plot_list <- list()
+  data_list <- list()
+  for (i in seq_len(length(info_per_plot_list))) {
+    df <- info_per_plot_list[[i]]
```
m-kolomanski: suggestion: Consider using some kind of `map` or `apply` function.

R/utils-slope_selector.R
```diff
@@ -1,170 +1,273 @@
...
+  # Get the intervals of the plots affected by the current rules
+  pknca_data$intervals <- inner_join(
+    intervals_to_update,
+    pknca_data$intervals,
+    by = intersect(names(intervals_to_update), names(pknca_data$intervals))
+  )
+  updated_plots <- suppressWarnings(get_halflife_plots(pknca_data)[["plots"]])
+  plot_outputs[names(updated_plots)] <- updated_plots
+  plot_outputs
+}
```
m-kolomanski: issue: Seems to me like this function is only useful in the context of the shiny app, and not in a package, should be in `inst/shiny`

inst/shiny/modules/tab_nca/setup/handle_table_edits.R
```diff
@@ -0,0 +1,216 @@
+
+#' Manual Slopes Table UI for Slope Selection
+#'
+#' UI module for displaying and editing the manual slopes table (inclusion/exclusion rules).
+#' Provides buttons to add/remove rules and a reactable table for editing.
+#'
+#' @param id Shiny module id
+#' @return Shiny UI element (fluidRow)
+handle_table_edits_ui <- function(id) {
```
m-kolomanski: issue: This is not a logic function, it does not **handle** anything, it just defines UI. The module does not handle anything really, in the common sense of this word. Why is this module not named just `slopes_table` or `manual_slopes_table` or something like this?

inst/shiny/modules/tab_nca/setup/handle_table_edits.R
```diff
@@ -0,0 +1,216 @@
...
+        }) |>
+          all()
+        if (!override_valid) {
+          msg <- "Manual slopes not compatible with current data, leaving as default."
+          log_warn(msg)
+          showNotification(msg, type = "warning", duration = 5)
+          return(NULL)
+        }
+        manual_slopes(manual_slopes_override())
+      })
```
m-kolomanski: question: Does this work with uploaded settings? Seems like this `if` statement will be executed only once and the observer will never be attached, so the manual override will never work, but I might be missing something.

inst/shiny/modules/tab_nca/setup/slope_selector.R
```diff
@@ -71,310 +81,167 @@ slope_selector_ui <- function(id) {
...
+
+    # Call the pagination/searcher module to:
+    # - Providing indices of plots for the selected subject(s)
+    # - Providing indices for which plots to display based on pagination
+    page_search <- page_and_searcher_server(
+      id = "page_and_searcher",
+      search_subject = search_subject_r,
+      plot_outputs = plot_outputs,
+      plots_per_page = plots_per_page_r
+    )
```
m-kolomanski: style:
```suggestion
    # Call the pagination/searcher module to:
    # - Providing indices of plots for the selected subject(s)
    # - Providing indices for which plots to display based on pagination
    page_search <- page_and_searcher_server(
      id = "page_and_searcher",
      search_subject = reactive(input$search_subject),
      plot_outputs = plot_outputs,
      plots_per_page = reactive(input$plots_per_page)
    )
```

inst/shiny/o_nca
m-kolomanski: question: What is this file?

tests/testthat/test-get_halflife_plots.R
m-kolomanski: issue: Apart for two first test, the test descriptions do not follow the structure. `get_halflife_plot marker colors and shapes and shapes - no exclusion/inclusion` is not a valid sentence. Should be something like `it("renders markers, colors and shapes appropriately with no exclusion or exclusion")`.

inst/shiny/modules/tab_nca/setup/slope_selector.R
```diff
@@ -71,310 +81,167 @@ slope_selector_ui <- function(id) {
...
-              const plotlyElements = $('.slope-selector-module .plotly.html-widget.html-fill-item.html-widget-static-bound.js-plotly-plot');
-              plotlyElements.css('height', '100%');
-              plotlyElements.css('aspect-ratio', '1');
-              window.dispatchEvent(new Event('resize'));
-            }"
-            # nolint end
-          )
-      })
+    # HACK: workaround to avoid plotly_click not being registered warning
+    session$userData$plotlyShinyEventIDs <- "plotly_click-A"
```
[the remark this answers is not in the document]
Gero1999: good catch, this line is actually not needed

R/get_halflife_plots.R
```diff
@@ -0,0 +1,275 @@
...
+#' @return A list with plotly objects and data
+#' @importFrom dplyr filter select mutate group_by ungroup group_split %>% any_of
+#' @importFrom stats lm predict as.formula
+#' @importFrom plotly plot_ly add_lines layout add_trace plotly_build
+#' @importFrom PKNCA pk.nca
+get_halflife_plots <- function(pknca_data, add_annotations = TRUE) {
+
+  # If the input has empty concentration or intervals, just return an empty list
+  if (nrow(pknca_data$conc$data) == 0 || nrow(pknca_data$intervals) == 0) {
+    return(list(plots = list(), data = list()))
```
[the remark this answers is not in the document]
Gero1999: I made this intentionally a list of lists because I want both things to be mapped with each other but I will think about this

R/PKNCA.R
```diff
@@ -694,3 +705,49 @@ PKNCA_hl_rules_exclusion <- function(res, rules) { # nolint
...
+##' @details
+##' - If any excluded half-life points are missing a reason, an error is thrown.
+##' - If no exclusions or all have reasons, the function returns the input object.
+##' - Used to enforce good practice/documentation before NCA calculation.
+##'
+##' @examples
+##' # Suppose processed_pknca_data is a valid PKNCA data object
+##' # checks_before_running_nca(processed_pknca_data)
+checks_before_running_nca <- function(processed_pknca_data, exclusions_have_reasons = TRUE) {
+  if (exclusions_have_reasons) {
```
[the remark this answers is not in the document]
Gero1999: you might be right but I don't follow this point. The arg allows the user to decide if they want to make that check or not. It is not for them to escape the check because they don't need it, but more in case they just simply don't want to comply with it

For now we only have this check, but is likely that we keep increasing it with more

R/utils-slope_selector.R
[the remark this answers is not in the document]
Gero1999: this is a good point, I will move most of them, but probably for the R script I will need at least `update_pknca_with_rules` so I guess I will keep this one. I will name the file in `/functions` the same as this one to show they are connected

Gero1999:
@js3110 regarding the very specific user bugs:

> If I add an exclusion, run NCA, then go back to setup and remove the exclusion, I get this error:
Warning: Error in if: missing value where TRUE/FALSE needed

This is fixed

> If I run through the workflow quite fast, and add an exclusion, I clikc the run NCA buton then nothing happens. I wait a few seconds and click again then nothing happens, repeat. Then eventually it runs and it Runs NCA like 3 times. Has happened to me twice now so I know its a real bug- but hard to recreate unorganically...

This I was not able to replicate it neither here nor in main. Maybe was fixed with after the merge. Otherwise please share me a screen recording and the console messages and I will try to look into it

_____________________________________________________________________________________

@m-kolomanski regarding the suggestions:

Most of them were addressed. Some style changes specifically for `get_halflife_plots.R` I will in principle not implement them because they conflict with PKNCA style, and the function might likely be transferred there at some point

js3110 (changes requested):
Hey @Gero1999 , thanks, the bugs from before don't seem to be there anymore!

Unfortunately I ran into some other  issues:

If I add a selection it doesn't update the plot:
image


If I click all the way to the end, I then can't change page anymore:

image

I also got an error after adding the following selections/exclusions:
image

```
ERROR [2026-01-07 11:08:17] Error calculating NCA results:
ℹ In index: 13.
Caused by error in `assert_numeric_between()`:
! Please report a bug.
: Error with interval start=0, end=24.018: Assertion on 'lambda.z' failed: Element 1 is not > 0
```
It worked after I removed the selection. I tried a different selection and this time it worked (the plot updated but only after I ran the NCA)


I removed an exclusion and ran the NCA again, then when I went to add it back it added two rows to the table:
image

Now every time I try to add a new exclusion/selection it adds all the previous ones too:


https://github.com/user-attachments/assets/5432e160-93a7-4977-a27f-eb09dcc88f86

PR #643, opened 2025-09-15
Bug: FREXINT included in excretion module

### LINKED ISSUE #642: Enhancement: FREXINT in excretion results
### Description
Excretion module needs to be updated to include FREXINT and VOLPK as parameter options
FE should be in %

### To reproduce
Excretion module -> FREXINT not available as parameter

### Expected behaviour
- [ ] FREXINT and VOLPK can be calculated in excretion module
- [ ] FREXINT as a %

js3110:
## Issue

Closes #642 

## Description

Added FREXINT (and other urine params) to excretion module in additional analysis section

## Definition of Done

- [x] FREXINT and VOLPK are included in excretion module calculations
- [x] FREXINT unit is % by default

## How to test

Upload dataset with urine (eg `adnca_dummy_sm_dataset.csv`), run NCA and see excretion parameters, and go to Additional Analysis -> Excretion to calculate by sample and see results)

## Contributor checklist
- [x] New logic is documented


inst/shiny/modules/tab_nca/excretion_analysis.R
```diff
@@ -158,6 +161,10 @@ excretion_server <- function(id, input_pknca_data) {
         filter(type_interval %in% input$interval_types) %>%
         arrange(PCSPEC, start, end)
 
+      data$units <- data$units  %>%
+        mutate(PPSTRESU = ifelse(PPTESTCD == "fe", "%", PPSTRESU),
```
Gero1999, line 165: question: I am a bit concerned about this. Don't we need any additional check on the units to make sure is safe to say is a pure fraction? @js3110

js3110: hey @Gero1999 , from what I can tell when trying this, it won't calculate FE if its not a pure fraction (you can try with the dummy preclinical dataset- once with body weight conversion and once without. the results are empty for no bw)

And if its not a pure fraction then its not the correct results so IMO this should work

PR #646, opened 2025-09-16
Bug: parameters not selected shown in results

### LINKED ISSUE #606: Bug: Parameters not selected shown in results if required for calculations

js3110:
## Issue

Closes #606 

## Description

If parameters are required for the calculation of otheres (eg in Depends col of metadata), or if they are required for rule sets, the parameters were shown in the results despite not being selected in parameter selection.
Now, the results are filtered based on parameter selection. They are filtered after being calculated to ensure that any params required for calculations or rule sets are available for processing.

## Definition of Done

- [x] Parameters filtered based on parameter selection
- [x] No additional parameters shown in results

## How to test

Run NCA with only one parameter eg RSQADJ, view difference between this branch and main. Check that rule sets are still correctly applied.

## Contributor checklist
- [x] New logic is documented

### DIFF
--- a/inst/shiny/modules/tab_nca/setup/settings.R
+++ b/inst/shiny/modules/tab_nca/setup/settings.R
@@ -279,7 +279,18 @@ settings_server <- function(id, data, adnca_data, settings_override) {
         current_profile <- profile_choices[1]
       }
       if (length(current_pcspec) == 0) {
-        current_pcspec <- pcspec_choices
+        # Select plasma/serum if available
+        plasma_serum_values <- grep("^plasma$|^serum$",
+                                    pcspec_choices,
+                                    value = TRUE,
+                                    ignore.case = TRUE)
+
+        # Assign to current_pcspec if found, otherwise select all
+        if (length(plasma_serum_values) > 0) {
+          current_pcspec <- plasma_serum_values
+        } else {
+          current_pcspec <- pcspec_choices
+        }
       }
 
       updatePickerInput(


inst/shiny/modules/tab_nca.R
```diff
@@ -210,10 +210,23 @@ tab_nca_server <- function(id, adnca_data, grouping_vars) {
 
         log_success("NCA results calculated.")
 
-        # Apply standard CDISC names and return the object
-        res %>%
-          mutate(PPTESTCD = translate_terms(PPTESTCD, "PKNCA", "PPTESTCD"))
+        # Reshape intervals, filter for TRUE
+        requested_params <- res$data$intervals %>%
+          pivot_longer(
+            cols = where(is.logical),
```
Gero1999, line 216: issue: parameters can also be numeric/NA (not only TRUE/FALSE). So it is better if we can reshape a bit this to prevent future issues. Consider a strategy similar to the one done in `parameter_datasets.R`, which should work fine: 

```
      params_not_requested <- res_nca_req$data$intervals %>%
        select(any_of(setdiff(names(PKNCA::get.interval.cols()), c("start", "end")))) %>%
        # For all logical columns, mutate FALSE to NA
        mutate(across(where(is.logical), ~ ifelse(.x, TRUE, NA))) %>%
        # Only select column that are only NA
        select(where(~ all(is.na(.x)))) %>%
        names()
      res_nca_req$result <- res_nca_req$result %>%
        filter(!PPTESTCD %in% translate_terms(params_not_requested, "PKNCA", "PPTESTCD"))
```

Gero1999 (changes requested):

m-kolomanski (approved):
LGTM

Gero1999 (changes requested):
Hey @js3110 this might not work for ratio parameters when customized... If possible follow the strategy indicated in this [previous feedback](https://github.com/pharmaverse/aNCA/pull/646#discussion_r2352540464)

PR #653, opened 2025-09-18
Enhancement: default selection of parameters using metadata table

### LINKED ISSUE #617: Enhancement: Default selection of parameters using metadata table
### Description
Parameters should be selected based on study type, using the metadata information (is_excretion, is_extravascular etc)

Currently the metadata just shows T or F based on whether the parameter should be calculated or not, but doesn't state whether its default. Options are:

1. Update metadata to say whether a parameter should be selected by default or not (eg Default, Optional, False (shouldn't be selected))
2. Keep list of default parameters as is, and just ensure that those that are F in metadata for specific study types are not selected
### Definition of Done
- [x] default selection of parameters per study type
- [x] study types where default parameter = F in metadata should not be selected

js3110:
## Issue

Closes #617 as part of #436 

## Description

Parameter selections are updated by default using the default primary and secondary parameters, as well as specifications on study types and parameters from the metadata. Eg excretion parameters are only selected for excretion data etc.

## Definition of Done

- [x] Default parameters selected correctly per study type
- [x] Option to override with manual selections
- [x] Errors/warning messages given if parameters are not able to be calculated.

## How to test
Test with dummy set and also Test dummy set `adnca_dummy_sm_dataset.csv` to see the different study types. See the different default parameter selections and run NCA to view that they have been applied in results.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

The long list of if statements seems a bit long and repetitive to me, but I couldn't figure out how to refine it. I tried a couple of things but then the functionality was lost... open to suggestions !

### DIFF
--- a/R/data.R
+++ b/R/data.R
@@ -25,6 +25,8 @@
 #'    Indicates if the parameter can be used in multiple dose analysis}
 #'   \item{can_extravascular}{Logical.
 #'    Indicates if the parameter can be used in extravascular analysis}
+#'    \item{can_metabolite}{Logical.
+#'    Indicates if the parameter can be used in metabolite analysis}
 #' }
 #' @source Generated for use in the `translate_nomenclature` function.
 "metadata_nca_parameters"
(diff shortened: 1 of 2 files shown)


lucyaspridis (changes requested):
Looks great! I see the variable can_metabolite has been newly added to metadata_nca_parameters, to forget to document it in R/data.R :)

PR #674, opened 2025-10-02
Bug: nca loading won't close if error

### LINKED ISSUE #655: Bug: NCA loading bar prevents exit if error
### Description
If for some reason NCA cannot be calculated and an error pops up (eg manual slope exclusion reason not given, or invalid parameter selected), the NCA loading bar just stays on the page and the user is unable to exit.

### To reproduce
Run NCA, add exclusion with no reason and run- see error message and frozen page.

### Expected behaviour
Ability for user to exit the loading page/ page to return to NCA tab if error is given.

js3110:
## Issue

Closes #655 

## Description

When an error caused the NCA run to fail, the modal loading pop up would not close so the user is stuck. I considered two possibilities: adding a close button so the user can close the modal, or closing it automatically upon error. I decided to go for the latter since having a manual close button could be misleading and make the user think that they can cancel the run (they cant)

## Definition of Done

- [ ] Loading bar NCA modal closes if error
- [ ] Loading bar NCA remains open until results are shown if no error.

## How to test

Run NCA for a couple of processes:

- Normal workflow (check everything is normal)
- Select all standard parameters (this will give an error because not all parameters can be calculated for the dummy data)
- Go to Slope Selector and add an exclusion with no reason -> Run NCA (should give error because no reason is given)
- Try any other runs that you think may be useful.

## Contributor checklist
- [ ] New logic is documented

### DIFF
--- a/inst/shiny/functions/loading_popup.R
+++ b/inst/shiny/functions/loading_popup.R
@@ -12,7 +12,7 @@ loading_popup <- function(text = "Loading...") {
   showModal(modalDialog(
     modal_body,
     title = span(text, id = "loading-title"),
-    footer = modalButton("Close"),
+    footer = NULL,
     size = "s"
   ))
 }
--- a/inst/shiny/functions/loading_popup.R
+++ b/inst/shiny/functions/loading_popup.R
@@ -12,7 +12,7 @@ loading_popup <- function(text = "Loading...") {
   showModal(modalDialog(
     modal_body,
     title = span(text, id = "loading-title"),
-    footer = NULL,
-    size = "m"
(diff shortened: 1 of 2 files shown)


Gero1999 (changes requested):
I think this is the right fix direction, but unfortunately it is not working for me with the Exclusion case

js3110:
hmmm, thats strange @Gero1999 as it works for me... could you try again and make sure you've pulled etc, then if it doesn't work still maybe you can show me a screen recording of what you are doing?

js3110:
@Gero1999 hmmm I have no idea why its not working for you. I did the exact same and it works for me...

Issue #618, opened 2025-09-01
Release package in CRAN


Gero1999:
### Description
We would like to have a non-beta first version of the package released in CRAN by November. So we might need to have already started the application by October. 

### Definition of Done
- [x] Create a `NEWS.md` document
- [x] Pass all tests and hopefully no-warnings/notes
- [ ] Release a first version of `aNCA` in CRAN
- [x] Update main author to Valentin for easier communication
- [x] Make (if possible) repo occupies <= 5MB

- [ ] Fix feedback from last review:
- Package names should always be in single quotes (e.g `'shiny'`)
- Description field should be one paragraph. Include on it references for the methods as well if relevant.
- Add \value to .Rd files regarding exported methods and the structure of the output (class). run_app.Rd
- -> Used ::: in documentation:
      man/convert_to_iso8601_duration.Rd:
         aNCA:::convert_to_iso8601_duration(200, "h")
      man/convert_to_iso8601_duration.Rd:
         aNCA:::convert_to_iso8601_duration(5, "d")
- \dontrun{} should only be used if the example really cannot be executed
(e.g. because of missing additional software, missing API keys, ...) by
the user. That's why wrapping examples in \dontrun{} adds the comment
("# Not run:") as a warning for the user. Does not seem necessary.
Please replace \dontrun with \donttest. Please unwrap the examples if they are executable in < 5 sec, or replace
dontrun{} with \donttest{}. Please replace the \dontrun{}-wrapper with if(interactive()){} for
interactive examples.
- Please do not set a seed to a specific number within a function. -> R/flexible_violinboxplot.R

#### Note: Feel free to edit my comment to edit the TODO if needed

PR #678, opened 2025-10-02
Fix: suppress warnings in test setup

### LINKED ISSUE #676: Enhancement: target and silence warnings in tests
### Description
there are currently 12 warnings from tests that need to be silenced for CRAN release:

```
Warning messages:
1: In PKNCAdata.default(data.conc = data.conc, data.dose = data.dose,  ... :
  PCSPEC=SERUM; USUBJID=2; PARAM=A; data_intervals=NULL: No intervals generated likely due to limited concentration data
2: In PKNCAdata.default(data.conc = data.conc, data.dose = data.dose,  ... :
  PCSPEC=SERUM; USUBJID=3; PARAM=A; data_intervals=NULL: No intervals generated likely due to limited concentration data
3: In PKNCAdata.default(data.conc = data.conc, data.dose = data.dose,  ... :
  PCSPEC=SERUM; USUBJID=8; PARAM=A; data_intervals=NULL: No intervals generated likely due to limited concentration data
4: In PKNCAdata.default(data.conc = data.conc, data.dose = data.dose,  ... :
  PCSPEC=SERUM; USUBJID=8; PARAM=B; data_intervals=NULL: No intervals generated likely due to limited concentration data
5: Requesting an AUC range starting (0) before the first measurement (0.5) is not allowed
6: Requesting an AUC range starting (0) before the first measurement (0.5) is not allowed
7: Requesting an AUC range starting (0) before the first measurement (0.5) is not allowed
8: Requesting an AUC range starting (0) before the first measurement (0.5) is not allowed
9: Requesting an AUC range starting (0) before the first measurement (0.5) is not allowed
10: Requesting an AUC range starting (0) before the first measurement (0.5) is not allowed
11: Requesting an AUC range starting (0) before the first measurement (0.5) is not allowed
12: Requesting an AUC range starting (0) before the first measurement (0.5) is not allowed
```

### Definition of Done
- [ ] warnings silenced

js3110:
## Issue

Closes #676 

## Description

Suppress warnings in setup, part of #618 

## Definition of Done

- [x] tests run fine
- [x] no warnings

## How to test

run `devtools::test()` and check that there are no warnings >> This is intended to be as clean as possible for CRAN release

## Contributor checklist
- [x] New logic is documented


js3110:
> ### Warnings solved, but some calls completely suppressed
> Good job @js3110, but would not be also possible to only suppress the particular warnings for `FIXTURE_PKNCA_DATA`? (not all warnings indiscriminately)

> Warning ('test-app-mapping.R:15:5'): Test for mapping interface: mapping interface appears without errors
> Adding new file snapshot: 'tests/testthat/_snaps/app_mapping-001_.png'
> ```

I could just add this file and then it won't give the warning I think? this file is created whenever I run tests so maybe better to add it...?

> and this other warning @osenan is more a thing of ggplot so I think can be commented in the CRAN-comments, but lmk if you think we should invest time in solving it differently (or perhaps silence it?):
> 
> ```
> Warning ('test-g_pkcg.R:12:5'): pkcg01: generates valid ggplots with LIN scale
> <ggplot> %+% x was deprecated in ggplot2 4.0.0.
> i Please use <ggplot> + x instead.
> ```

The issue here is that the remote is using ggplot2 4.0.0, which uses + instead of %+%. This version of ggplot2 cannot be installed on the version of R that I am using (i am not sure what the lowest package level is) but I'm not sure what the best way to solve this is-> update our minimum R version in the DESCRIPTION? or set a maximum ggplot2 version for now? or your suggestions

Issue #486, opened 2025-06-23
Enhancement: Better tooltips in column mapping


Gero1999:
### Description
The descriptions for the column mapping are inconsistent, sometimes provide enough information other not. We need to make them perfect for the user so they understand exactly the column format the App expects. There are some variables which tooltips can also be wrong/outdated (`ROUTE` is not necessarily intravascular/extravascular). 
Think of: format (factor/numeric/character) and mathematical domain/code lists (can even add link to guidelines 

### Definition of Done
The requirements for the feature to be complete:
- [ ] update when relevant the tooltips (`ROUTE`)
- [ ] tooltip messages are consistent in their writing style and accurate enough for the user to replicate

PR #687, opened 2025-10-07
Flexibilize Column Mapping Interface & tooltips

### LINKED ISSUE #681: Enhancement: Column mapping can include options for other non-unit columns
### Description
Column mapping should be more flexible and allow to map unique values instead of forcing column names always for the inputs.

### Definition of Done
- [ ] Instead of mapping a ROUTE column, user should be able to map a value `Intravascular / Extravascular`
- [ ] Instead of mapping a PCSPEC column, user should be able to map a value `BLOOD / PLASMA / URINE`
- [ ] Others perhaps that may be interesting

Gero1999:
## Issue

Closes #681 #486

## Description
The changes link now the mapping definitions with the `metadata_nca_variables` object. Additionally, more default options as well as choices will be provided for the user based on the column (e.g, ROUTE has `intravascular` and `extravascular`, PCSPEC has `Plasma` or `Urine`)

## Definition of Done
#681
- [x] As well as mapping a ROUTE column, user should be able to map a value `Intravascular / Extravascular`
- [x] As well as mapping a PCSPEC column, user should be able to map a value `BLOOD / PLASMA / URINE`
- [x] Move DRUG to supplementary variables (and automatically map PARAM if its NA)

#486
- [x] update when relevant the tooltips (`ROUTE`)
- [x] tooltip messages are consistent in their writing style and accurate enough for the user to replicate

## How to test
- Go to mapping and try different mappings with ROUTE, PCSPEC
- Review also that relevant functionalities related to the mapping have not been lost

## Contributor checklist
- [x] New logic is documented


inst/shiny/functions/update_selectize_inputs.R
```diff
@@ -6,89 +6,52 @@
...
-      }
-    }
-    updateSelectizeInput(
-      session, input_id,
-      choices = c("Select Column" = "", special_cases[[input_id]]),
-      selected = selected
+      choices = list(
+        "Select Column" = "",
+        "Mapping Columns" = data_colnames,
+        "Mapping Values" =  strsplit(alternative_choices_list[[input_colname]]$Values, ", ")[[1]]
```
lucyaspridis: nitpick: very tiny detail that I noticed
```suggestion
        "Mapping Values" = strsplit(alternative_choices_list[[input_colname]]$Values, ", ")[[1]]
```

inst/shiny/modules/tab_data/data_mapping.R
```diff
@@ -1,32 +1,45 @@
...
+NON_STD_MAPPING_INFO <- data.frame(
+  Variable = c("Grouping_Variables", "Metabolites"),
+  Label = c("Variables to group and summarise results", "PARAM values to flag as metabolites"),
+  Order = c(100, 16),
+  Values = c("", ""),
+  mapping_tooltip = c(
+    "Additional column(s) to use to group the data in the outputs (i.e, 'AGE', 'SEX')",
+    paste0(
+      "Choose the PARAM values to flag as metabolites of the drug parent (METABFL = 'Y'). ",
+      "If empty is assumed that all PARAM correspond to the parent drug (METABFL = '')"
```
lucyaspridis:
```suggestion
      "Choose the PARAM values to flag as metabolites of the parent drug (METABFL = 'Y'). ",
      "If empty, it is assumed that all PARAM values correspond to the parent drug (METABFL = '')"
```

lucyaspridis (approved):
LGTM but requesting some very small nitpick changes:)

lucyaspridis (changes requested):
I also just noticed, using `adnca_dummy_sm_dataset.csv`, that VOLUMEU does not seem to get automatically mapped anymore. Moreover, there aren't any options for Mapping Values in the VOLUMEU selectize input because they are not specified in the metadata_nca_variables dataset.

lucyaspridis:
I've seen some studies where VOLUMEU was "g" so I would add it to the list of possible values:)

Gero1999:
> I've seen some studies where VOLUMEU was "g" so I think it should be added to the list of possible values:) Thank you for having made some of the changes I suggested, I'll approve when all the tooltips are correct:)

I made an update This is something some companies do, but is wrong based on ADNCA standards, so if it is fine I will keep it like this now!

lucyaspridis (approved):
LSTM (looks slay to me)

PR #704, opened 2025-10-14
Bug: apply mapping duplicates filtered wrong

### LINKED ISSUE #703: Bug: Apply mapping duplicates filtered wrong
### Description
In apply column mapping function, after merge of #640 the duplicate filtering is now incorrect, due to wrong grouping variables. For the initial duplicate filters, we should group by all columns except for ARRLT, NRRLT and NCA_PROFILE, to ensure that only exact duplicates are filtered. (including predose/last dose switches hence exclusion of those 3 columns.)

### To reproduce
Use DummyROADNCA data in data folder, it used to give a popup for duplicates filtering but now everything is just filtered out automatically.

### Expected behaviour
If AVAL not matching, duplicates should be shown in a popup.

js3110:
## Issue

Closes #703 

## Description

Fix bug that filtered all duplicates as opposed to only duplicates with matching AVAL. Other duplicates are now shown in pop up window.

## Definition of Done

- [x] Only exact AVAL duplicates automatically filtered

## How to test

Use DummyRO_ADNCA dataset, after column mapping popup should appear for duplicates.

## Contributor checklist
- [x] New logic is documented

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.0.0.9169
+Version: 0.0.0.9170
 Authors@R: c(
     person("Ercan", "Suekuer", email = "ercan.suekuer@roche.com", role = "aut",
            comment = c(ORCID = "0009-0001-1626-1526")),


lucyaspridis (approved):
LGTM. This might be the right moment to mention that the message on the pop-up window with duplicates says ""The following rows are duplicates based on TIME, STUDYID,
              PCSPEC, DRUG, USUBJID, PARAM." How about changing TIME to AFRLT for more precision?

Issue #707, opened 2025-10-14
Enhancement: Align variable names with ADNCA standards


Gero1999:
### Description
We have some variables that are not matching their correct ADNCA names. The specific changes needed involve:

### Definition of Done
Make the next renaming within the App code, metadata, data and website:

- [ ] TAU > TRTINT
- [ ] DRUG > (METABFL & TRT)
- [ ] NCA_PROFILE > AVISIT

PR #708, opened 2025-10-15
Enhancement: Align variable names with ADNCA standards

Gero1999:
## Issue

Contributes to #707 
(merges to PR #687)

## Description
In general just the next changes happened in all relevant docs: `TAU > TRTINT`, `DRUG > TRT`, `NCA_PROFILE > AVISIT`

- The logic for `is.metabolite` was simplified in `create_start_impute.R` & `detect_study_types.R` to use `METABFL` instead of `PARAM != DRUG`
- In `export_cdisc` we now can assume AVISIT is there intact and can be used for PPGRPID
- The datasets `Dummy_complex_data.csv` & `Dummy_sm_data.csv` have now `DRUG > TRT` & `METABFL` columns

## Definition of Done
Make the next renaming within the App, code, metadata, data and website:

- [x] TAU > TRTINT
- [x] DRUG > (METABFL & TRT)
- [x] NCA_PROFILE > AVISIT

## How to test
- Regular mapping & NCA run works
- Check NCA > Summary  and make sure the study descriptions seem correct as well as the imputations 
- Check unit tests locally & any contributor checklist you feel you can run :)

## Contributor checklist
- [ ] New logic is documented


js3110 (commented):
A couple more questions:

1. TRTRINT is only allowed to be planned? because in our ADNCA data, TAU is actual so either we need to change the label or find out if ADNCA have an actual trtrint variable. For the mapping the user can map either.

2. What if data does not have METABFL column? we should still assume that if PARAM != TRT then it is a metabolite? Currently this doesn't happen. For non ADNCA datasets this is an additional column not required, so they might want a better workaround to ensure metabolites are identified...
Not sure if its ever a scenario that NCA will be run on an Analyte that is not a TRT and also not a metabolite? maybe worth asking!

Gero1999:
@Shaakon35 I think I addressed all but lmk if I missed something

Gero1999:
Yep I don't like that assumption. I would prefer to keep things simple and consistent for all variables. If you don't have METABFL, then there is no metabolite data. I think this is more intuitive for the user. However, I will include a warning as I did with the others.

Gero1999:
Right now in the dropdown they can choose values for "Y"/"". In the future for non-automated processes we can think of allowing the user to choose their metabolites on it. I will think now if there is a clean way to do this in the UI that does not overcomplicate later the R script. If I don't find it I will open an issue

R/apply_mapping.R
```diff
@@ -134,3 +134,13 @@ apply_mapping <- function(
...
 }
+
+#' Create METABFL Column Based on Selected Metabolites
+#' @param dataset A data frame containing a dataset with a PARAM (parameter) column.
+#' @param metabolites A character vector of parameter names representing metabolites.
+#' @importFrom dplyr mutate
+#' @returns The input dataset with an additional METABFL column indicating metabolite records ("Y")
+#' or non-metabolite records ("").
+create_metabfl <- function(dataset, metabolites) {
+  mutate(dataset, METABFL = ifelse(PARAM %in% metabolites, "Y", ""))
```
js3110: Question: does this need to be a function? since its only one line

Gero1999: I was very hesitant with this one and trying both... I just feel that this way looked more clean for the script, but what do you think?

PR #710, opened 2025-10-15
Enhancement: swap filters and column mapping

### LINKED ISSUE #568: Enhancement: Swap Filters and Column mapping
### Description
Filter should be after column mapping so that the user can go back and update them without having to resubmit the mapping
This means the filters need to be applied to the PKNCA object instead.

### Definition of Done
- [ ] Filters after column mapping
- [ ] Filter applied to whole dataset
- [ ] Ability to go and change filters in the middle of the workflow

js3110:
## Issue

Closes #568 

## Description

Swap filters and column mapping section so that the user can go back and filter without having to resubmit mapping. Makes the workflow easier.

## Definition of Done

- [x] Filters and column mapping swapped
- [x] Functionaility remains

## How to test

Try adding and removing filters at any step during process. Make sure all functionality is still intact.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.0.0.9175
+Version: 0.0.0.9176
 Authors@R: c(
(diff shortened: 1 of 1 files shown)


lucyaspridis (approved):
LGTM:)

PR #711, opened 2025-10-15
Bug: import rds with labels fails

### LINKED ISSUE #622: Bug: Import RDS does not work
### Description
Import RDS does not work

Error in [[: index out of bounds

Please test with RDS but also with all other formats accepted by the button

js3110:
## Issue

Closes #622 

## Description

Import rds when using an rds file that already contained labels was failing with error: `subscript out of bounds`  in generate col defs function. Due to incorrect extraction of label values. The function has been refined using imap so that it works for rds data.

## Definition of Done

- [x] .rds file with labels upload successful
- [x] Regualar data upload successful

## How to test

@Shaakon35 has .rds file to test, otherwise run through main app and download zip folder to get ADNCA rds file of dummydata. Test upload of different file types with and without labels.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.0.0.9172
+Version: 0.0.0.9173
 Authors@R: c(
     person("Ercan", "Suekuer", email = "ercan.suekuer@roche.com", role = "aut",
            comment = c(ORCID = "0009-0001-1626-1526")),


lucyaspridis (commented):
I noticed that if I download the zip file for the dummy data and then re-upload the .rds file for the adpc, the column DOSEU is missing. This is happening on main too. Do you know why this is happening?

lucyaspridis (approved):
Otherwise the bug fix works, LGTM!

Issue #696, opened 2025-10-10
Enhancement: Rework individual and mean exploration plots


js3110:
### Description
Individual and mean plots functions are similar but formatted differently, when they could be aligned to the same structure and format, with some slight deviations. Overall base structure should be the same in the function and in the UI.

### Definition of Done
The requirements for the feature to be complete:
- [x] Harmonisation of individual and mean plots UI
- [ ] Cleaner and harmonised individual and mean plotting functions

Issue #675, opened 2025-10-02
Bug:  Individual plots crash if multiple PCSPEC selected


js3110:
### Description
If > 1 PCSPEC selected in individual plots, the app crashes

### To reproduce
Exploration -> Individual plots -> Select >1 PCSPEC

### Expected behaviour
No crash.
App should automatically group by PCSPEC if >1 selected
App should automatically group by NCA_Profile if by dose profile selected
App should automatically group by PARAM is >1 param selected

PR #694, opened 2025-10-10
Refactor: Move creation of PKNCA obj & update qc_plot_server

Gero1999:
## Issue

Closes #632

Relates to PRs: #688 #641

## Description
Moved the creation of the PKNCA object to the data tab

## Definition of Done
- [x] move creation of PKNCA object to app.R
- [x] update qc plot server function

## How to test
Check the App keeps working well in all tabs

## Contributor checklist
- [ ] Code passes lintr checks
- [ ] Code passes all unit tests
- [ ] New logic covered by unit tests
- [ ] New logic is documented
- [ ] Package version is incremented

## Notes to reviewer

Anything that the reviewer should know before tacking the pull request?

PR #688, opened 2025-10-08
Allow exportation of dose escalation slides (pptx / quarto)

Gero1999:
## Issue

Closes #579
Closes #685
Contributes to ##632 (working on it parallely)

## Description
Adapted our exploration plotting functions to accept more PKNCA-related inputs. Then I used them for generating an object that stores the outputs differentiated by groups. Currently all is set by default options but the purpose is that an interface for better customization will be implemented in #638.

The outputs are generated in quarto (prepared for html) and pptx so the user can keep customizing for their preferred format. 

## Definition of Done
#579
- [ ] Should include summary statistics, mean plots and individual plots seggregated by groups
- [ ] Would be good to include individual parameters & plots for each subject within the group
- [ ] Capacity to include it in the ZIP folder (pptx and if possible qmd)
- ~~Options to select output configurations, blocked by https://github.com/pharmaverse/aNCA/issues/638~~

#685
- [ ] Remove log10 label (meanplot, individual plot)

## How to test
- Map data > Run NCA > Download ZIP > Check `presentations/` folder > pptx / qmd files
- Map data > Exploration >  Individual plots > Checkbox log-scale > Y label should not change 
-  Map data > Exploration > Mean plots > Checkbox log-scale > Y label should not change 

## Contributor checklist
- [ ] Code passes lintr checks
- [ ] Code passes all unit tests
- [ ] New logic covered by unit tests
- [ ] New logic is documented
- [ ] Package version is incremented

PR #712, opened 2025-10-16
Enhancement: Harmonise mean and ind plots

### LINKED ISSUE #701: Enhancement: Harmonise UI for mean and individual plots
### Description
Individual and mean plots should have the same UI template, with addition of SD and CI for mean plots.

### Definition of Done
- [ ] Ind and Mean plots use same UI module

js3110:
## Issue

Closes #701 #810 as sub issue of #696 

## Description

Harmonise individual and mean plots so they use the same UI module and same plotting function, with just different data processing.

## Definition of Done

- [x] Same UI and functions for mean and individual plots
- [x] Features work as before

## How to test

Go to exploration section and view plots.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

The plot does not render when >1 PCSPEC are selected due to multiple units. This will be solved in #675


js3110:
Will need to either wait until #688 and #694 is merged, or harmonise code while waiting.

R/g_lineplot.R
```diff
@@ -0,0 +1,210 @@
...
+    .add_threshold(show_threshold, threshold_value),
+    .add_dose_lines(show_dose, dose_data, facet_by),
+    .add_mean_layers(
+      is_mean_plot,
+      show_sd_min,
+      show_sd_max,
+      show_ci,
+      colorby_var,
+      y_var
+    )
```
Gero1999: I really like how clean this ends being when reading the function. But I think would be even nicer if you could reuse these helper functions for the other plots. Have you considered putting all explore plotting functions in 1 file and reuse the common options? This will help a lot with consistency and for people to understand the code

R/g_lineplot.R
```diff
@@ -0,0 +1,210 @@
...
+#' @import ggplot2
+#' @import dplyr
+#' @export
+g_lineplot <- function(data,
+                       x_var,
+                       y_var,
+                       group_var,
+                       colorby_var,
+                       facet_by = NULL,
+                       yaxis_scale = "lin",
```
Gero1999: as we only have two possible values I would prefer this argument named something like `yaxis_log10 = FALSE`. The same with the other functions

R/plot_creation.R
```diff
@@ -0,0 +1,178 @@
...
+#' @returns A `ggplot` object.
+#'
+create_meanplot <- function(data,
+                            selected_analytes,
+                            selected_pcspec,
+                            cycle = NULL,
+                            colorby_var,
+                            facet_by = NULL,
+                            yaxis_scale = "lin",
+                            time_scale = "All Time",
```
Gero1999: same comment as before. no `cycle` + `time_scale` but just `selected_cycles` and all are represented

R/g_lineplot.R
js3110: Sounds like a nice idea, lets create a separate issue :)

js3110: created issue #782

js3110:
thanks @Gero1999 , as we discussed, the issue with multiple PCSPEC is to be resolved in #675 
The confidence intervals I also am not sure, but again its not related to this PR and is only because you can now do mean plots over All time not just dose profile, but the code for the CIs is the same so we can open another issue for this.

R/g_lineplot.R
```diff
@@ -0,0 +1,246 @@
...
+#' @param data A data.frame containing the data to be plotted. This should be
+#'   pre-processed by either `process_data_individual` or `process_data_mean`.
+#' @param x_var A character string specifying the column name for the x-axis.
+#' @param y_var A character string specifying the column name for the y-axis.
+#' @param group_var A character string specifying the column name used to group
+#'   the lines, ensuring `geom_line` connects the correct points.
+#' @param colorby_var A character vector specifying the column(s) from the original
+#'   dataset that are used to determine the color of the lines and points.
+#' @param facet_by A character vector of column names to facet the plot by.
+#'   Default is `NULL` for no faceting.
```
m-kolomanski: issue: The argument names do not follow the same schema, I would go with `group_by`, `color_by` and `facet_by`

R/g_lineplot.R
```diff
@@ -0,0 +1,246 @@
...
+#'  a logarithmic scale for the y-axis.
+#' @param show_threshold A logical value (`TRUE` or `FALSE`) indicating whether to
+#'   display a horizontal threshold line.
+#' @param threshold_value A numeric value for the y-intercept of the threshold line.
+#'   Only used if `show_threshold` is `TRUE`.
+#' @param show_dose A logical value (`TRUE` or `FALSE`) indicating whether to show
+#'   vertical lines at dosing times.
+#' @param dose_data An optional data.frame containing dosing information.
+#'   It is required if `show_dose` is `TRUE`. The data frame must contain the faceting variables
+#'   (if any are used) and a `TIME_DOSE` column for the x-intercepts.
```
m-kolomanski: issue: As above, we could just have `dose_data` and if it is `NULL`, then it it the same as having `show_dose = FALSE`. If the data is provided, we show it.

R/g_lineplot.R
```diff
@@ -0,0 +1,246 @@
...
+  } else {
+    x_lab <- paste0("Time [", unique(data$RRLTU), "]")
+    y_lab <- paste0("Concentration [", unique(data$AVALU), "]")
+    title <- "PK Concentration - Time Profile"
+  }
+
+  # --- Tooltip Construction ---
+  if (nrow(data) == 0) {
+    data$tooltip_text <- character(0)
+  }
```
m-kolomanski: issue: I think this is redundant. If `!is.null(tooltip_vars)`, then it will get overwritten anyway, else doing this is equivalent:

```R
data$tooltip_text <- rep(NA_character_, nrow(data))
```

So I do not think this part does anything.

R/g_lineplot.R
```diff
@@ -0,0 +1,246 @@
...
+      show_sd_max,
+      show_ci,
+      colorby_var,
+      y_var,
+      x_var,
+      group_var
+    )
+  )
+
+  return(plt + optional_layers)
```
m-kolomanski: style: implicit return
```suggestion
  plt + optional_layers
```

tests/testthat/test-g_lineplot.R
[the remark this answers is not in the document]
js3110: i think we decided to avoid using snapshots where possible because they were causing issues with the CI/CD workflows and peoples branches for some reason, but I may be wrong

inst/shiny/modules/tab_explore.R
```diff
@@ -151,259 +36,122 @@ tab_explore_server <- function(id, pknca_data, extra_group_vars) {
...
+        selected_analytes = individual_inputs()$param,
+        selected_pcspec = individual_inputs()$pcspec,
+        colorby_var = individual_inputs()$colorby,
+        ylog_scale = individual_inputs()$log,
+        profiles_selected = individual_inputs()$profiles,
+        facet_by = individual_inputs()$facetby,
+        show_threshold = individual_inputs()$show_threshold,
+        threshold_value = individual_inputs()$threshold_value,
+        show_dose = individual_inputs()$show_dose,
+        palette = master_palettes_list()
```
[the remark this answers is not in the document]
js3110: Is this still useful after splitting back to data processing and then plotting? would require a bit of post processing to sepearate the different arguments for the different functions so I'm not sure the code would be much cleaner.. take a look at my updates and lmk if you still think this is suggested! :)

js3110:
@Gero1999 I've updated now so that the lineplots module takes a PKNCAobject as input, and so now the grouping is only the PKNCA groups and the additional_grouping vars, to avoid unecessary grouping column options

R/g_lineplot.R
```diff
@@ -0,0 +1,262 @@
...
+  } else {
+    data$tooltip_text <- rep(NA_character_, nrow(data))
+  }
+
+  # Create color var for aesthetic mapping
+  plot_data <- data %>%
+    mutate(
+      color_var = interaction(!!!syms(color_by), sep = ", ")
+    ) %>%
+    arrange(.data[[x_var]])
```
m-kolomanski: issue(non-blocking): Use `!!sym()`. Why the sudden switch away from using tidy evaluation? You are using `syms()` just the line above.
```suggestion
    arrange(!!sym(x_var))
```

R/plot_data_creation.R
```diff
@@ -0,0 +1,157 @@
...
+  if (isTRUE(ylog_scale)) {
+    summarised_data <- summarised_data %>%
+      filter(Mean > 0)
+  }
+
+
+  list(
+    summarised_data = summarised_data,
+    time_col = time_col
+  )
```
m-kolomanski: issue: I don't like this return, functions in principle should be responsible for a single thing, do one thing and return one thing. Here it has two responsibilities - 1) processes data for plotting and 2) devises what the time column is. Especially that the second "feature" is really just one line, I think we are needlessly complicating the return value and the calling function (`tab_explore` module) could just do this on its own instead of relying in the output here.

PR #723, opened 2025-10-21
Chore:  update workflow to setup-r-dependencies

### LINKED ISSUE #506: Error: Workflow crash due to Ubuntu version for staged-dependencies-action
### Description
Ubuntu version causing checks to fail, specifically in `staged-dependencies-action` . After discussing with pharmaverse slack, it seems that staged dependencies is deprecated and not maintained, so it has been suggested to migrate to https://github.com/insightsengineering/setup-r-dependencies instead.

Example from NEST migration to assist:
https://github.com/insightsengineering/nestdevs-tasks/issues/65

[workflow-issue](https://github.com/pharmaverse/aNCA/actions/runs/18647776121/job/53158644919#step:8:54)

js3110:
## Issue

Closes #506 

## Description

Due to deprecation of `insightsengineering/staged-dependences-action` we are moving the workflows to use `setup-r-dependencies` instead. Here I have updated the YAML files to use `setup-r-dependencies`

I had to specify extra packages roxygen for manuals, devtools for tests, and cyclocomp for lintr.
I had to update RoxygenNote version so it was compatible with the new workflow

## Definition of Done

- [x] Checks pass
- [x] Setup-r-dependencies workflow working

## How to test

Check that the checks are all passing. You can also see that the code coverage check has also passed in Actions -> Code Coverage.

## Contributor checklist
- [x] New logic is documented

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.0.0.9163
+Version: 0.0.0.9164
 Authors@R: c(
     person("Ercan", "Suekuer", email = "ercan.suekuer@roche.com", role = "aut",
            comment = c(ORCID = "0009-0001-1626-1526")),


js3110:
> LGTM, only noticed this warning: `WARNING: ignoring environment value of R_HOME` and also that lintr-check does not provide a return message (i.e, "No lints found. ")

From what I can tell, the warning is just a log warning because R_HOME is set in the admiral container, and the setup-r-dependencies is just ignoring this. If we find out later that the warning is a problem then we can open a new issue :)

The workflow doesn't have anything set up to give the "No lints found" message. I checked and it still does this locally. But in checks its only if it finds lints, so I think thats also ok :)

PR #742, opened 2025-10-27
Fix: FREXINT (fe) only calculated if RCAMINT (ae) included in excretion analysis (#669)

### LINKED ISSUE #669: Bug: FREXINT (fe) only calculated if RCAMINT (ae) included in excretion analysis
### Description
In excretion analysis tab, if `FREXINT `is selected but `RCAMINT `is not, then the `FREXINT `will be = 0. Need to ensure that `FREXINT `can be calculated correctly whether `RCAMINT `is selected or not (maybe filter after results are calculated?)

### To reproduce
Using preclinical excretion dataset to calculate `FREXINT `without `RCAMINT`.

### Expected behaviour
Parameters should be able to be calculated independently of one another.

Gero1999:
## Issue
Closes #669

## Description
Temporary fix defining dependency of FREXINT and RCAMINT

## Definition of Done
- [ ] Parameters should be able to be calculated independently of one another.

## How to test
Using preclinical excretion dataset to calculate `FREXINT `without `RCAMINT` produces same results as with it.  Make sure to select PCSPEC=URINE in the setup.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer
AE will show up in the pivoted results as well (not in parameter datasets though). This needs another fix that involves other parameters as well and can become trickier.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.0.0.9168
+Version: 0.0.0.9169
 Authors@R: c(
     person("Ercan", "Suekuer", email = "ercan.suekuer@roche.com", role = "aut",
            comment = c(ORCID = "0009-0001-1626-1526")),


js3110 (approved):
LGTM! thanks

Issue #481, opened 2025-06-19
Collaboration: Things to consider for next PKNCA update


Gero1999:
### Description
Here is a list of things that may be needed to be consider for changes or new features in the App when PKCNA 0.12 is released in CRAN:

### 0.12.1
- [x] lambda.z.time.last is a new calculated parameter (add it in `pknca_cdisc_terms`) ✅ 
- [x] group_vars consider "start" & "end" ✅ 
- [x] AUCPEO criteria better considered  ✅  >> Use custom
- [x] perhaps exclusion flags improvements are included ✅ 
- [x] FE units are good, so we can have it in our App  ✅ 

### 0.12.2 ⏭️ 
- [ ] units based on columns are automatically created ❓ 
- [ ] VOLPK, renal clearance dose normalized ❓ 
- [ ] CORRXY ❓ 
- [ ] possibility for SDTM column transformation to the output ❓ 
- [ ] parameter ratios are done ❓ 
- [ ] units in output follow CDISC standards (like `fraction > fraction for 1`, `unitless > ""` ) ❓

PR #745, opened 2025-10-27
Add excretion rate parameters (#201)

### LINKED ISSUE #201: Enhancement: New parameter functions for excretion parameters
### Description
At the moment, the app runs well with plasma and serum concentrations in NCA.
We want to ensure the user can also calculate parameters based on Urine amounts.
For some of these, the data needs to have an end time column. The user should receive a pop up window that asks them to map actual and nominal end time for the data, if they want to caluclate these params.

Parameters (not yet in pknca):
- [ ] Accumulation index: "ACCI", "Accumulation Index",
- [ ] Tmax excretion rate: "ERTMAX", "Time of Max Excretion Rate"
- [ ] Max excretion rate: "ERMAX", "Max Excretion Rate" ( = Amount excreted / Time (end - start))
- [x] Urine volume: "VOLPK", "Sum of Urine Vol" (this comes from the columns VOLUME, VOLUMEU if available in ADPC, would be nice to propose to the user VOLPK only if VOLUME and VOLUMEU are in the dataset)
- [ ] Midpoint last:     "MIDPTLST", "Midpoint of Collection Interval"
- [ ] Renal CL Norm by Dose: "RENALCLD", "Renal CL Norm by Dose", `Next PKNCA version (?)`

### Definition of Done
- [ ] Dataset including urine data can be uploaded and run through NCA with no issues
- [ ] Calculation of additional parameters
- [ ] Results shown in NCA table (and Non-NCA data also in Excretion Analysis section)

Gero1999:
## Issue

Closes #201

## Description
- [x] Allows calculation of Excretion Rate parameters. Written functions for each of them using their CDSIC nomenclature & added them in `metadata_parameters.rds`
- [x] Added an end time column for the collection in the mapping for their calculation. 
- Changed `example-ADNCA.csv` to prevent collapsing sampling times (AFRLT and AEFRLT) of a subject's urine samples
- Accepted incoming changes from `metadata_nca_parameters` and included ERTLST and ERTMAX. Manually, both of them were specified as `can_metabolite = TRUE` 

## Definition of Done
Add excretion parameters (not yet in pknca):
- [x] Tmax excretion rate: "ERTMAX", "Time of Max Excretion Rate"
~~ Max excretion rate: "ERMAX", "Max Excretion Rate" ( = Amount excreted / Time (end - start)) ~~ >> Will be well integrated once PKNCA new versions arrives
- [x] ~~Midpoint last:     "MIDPTLST", "Midpoint of Collection Interval"~~ `ERTLST`, "Midpoint of Interval of Last Nonzero ER"

## How to test

Use a dataset with urine data (`tests/testthat/data/dummy_sm_data.csv`). Map `PCSPEC = "URINE"`

Request parameter calculation of `ERTMAX`, `ERTLST` and:
1) Check that their values are present in all expected outputs and input widgets (boxplot, pivoted_results...)
2) Check that values are indeed correct (scientist job!) ⚠️ 

## Contributor checklist
- [x] New logic is documented
- [x] Added to NEWS

## Notes to reviewer
ERMAX was not included in this PR directly. It has too many dependencies with internal functions from PKNCA, so it will get very complicated to incorporate all those functions updated inside this PR. We will need to test at the time of the release (#481) if everything is correctly integrated for it as well.


Shaakon35 (approved):
Works for ERTLST, for ERTMAX it's partly empty and partly equal to 0, can you check if that is correct, anyways, we would need the review from Scientists on this one.
image

js3110 (approved):
Worked for me on a preclinical dataset!

js3110 (changes requested):
Data review: comparing the results with other software i found that the results do not match.
For example the Max Rate is at the sample with AFRLT 0 and AEFRLT 3, so the ERTMAX should be 1.5, but is 0 in aNCA.
So the calculations are not correct.

js3110:
@Gero1999 I tested again with both datasets and the results are matching now ! 👏 
What was causing the times to collapse? was it due to having <0 start times? Its not completely clear from the commit.

js3110 (approved):
Nice, I am happy :)

PR #747, opened 2025-10-27
Bug: Ratio calculation failing for if-needed

### LINKED ISSUE #730: Bug: ratio calculation failing for if-needed
### Description
For preclinical test data, ratio calculation results in an error if `aggregate_subjects` is set to 'if-needed'. In the function `calculate_ratio_app`, the variable `match_cols`, which is a list of character vectors, has either length 1 or length 2. It can only contain 2 elements if `aggregate_subjects` is set to 'if-needed'. The ratios are calculated within a loop and the results are bound together with `bind_rows`. In this specific case, the first iteration (not aggregated by USUBJID) results in an empty dataset. The issue arises from the fact that in the empty ratio table (first iteration), PPORRESU has type `logical` and in the second one, PPORRESU has type `character` (as expected), which makes bind_rows fail.

### To reproduce
Using tests/testthat/data/adnca_dummy_sm_dataset.csv, choose any parameters for which to compute a ratio and set aggregate_subjects to "if-needed".

### Expected behaviour
In the empty dataset, PPORRESU should not have type `logical`.

lucyaspridis:
## Issue

Closes #730

## Description

For preclinical test data, ratio calculation resulted in an error if aggregate_subjects was set to 'if-needed'. In the function calculate_ratio_app, the variable match_cols, which is a list of character vectors, has either length 1 or length 2. It can only contain 2 elements if aggregate_subjects is set to 'if-needed'. The ratios are calculated within a loop and the results are bound together with bind_rows. The first iteration resulted in an empty dataset with PPORRESU having the wrong type (logical instead of character), making the binding impossible. 

This bug was fixed by adding an early return in `calculate_ratios.data.frame` for empty datasets, avoiding modifications of variable types in empty ratio tables caused by mutate calls later in the function.

## Definition of Done

- [x] No crashing for `aggregate_subject = if-needed` ratio calculations
- [x] Ratio calculations work as they used to for other `aggregate_subjects` options

## How to test

1. Run app loading the preclinical dummy dataset `tests/testthat/data/adnca_dummy_sm_dataset.csv`
2. Choose a parameter and ref&test groups and set `aggregate_subjects = if_needed`. The NCA should not crash and the calculation should be done for aggregated subjects. 
3. Repeat for other `aggregate_subjects` options and make sure the app works as before (warning because of empty table for `no`, calculation working correctly for `yes`.

## Contributor checklist
- [x] New logic is documented

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.0.0.9176
+Version: 0.0.0.9177
 Authors@R: c(
     person("Ercan", "Suekuer", email = "ercan.suekuer@roche.com", role = "aut",
            comment = c(ORCID = "0009-0001-1626-1526")),


js3110 (approved):
LGTM! Thanks for investigating this :)

Shaakon35 (approved):

R/ratio_calculations.R
```diff
@@ -215,6 +236,12 @@ calculate_ratios.data.frame <- function(
         ifelse(n > 1, paste0("RA", test_parameter, " (mean)"), paste0("RA", test_parameter))
       }
     ) %>%
+    # Make sure all basic columns are still character (even when empty)
```
Gero1999: nitpick: consider adding the issue reference so we know why this "odd" part of the code was done
```suggestion
    # Make sure all basic columns are still character (even when empty #730)
```

R/ratio_calculations.R
```diff
@@ -151,6 +151,27 @@ calculate_ratios.data.frame <- function(
...
+
+        # Ensure PPORRESU and PSSTRESU have the right type
+        PPORRESU = if ("PPORRESU" %in% names(.)) character(0) else NULL,
+        PPSTRESU = if ("PPSTRESU" %in% names(.)) character(0) else NULL
+      ) %>%
+      select(any_of(c(names(df_test), "PPANMETH")))
+
+    return(out_df)
+  }
+
```
Gero1999: I think you might not need this part if you follow the `as.character()` strategy 

```suggestion
```

tests/testthat/test-ratio_calculations.R
```diff
@@ -244,4 +244,52 @@ describe("calculate_ratios", {
     expect_equal(ratios_df$PPSTRES, rep(c(2 / 3), 4))
     expect_true(all(grepl("RACMAX", ratios_df$PPTESTCD)))
   })
+
+  it("handles early exit with no matches (data.frame) and returns correct structure", {
```
Gero1999: nitpick: I guess now we don't do early exit now (?) depends on what you decided before
```suggestion
  it("handles no test and reference matches (data.frame) and returns correct structure", {
```

lucyaspridis:
@Gero1999 no worries, you're right:)

Gero1999 (approved):
LPTM, thanks! 👌🏽

Issue #735, opened 2025-10-23
Bug: Refactor logic in e2e tests


js3110:
### Description
The e2e test workflow is not working as expected, and failing for a couple of PRs despite the workflow working, notably #687 and #664 

Most of the devs are unable to run the tests locally, which makes it difficult to debug.

### Expected behaviour
e2e tests that work and check the correct steps at the appropriate time, and work for all users so that developers are able to debug.

PR #753, opened 2025-10-31
Skip e2e tests (temporarely until they work again)

Gero1999:
## Issue
This is not for closing an issue but to solve the issues associated with these column mapping branches (#710,  #687, #568). Let's keep this PR linked with #735  in case at some point can be helpful to know this

## Description
Just skipping e2e tests in testthat


js3110 (approved):
great!

Issue #752, opened 2025-10-31
Enhancement: Exclusion flags mapping


Gero1999:
### Description
Automating column mapping as much as possible when relevant to include these flags and then use PKNCA exclusion methods to perform them. 

`export_cdisc` will also have to include these columns cleverly

### Definition of Done
The requirements for the feature to be complete:
- [x] NCAwXRS exclusion flags in input and CDISC outputs
- [ ] PKwXRSN (summary statistics) flags in input and CDISC outputs

Blocked by #724

PR #765, opened 2025-11-07
Change filtering duplicates for exclusions integrated with PKNCA

### LINKED ISSUE #724: Enhancement: separate duplicates filtering (DFLAG) behavior from mapping / pknca creation
### Description
For a better R script we need to separate the filtering of the function from `apply_mapping()` and `PKNCA_create_adnca()`. Instead we need to make a simple/clear function after `apply_mapping` that takes as input the `adnca_dataset` + `filtered_rows_df`

Additionally we can use the chance to change the name of the column (DFLAG) to a standard ~(NCA1XRS = "Duplicated sample")~  (DTYPE = "TIME DUPLICATE")

### Definition of Done
- [ ] Separate indirect filtering behaviour from apply_mapping. Make this explicit for the user / script
- [ ] Instead of `DFLAG=TRUE` make the flag for duplicated records ~(NCA1XRS = "Duplicated sample")~  (DTYPE = "TIME DUPLICATE")
- [ ] Integrate better PKNCA if possible to handle this

### Important Note
Start developing this parallel to `681-feat-flexibe-value-col-mapping` if still was not merged
Additionally, please for `filtered_rows_df` use as few columns as needed

Next implementations in #752 will derive from DTYPE a new NCAwXRS

Gero1999:
## Issue

Closes #724

## Description
- Integrates the duplicate records exclusion with PKNCA

Separates the filtering of the function from `apply_mapping()` and `PKNCA_create_adnca()`. Instead we need to make a simple/clear function after `apply_mapping` that takes as input the `adnca_dataset` + `filtered_rows_df`

Additionally we can use the chance to change the name of the column (DFLAG) to a standard (NCA1XRS = "Duplicated sample")

### Definition of Done
- [x] make a separate function filter_duplicated_samples(adnca_dataset, filtered_rows_df)
- [x] Instead of `DFLAG=TRUE` make the flag for duplicated records ~(NCA1XRS = "Duplicated sample")~  (DTYPE = "TIME DUPLICATE")
- [x] Integrate better PKNCA if possible to handle this

## How to test
-  **Excluding points through PKNCA works well:** Use DummyRO and see if duplicates are handled as before, results are the same
 - **Modal message handling works well**. Make sure that if you don't select the points, it asks you to do so. Play with it
 - ADNCA in review do not create other new columns apart from DTYPE

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer
DTYPE indicates sample duplicates. For now the exclude column is using DTYPE, but after we do #752 DTYPE will be used to derive NCAwXRS. `w`  may vary on the number of user's mapped columns.


js3110 (commented):
Hi, looks good and the back end processing is nice!
Just a question because I am slightly confused

is DTYPE ADNCA standard column?
You said that the NCA1XRS is the standard column- for which dataset? and how do I see it because currently I don't see it being created anywhere.

Would be good if you could provide a bit more info in the how to test!

js3110 (approved):
LGTM :)

Shaakon35 (approved):

PR #777, opened 2025-11-13
Include new data types in default App dataset

### LINKED ISSUE #750: Enhancement: Add all types of general case data
### Description
Add new columns:
- VOLUME, VOLUMEU (with PCSPEC = Urine)
- WEIGHT, WEIGHTU

Consider including also duplicate rows

### Definition of Done
- [ ] duplicated rows 
- [ ] PCSPEC rows with Urine
- [ ] Different administration forms

Gero1999:
## Issue

Closes #750 

## Description
The data has 10 subjects with 2 analytes (drug + metabolite), each with 8 doses:

- First 5 subjects have infusion (3h) 
- First 3 subjects also have URINE samples (another PCPSEC), which constitutes a different concentration unit (AVALU)
- Odd subjects have a dose administration of 5g while even subjects have a dose administration of 10g. 

Race and sex  have two categories equally distributed among the subjects. They have small effects on the treatment.

## Definition of Done
- [x] ~duplicated rows~ I actually think this can become pretty bothering if we have to over and over pass the modal window
- [x] PCSPEC rows with Urine
- [x]  Different administration forms

## How to test
- Inspect the dataset (now is much smaller) and make sure everything looks alright regarding bolus, infusion doses. Also check the urine samples and metabolite data.
-  Check the class of each column
- Make sure we don't miss any (really) relevant column 

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer
The datasets are only removed from the repo. You don't need to remove them locally in case in the future we miss something from them (i.e, duplicate rows).


js3110:
> It's easier to have one so people can see the input and not get confused. Also for the CRAN submissions will make things easier

I would prefer to keep them but maybe we can put them in a separate folder?
As they are extremely useful for development testing. 
Maybe add to tests folder?

js3110 (changes requested):
Hi, a couple things:

1. Would still be nice to have >1 PCSPEC in the data.
2. Also nice to have different units for metabolite and analyte as this often causes errors so better to find them now
3. Would be nice if METABLFL is already populated so that we don't have to map them every time.

I tried with the data but the app crashed when I tried to run NCA on DOSE 6.
With message:
```

Caused by error in `assert_time()`:
! Assertion on 'time' failed: Contains duplicated values, position 2.
  2: shiny::runApp
  1: run_app [C:/Users/spinnerj/Documents/aNCA/R/run_app.R#39]
Warning: Error in mutate: ℹ In argument: `is.possible.c0.logslope = !...`.
ℹ In group 57: `STUDYID = "XX01"`, `PCSPEC = "SERUM"`, `DOSETRT = "DrugA"`, `USUBJID = "XX01-04"`, `PARAM = "DrugA"`, `DOSNOA = 2`, `start = 240.06`, `end = 250.085`, `type_interval = "main"`.
```

I had a look at the DOSNOA since you mentioned there are issues, and from what I can see, the ARRLT and AFRLT are not always consistent within doses. In the DOSNOA derivation we have a buffer of 0.02 for rounding errors, but I noticed there are a couple of mismatches of >0.03, which causes two different dose times to be calculated and therefore 2 different DOSNOA.

Would be good to create an issue to stop this from crashing the app and give the user an error, but either way for the data this is not correct. I didn't check all of them for obvious reasons, but i noticed two dose times in SUBJ 2 DOSE 2, and SUBJ 1 DOSE 6 at least.
Maybe this is the cause of the issues you mentioned?

Shaakon35 (changes requested):
Data looks very good!!

A few comments:
- Don't you want to create TRTRINT? and TRTRINTU
- Add AGE and AGEU (supplemental)
- Adding TRT01A and TRT01P for the TLGs might be wise (supplemental), can you set it as "Arm 1 - 5 mg Bolus", "Arm 2 - 5 mg Infusion", "Arm 3 - 10 mg Bolus", "Arm 4 - 10 mg Infusion"
- DOSETRT is not correct, it should be only DrugA
image


Can you modify the data so that when computing the pk parameters, the purple color appears? 

Otherwise, all good :)

Shaakon35:
> @Shaakon35 notes:
> 
> * I can add a TRTINT, `TRTINTU = RRLTU` (we made this and specify it to simplify)
> * Sure `AGE` and `AGEU`. I personally prefer to keep as simple as possible the dataset new users but I can add them
> * This is fine, the example you sent is a dummy case that only specifies the drug. The purpose of `DOSETRT` is to reflect the `TRTA/TRTP` (they are all treatment regimens, not compounds)

Yes I see you wanted to simplify, but the best would be to have the maximum variables that are available in mapping, TRTINT, AGE, etc.

I would name TRT01A, not TRTA: see https://www.mycsg.in/adam.php?area=adam&concept=ADSL&lesson=ADaM_ADSL_L1101#SPEC

For DOSETRT, I am pretty sure that it's derived from EXTRT, nothing to do with TRT01A or ACTARM, where did you get this info?

Shaakon35 (approved):
All good, if you think it's useful, you could add:

- TRTRINTU
- Modify TRT01A to have "Arm 1 - 5 mg Bolus", "Arm 2 - 5 mg Infusion", "Arm 3 - 10 mg Bolus", "Arm 4 - 10 mg Infusion"
- Modify the data so that we have a purple line (MISSING) when runnning the NCA - this might be challenging

js3110 (approved):
LGTM :)

PR #781, opened 2025-11-14
fix: remove RENALCL parameters from Parameter Selection (wrongly calculated in PKNCA)

### LINKED ISSUE #780: Bug: Remove inacurate PKNCA parameters
### Description
There are some parameters that are currently not 100% well calculated by PKNCA:

- RENALCL parameters use the same profile for the parameter ratio. 

### To reproduce
Urine dataset input > Map data > Select PCSPEC only URINE or only other > Select in Parameters RENALCL > Run NCA > Check result values

### Expected behaviour
All parameters should be accurate. Otherwise they should be removed

Gero1999:
## Issue

Closes #780 

## Description
We remove RENALCL parameters from `metadata_nca_parameters` as they are wrongly calculated in PKNCA

- RENALCL is no longer calcualted through PKNCA, so it is not available in `Parameter Selection`. 
- Parameter Ratios will recognise it if introduced

## Definition of Done
- [x] All RENALCL parameters do not appear in the parameters input

## How to test
`Urine dataset > Select Urine PCSPEC > In parameter selection no RENALCL is available`
In addition, compute it through parameter ratios:

`TestGroups = PCSPEC: URINE`, `TestParameter = RCAMINT`, `RefParameter = AUCLST/AUCIFO/AUCIFP`

Remember to select multiple doses (ATPTREF) so there are results available for comparison

Make sure the label appears as well in the results

## Contributor checklist
- [ ] New logic is documented


m-kolomanski (approved):

inst/shiny/modules/tab_nca/setup/ratio_calculations_table.R
```diff
@@ -311,6 +311,9 @@ ratios_table_server <- function(
       PPTESTCD = case_when(
         startsWith(RefGroups, analyte_col) ~ paste0("MR", TestParameter),
         startsWith(RefGroups, profile_col) ~ paste0("AR", TestParameter),
+        startsWith(TestGroups, paste0(pcspec_col, ": URINE")) &
+          startsWith(TestParameter, "RCAMINT") &
+          startsWith(RefParameter, "AUC") ~ "RENALCL",
```
js3110:
```suggestion
        startsWith(TestGroups, paste0(pcspec_col, ": URINE")) &
          startsWith(TestParameter, "RCAMINT") &
          startsWith(RefParameter, "AUC") ~ "RENALCL",
        startsWith(RefGroups, analyte_col) ~ paste0("MR", TestParameter),
        startsWith(RefGroups, profile_col) ~ paste0("AR", TestParameter),
```

js3110 (commented):
If i rty to create RENALCL via the ratios it doesn't automatically call it RENALCL:

image


Unless I change the Ref group to PCSPEC: URINE
image


Which obvs isnt correct. From the code I think this part needs to be specified above the others in the `case_when`? as otherwise the `AR, TestParameter` part takes priority

Gero1999:
Hmm let me check I don't think the problem is there for what you described

Gero1999:
Should be fine now :)

js3110 (approved):
Awesome Sauce!!!

PR #789, opened 2025-11-28
Create R script to replicate App outputs (#467)

### LINKED ISSUE #467: Enhancement: R Script output for NCA workflow
### Description
For reporting and filing, it would be good if we could create an R Script output that can be run independently of the app, that will provide the same NCA outputs. From data manipulation to creation of the NCA results/ ADPP.

### Definition of Done
- [ ] R Script export
- [ ] Can be run independently and give same results as in app
- [ ] Using aNCA functions

Gero1999:
## Issue

Closes #467 

## Description
Implemented the script file. It was made in order to eventually also be used using a settings file

## Definition of Done

- [x] R Script export
- [x] Can be run independently and give same results as in app
- [x] Using aNCA functions

## How to test
Do your choices in the App and export the ZIP folder. In the subfolder find the `/code/session_script.R` and run it. Compare the script outputs with the result files in the exported ZIP

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer
- The tests for the function `calculate_ratios_app` will be included here #287 independently 
- The tests for the function `get_session_script_code` has probably to be also done as an e2e test. To indeed make sure that both session and script are matching

### DIFF
--- a/R/ratio_calculations.R
+++ b/R/ratio_calculations.R
@@ -85,30 +85,28 @@ multiple_matrix_ratios <- function(data, matrix_col, conc_col, units_col,
 #' @export
 #' @export
 calculate_ratios <- function(
-  data,
-  test_parameter,
-  ref_parameter = test_parameter,
-  match_cols,
-  ref_groups,
-  test_groups = NULL,
-  adjusting_factor = 1,
-  custom_pptestcd = NULL
-) {
+    data,
+    test_parameter,
+    ref_parameter = test_parameter,
+    match_cols,
+    ref_groups,
+    test_groups = NULL,
+    adjusting_factor = 1,
+    custom_pptestcd = NULL) {
   UseMethod("calculate_ratios", data)
 }
 #' @export
 
 #' @export
 calculate_ratios.data.frame <- function(
-  data,
-  test_parameter,
-  ref_parameter = test_parameter,
-  match_cols,
-  ref_groups,
-  test_groups = NULL,
-  adjusting_factor = 1,
-  custom_pptestcd = NULL
-) {
+    data,
+    test_parameter,
+    ref_parameter = test_parameter,
+    match_cols,
+    ref_groups,
+    test_groups = NULL,
+    adjusting_factor = 1,
+    custom_pptestcd = NULL) {
   if (!any(data$PPTESTCD == test_parameter)) {
     warning(
       paste0(
@@ -237,15 +235,14 @@ calculate_ratios.data.frame <- function(
 
 #' @export
 calculate_ratios.PKNCAresults <- function(
-  data,
-  test_parameter,
-  ref_parameter = test_parameter,
-  match_cols,
-  ref_groups,
-  test_groups = NULL,
-  adjusting_factor = 1,
-  custom_pptestcd = NULL
-) {
+    data,
+    test_parameter,
+    ref_parameter = test_parameter,
+    match_cols,
+    ref_groups,
+    test_groups = NULL,
+    adjusting_factor = 1,
+    custom_pptestcd = NULL) {
   # Check if match_cols and ref_groups are valid group columns
   # Make checks on the input formats
   cols_used_for_ratios <- c(match_cols, names(ref_groups), names(test_groups))
@@ -332,15 +329,14 @@ calculate_table_ratios <- function(res, ratio_table) {
 #' @param custom_pptestcd Optional character. If provided, will be used as the PPTESTCD value.
 #' @returns A data.frame with the calculated ratios for the specified settings.
 calculate_ratio_app <- function(
-  res,
-  test_parameter,
-  ref_parameter = test_parameter,
-  test_group = "(all other levels)",
-  ref_group = "PARAM: Analyte01",
-  aggregate_subject = "no",
-  adjusting_factor = 1,
-  custom_pptestcd = NULL
-) {
+    res,
+    test_parameter,
+    ref_parameter = test_parameter,
+    test_group = "(all other levels)",
+    ref_group = "PARAM: Analyte01",
+    aggregate_subject = "no",
+    adjusting_factor = 1,
+    custom_pptestcd = NULL) {
   reference_colname <- gsub("(.*): (.*)", "\\1", ref_group)
   match_cols <- setdiff(unique(c(dplyr::group_vars(res), "start", "end")), reference_colname)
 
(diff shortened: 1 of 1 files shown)


R/get_session_script_code.R
```diff
@@ -0,0 +1,108 @@
...
+  # with a warning message
+  if (matches[1] == -1) {
+    script_lines <- strsplit(script, "\n")[[1]]
+    writeLines(script_lines, output_path)
+    warning(
+      "Please, the script template should have session$userData calls in order to work. ",
+      "Did you accidentally modify the original template?"
+    )
+    return(invisible(output_path))
+  }
```
m-kolomanski: suggestion: IMHO this should be a straight error or a warning that returns NULL stating that the function failed to parse the template - if the template was meddled with, I don't think we want to include it.

R/get_session_script_code.R
```diff
@@ -0,0 +1,108 @@
...
+
+  # Replace all matches with deparsed values in one pass
+  match_positions <- gregexpr(pattern, script, perl = TRUE)[[1]]
```
m-kolomanski: issue: We already define `matches` this is redundant.

R/get_session_script_code.R
```diff
@@ -0,0 +1,108 @@
...
+}
+
+# Helper to cleanly deparse an object (data.frame, list, etc.)
+clean_deparse <- function(obj, indent = 0) {
+  ind <- paste(rep("  ", indent), collapse = "")
+  if (is.data.frame(obj)) {
+    # Multi-line, indented, no trailing comma on last column
+    cols <- lapply(obj, function(col) {
+      d <- paste(deparse(col, width.cutoff = 500), collapse = "")
+      if (grepl("^c\\(", d)) d else paste0("c(", d, ")")
```
m-kolomanski: style: IMHO it is a bit hard to follow, something like this is less concise but I think a bit more readable:
```suggestion
      d <- col |>
        deparse(width.cutoff = 500) |>
        paste(collapse = "")
        
      if (!grepl("^c\\(", d)) {
        d <- paste0("c(", d, ")")
      }
      
      d
```

R/get_session_script_code.R
m-kolomanski: issue: We definitely need unit tests for `clean_deparse` to make sure we are doing things right.

m-kolomanski (changes requested):

js3110 (commented):
Its working! But in the pivoted_results, I have 4 rows for each results row- maybe something is going wrong with some merging at some point?

image

inst/shiny/www/templates/script_template.R
```diff
@@ -0,0 +1,118 @@
...
+    ),
+    silent = FALSE
+  ) %>%
+
+  # Derive METABFL column using PARAM metabolites
+  create_metabfl(mapping$Metabolites)
+
+## Setup NCA settings in the PKNCA object ########################
+auc_data <- session$userData$settings$partial_aucs
+units_table <- session$userData$final_units
```
js3110: I tried this with the multispec-data but the units table was so long that it wasn't fully expanded and therefore the ) at the end of the line wasn't found and so I couldn't run the code.
Is there a way to set a limit for how long the line of code is before it starts to overflow?

inst/shiny/www/templates/script_template.R
```diff
@@ -0,0 +1,125 @@
...
+###########################################################
+## Please, set your working directory to this file:
+# setwd("path/to/this/file/script_template.R")
+
+if (!require("aNCA")) install.packages("aNCA")
+library(aNCA)
+library(dplyr)
+
+# Load raw data #
+data_path <- "../data/data.rds"
```
js3110: Nitpick: Should add a comment here that makes it clear that this part needs to be modified by the user, or comment it out so its obvious

js3110 (approved):
LGTM, some non-blocking comments.

inst/shiny/www/templates/script_template.R
Gero1999: this path should already work if they open the file and don't change the directory. Otherwise they will need to change the directory as indicated in the previous comment

js3110: ah, i see, youre right. thanks!

js3110 (commented):
I am still getting this message every time I checkout your branch:

```
Warning message:
── Conflicts ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────── aNCA conflicts
──
✖ `calculate_table_ratios_app` masks `aNCA::calculate_table_ratios_app()`.
ℹ Did you accidentally source a file rather than using `load_all()`?
  Run `rm(list = c("calculate_table_ratios_app"))` to remove the conflicts. 
```
  
  the only way to solve is to clear my env before I run load_all()


In addition, I have a similar issue to @Shaakon35 : the PK Params in the r script version are more than in the output results. I investigated a little and I think its because in the app we filter out the params calculated due to dependencies but not requested, but in the rscript its not filtering them.

Gero1999:
> At the very least, if we want to implement it this way right now and then add the functionality of parsing a settings file instead of session object, we should have: more generic name, TODO comments

I created a TODO and the implementation will be developed in #835. The name/refactoring of this function I will be done there just for easiness.

Here, in this PR, I removed the redundant check, renamed `calculate_table_ratios_app` > `calculate_table_ratios` & refactored to use `apply`

Shaakon35 (approved):
Here we go!! Congratulations @Gero1999 👍

Shaakon35:
@Gero1999 Can you add in the README, in the Description bullet list:
- reproduce your NCA outside of the app, with the R script created

R/pivot_wider_pknca_results.R
```diff
@@ -170,3 +204,60 @@ pivot_wider_pknca_results <- function(myres) {
...
+  }, df[, mapping_cols], attrs, SIMPLIFY = FALSE))
+  df
+}
+
+.create_flags_for_profiles <- function(final_results, myres, flag_rules) {
+
+  # Add flaging columns in the pivoted results
+  applied_flags <- purrr::keep(flag_rules, function(x) x$is.checked)
+  flag_params <- names(flag_rules)
+  flag_thr <- sapply(flag_rules, FUN =  function(x) x$threshold)
```
m-kolomanski: nitpick:
```suggestion
  flag_thr <- sapply(flag_rules, FUN =  \(x) x$threshold)
```

also consider `purrr::map_chr()`

PR #664, opened 2025-09-26
Enhancement 436: Default parameter selections by study type (593, 616, 617)

js3110:
## Issue

Closes #436

## Description

Cumulative issues 593, 616, 617 (all approved and merged) and requesting to merge to main. Parameter selection finalised and working as expected.

## Definition of Done

- [ ] Default parameter selection by study type
- [ ] OPtion to select manually

## How to test

Check for bugs that may have occured when combining issues. Try different datasets and different study types and parameter selections

## Contributor checklist
- [ ] Code passes lintr checks
- [ ] Code passes all unit tests
- [ ] New logic covered by unit tests
- [ ] New logic is documented
- [ ] Package version is incremented

## Notes to reviewer

Anything that the reviewer should know before tacking the pull request?

PR #795, opened 2025-12-02
Parameter Selection 2

### LINKED ISSUE #436: Enhancement: set default parameter selections based on study type

lucyaspridis:
## Issue

Closes #436 

## Description

Rework of parameter selection, new solution suggested after parts of #664 were found not to work as expected.
Overview of issue:
Updating parameter selection so that rather than having the user select just one set of parameters, and all being calculated, they can select different parameters for the different study types detected.
The parameter selection table will by default select the relevant parameters for each study type, based on the parameter metadata which contains T or F values for parameters and study types.
The user can still manually select any parameters they want through virtual select inputs separated into tabs (one for each study type) and the settings upload should also overwrite the selections to fit the settings.

Note: files in the `R`, `man` and `tests` folder are identical to #664 (up to small changes made to the documentation due to merges from main).

The only difference between changes made in #664 and this branch is the process of parameter selection: 
Instead of through an interactive reactable, the user selects parameters per study type through Virtual Select widgets, in which parameters are aggregated by groups (Standard, Urine, Sparse). This updates a reactable that displays the parameters selected, only shown if selected for at least one study type.

## Definition of Done

- [x] Checks for different types of studies
- [x] Allow manual editing of these selections
- [x] Default parameter selections based on study types
- [x] Uploading of settings correctly updates selections
- [x] NCA runs fine with those parameters removed
- [x] Each study type is run with corresponding selected parameters

## How to test

Using dummy data and preclinical dummy data, play around with the selection of parameters, download and upload settings and check the parameters are correctly selected. Run NCA and check results.

## Contributor checklist

- [x] New logic is documented

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.1.0.9014
+Version: 0.1.0.9015
 Authors@R: c(
     person("Ercan", "Suekuer", email = "ercan.suekuer@roche.com", role = "aut",
            comment = c(ORCID = "0009-0001-1626-1526")),


Shaakon35 (approved):
It works well for me, very good job, we still need a couple of reviews:
- copilot
- @m-kolomanski 
- @js3110 and or @Gero1999

R/intervals.R
```diff
@@ -0,0 +1,232 @@
...
+  # 3. Iterate over selections to set new flags to TRUE
+  updated_intervals <- intervals_with_types
+  if (length(parameter_selections) > 0) {
+    for (study_type in names(parameter_selections)) {
+      params_for_type <- parameter_selections[[study_type]]
+      if (length(params_for_type) > 0) {
+        updated_intervals <- updated_intervals %>%
+          mutate(across(
+            .cols = all_of(params_for_type),
+            .fns = ~ if_else(type == study_type, TRUE, .)
```
m-kolomanski: issue(non-blocking): Please use anonymous functions instead of formulas if possible.
```suggestion
            .fns = \(x) if_else(type == study_type, TRUE, x)
```

Gero1999:
Nice work! The functionality seems good to me. Some things realized while checking this branch that might be worthy to open new issues:

- Regarding the functionality it might be worthy to consider for the future removing the collapsibles and just using the label titles of each input to differentiate them ("IV bolus parameters:", "metabolite IV bolus parameters:" ....) just to prevent redundancy

- It seems to "partially work" with parameter ratios (when the parameters are not there and then we do a ratio parameter it sends a warning on why was not calculated). The best in the future would be these parameter ratios to work even if the parameter was not computed

- It does not seem to work with the bioavailability widget (test/data/multispec-ADNCA.csv). Not only because the parameters are not included on it, but also because the App won't run NCA with the default selections.

inst/shiny/modules/tab_nca/setup/parameter_selection.R
```diff
@@ -39,6 +39,8 @@ parameter_selection_server <- function(id, processed_pknca_data, parameter_overr
   moduleServer(id, function(input, output, session) {
     ns <- session$ns
 
+    initialised_modules <- new.env()
```
m-kolomanski: praise: Nice solution, I would usually use `reactiveVal()` with a list of ids, but using a separate env is neater!

js3110 (approved):
One last time <3

PR #799, opened 2025-12-03
feat: integrate NCA exclusion flags (NCAwXRS) in the mapping

### LINKED ISSUE #752: Enhancement: Exclusion flags mapping

Gero1999:
## Issue

Contributes to #752 

## Description

Flags for NCA exclusion (NCAwXRS) are incorporated in the mapping as a selectInput. It allows multiple mapping and uses PKNCA system to do exclusions (that means the records won't be filtered, so the data is not changed)

The dummy data also has now two flag columns that are flagged with:

- `Invalid administration` for patients 2 & 4 in DOSES 1 & 2
- `No AVAL value` for a missing value that was manually added. Records at rows 100 & 233
- In SUBJ2 you can also find a case in which both exclusion reasons are happening

## Definition of Done
- [x] NCAwXRS exclusion flags in input and CDISC outputs (ADNCA)

## How to test

- Map data (default) > Run NCA > Check that patients 2 has no DOSE 2 results

-  Map data (without NCA1XRS) > Run NCA > Check that patients 2 has DOSE 2 results

- Check the ADNCA CDISC files contains the columns that were used in the mapping (if you don't include it in the mapping, they should not be)! And of course they always should use CDISC nomenclature, not the user's one

## Contributor checklist
- [x] New logic is documented


Shaakon35 (changes requested):
The zip file does not contain all the columns of the ADNCA dataset displayed in the app, could you please fix this? and have a look at ADPP as well please.

Otherwise, all good!!

Shaakon35 (approved):

js3110 (commented):
Looks like maybe some merges or smth are adding additional columns:
image

Otherwise looks good, I'll wait for this to be fixed then approve

Gero1999:
@js3110 the problem was in the dataset, now with the merge of the new ADNCA hopefully is fixed

js3110 (commented):
Thanks for the updates, but now it looks like all subjects have NCA results even if the entire profile has NCA1XRS. Is this correct? In the how to test you said it shouldn't have results but maybe its different with the new data

js3110 (approved):
lgtm!

Issue #167, opened 2025-01-20
Enhancement: Vertical/Horizontal lines for the TLGs - Graphs


Shaakon35:
### Description
What is the enhancement about, why should it be implemented?

We should be able to add x or y line to the plots

### Definition of Done
The requirements for the feature to be complete:
- [ ] needs a button to add 1 or multiple horizontal lines on the plot
- [ ] needs a button to define some text next to the horizontal line (by default: "threshold 1", "threshold 2" for the second line, etc.) 
- [ ] every new line has a different color
- [ ] Same buttons for vertical line

m-kolomanski:
Hey, a couple of questions for this:

- Does that apply to specific plot types or basically all of them?
- Do we want to allow the user to pick the color for each line?
- I do not quite understand the last point, would you mind explaining it further?

Shaakon35:
@m-kolomanski 
This applied to all plots yes.
No, the color will be automated

So the last point is about duplicating the predose point so that if we facet the plot by "cycle" (like on this plot where it's facet by dose). The last of point at 3H appears on the first plot, but also appears at 0H in the second plot.
@js3110 Will take care of this issue I guess

![Image](https://github.com/user-attachments/assets/f082c4b4-420b-4624-b447-a13cc7aac408)

js3110:
@m-kolomanski the last point is referring to the create_duplicates function that I have created in #177 , as this can be applied to other concentration-time plots.

PR #809, opened 2025-12-07
Closes #808 - enhancement/ increase test coverage ratio_calculations.R

### LINKED ISSUE #808: Enhancement:  Add tests to R/ratio_calculations.R to reach 100% coverage
Guidance: 

> So basically this function is pretended to be used in a PKNCAresult obj (i.e, `FIXTURE_PKNCA_RES$result`) or the data.frame inside of it (i.e, `FIXTURE_PKNCA_RES$result`). The function searches on the object the PPTESTCD (test_parameter, ref_parameter) and does a ratio with them. So for the lines you mentioned:
> 
> - `112: if (!any(data$PPTESTCD == test_parameter)) {`
> 
> We need to just test what happens when test_parameter is not in FIXTURE_PKNCA_RES$result. Something similar to: 
> ```
> expect_warning(calculate_ratios(FIXTURE_PKNCA_RES$result, test_parameter = "FAKE_PARAMETER", ...), "No test parameter ..." )
> ```
> 
> - `121: if (!any(data$PPTESTCD == ref_parameter)) {`
> 
> Similar to the previous, but specifing this time the denominator parameter wrong: 
> 
> ```
> expect_warning(calculate_ratios(FIXTURE_PKNCA_RES$result, ref_parameter = "FAKE_PARAMETER", ...), "No test parameter ..." )
> ```
> 
> - `146 (145?):    } else { anti_join(df_test, ref_groups) }`
> 
> Most times we specify how the parameters should be combined. Basically just by indicating the groups (`ref_groups`, `test_groups`) we tell the function what should be the numerator and what the denominator. However, if we don't specify the `test_groups` and leave it NULL, that means that we just want to use all the groups available. In other words, all denominator rows of that parameter that are not per ce `ref_groups` should be combined with the `ref_groups parameters`. So is just a question of either making a `expect_equal()` test that imitates the result (this is feasible if you use a small object as input with at least two different result rows for at least one parameter) or check that the results are a smaller subset when you specify the `ref_groups`
> 
> A bit of a dummy example would be:
> ```
> ratio_res_with_test_groups <- calculate_ratios(FIXTURE_PKNCA_RES$result,  test_parameter = "CMAX",  ref_parameter = "CMAX", test_groups = data.frame(PARAM = "A"), ref_groups = data.frame(PARAM = "A") ...)
> ratio_res_without_test_groups <- calculate_ratios(FIXTURE_PKNCA_RES$result,  test_parameter = "CMAX",  ref_parameter = "CMAX", ref_groups = data.frame(PARAM = "A") ...)
> 
> expect_equal(ratio_res_with_ref_groups, data.frame(...))
> expect_equal(ratio_res_without_ref_groups, data.frame(...))
> ```
>   

 _Originally posted by @Gero1999 in [#783](https://github.com/pharmaverse/aNCA/issues/783#issuecomment-3600679442)_

jimrothstein:
## Issue

Closes #808

## Description
Tests to 100% cover the logic in `ratio_calculations.R`

## Definition of Done
- [x] 100% coverage `ratio_calculations.R`

## Contributor checklist
- [x] New logic is documented


Gero1999:
@jimrothstein Let me give you more context. When we do Non-Compartmental Analysis (NCA), our purpose is to characterize some drug properties by calculating what are called pharmacokinetic parameters (PP). These PPs CDISC mandates to collect them in a standard dataset format (called SDTM.PP or PP in short). This format specifies what standard names (PPTEST) or code names (PPTESTCD) should they have.  Our whole list of PPs can be found with their descriptions in [metadata_nca_parameters.csv](https://github.com/pharmaverse/aNCA/blob/main/data-raw/metadata_nca_parameters.csv)

Ratio Parameters are a special kind of PPs, because they are derived by dividing some PPs (like CMAX) with others. Their general purpose is to have a measure that somehow compare the behavior of the drug under different circumstances. So in order to specify those circumstances, you need to tell the App what groups will be used as the test (fraction nominator) or the reference (fraction denominator) of the ratio calculation. To understand better the function I also recommend you to run the App, map data and then go the NCA tab at left and find the collapsible for `Parameter ratios`. There you will find a help button (icon `?`) that explains better how the operation is done. It will also probably give you more understanding on why we have these arguments for the function.

jimrothstein:
@Gero1999 

Found this at bottom of NCA tab.

I am little bit old-school; need to use pencil/paper calculations.

Can you prioritize:

1)  To finish this PR, (code coverage) , I need to **stare** at code some more; should be simple, but I  guessing.
2) For couple of days, find examples of paper/pencil calculations of ratios.  Then return to PR?

Also, you mentioned SDTM.  In `Admiral`, the process is about moving from SDTM to ADAM.   aNCA uses SDTM?




Screenshot 11-12-2025 at 19 37

jimrothstein:
@Gero1999 **Update**

Line # 146 of `ratio-calculations`   ("anti-join").
Want to run test to run this line  (for **code coverage**)

SEE test-ratio-calculations.    At very bottom, "experimental"  fake data to run this line of code.   (Gemini LLM)
But there is a problem deeper in code:    by = c("start", "end", "USUBJID") and poss.    `bad` includes USUBJID. 
```
bad <- !charmatch(by, poss, 0L)   # 
```
Need to look at this again went I am fresh.

jimrothstein:
@Gero1999 

2 more lines with `NULL` to check:  
#161, #167

thinking how to remove `PPSTRESU` from names (merged file)

https://github.com/pharmaverse/aNCA/blob/dedaef77d1b5acf74d2b519e51243d6b1a1de6bf/R/ratio_calculations.R#L154-L170

Gero1999:
Hey @jimrothstein this last one is actually not well covered and that is why you are finding it difficult. I will make some commits to fix it

Gero1999 (changes requested):
some additional comments I have after a quick look. Looks like we are almost there for the merge! ^^

Gero1999 (approved):
LGTM! Just added some last style aspects. Because we have made changes to the main code (`ratio_calculations.R`) it would be nice to also bump the package version in `DESCRIPTION` before merging

Gero1999:
you just need the PRO version and then Copilot appears as an optional reviewer

tests/testthat/test-ratio_calculations.R
```diff
@@ -271,5 +291,39 @@ describe("calculate_ratios", {
     expect_true(is.character(ratios_empty$PPORRESU))
     expect_true(is.character(ratios_empty$PPSTRESU))
   })
+ it("uses all alternative groups to ref_groups when test_groups is NULL", {
+    test_groups <- NULL
```
Gero1999: my bad I think I miswrite in my suggestion this one

lintr: identation issue
```suggestion
  it("uses all alternative groups to ref_groups when test_groups is NULL", {
    test_groups <- NULL
```

jimrothstein:
@Gero1999 

I'm collecting various comments you have made.
Useful to me.
[810_pharmaverse_NOTES.pdf](https://github.com/user-attachments/files/24226952/810_pharmaverse_NOTES.pdf)

(If you don't see it, I will send another way)

m-kolomanski (approved):
LGTM

PR #818, opened 2025-12-12
Bug Fix: Plots crash when >1 unit included

### LINKED ISSUE #675: Bug:  Individual plots crash if multiple PCSPEC selected

js3110:
## Issue

Closes #675 

## Description

Plots were crashing if >1 unit included in the filters. I have fixed it by collapsing unique units into a string.

## Definition of Done

- [x] No crash
- [x] Multiple units handled in plots

## How to test

In mean and ind plots -> select URINE and SERUM and check plots. Should not crash, and should display both units in the y axis label.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer


m-kolomanski:
Hey, I was wondering the same thing as @Shaakon35, seems odd to have two different units on one axis. Would it make sense to just automatically facet the plots based on matrices, as opposed to units?

Gero1999:
hey @m-kolomanski we actually talked yesterday and we precisely agreed on this strategy because it is the one the App used to have before and we did not receive bad feedback on it.

The thing is that sometimes matrices (`PCSPEC`) might be the reason for the difference in units, others might be the `ANALYTE` studied and in weird cases it can even be the treatment (`DOSETRT`) or for pre-clinical studies the specie of the subject. So it is better to fix always UNIT in order to prevent misinterpretations

PR #832, opened 2025-12-15
Bug: zip settings not aligned with download settings

### LINKED ISSUE #824: Bug: ZIP settings not aligned with download settings and don't include slopes
### Description
The settings file that can be downloaded in the NCA setup tab has the following structure:

```
list(
settings = list( analyte, profile, pcspec ...)
slope_rules = list(...)
)
```

but the ZIP export is just:

`list( analyte, profile pcspec ...)`

with no slope rules. Therefore its not being read correctly when uploaded as there is no $settings section.

### To reproduce
Compare uploading settings from direct download vs from zip export

### Expected behaviour
Both settings files should match and inlcude slope rules, and function when uploaded.

js3110:
## Issue

Closes #824 

## Description

Zip file settings was different to manual download settings and therefore not actually being applied when Uploaded.
I updated so the zip file now creates a list of settings and slope rules, in the same way as the current settings file.

## Definition of Done

- [x] Settings upload and download works for both methods
- [x] Slope selector settings saved

## How to test

Run NCA -> Save settings via NCA setup tab, and via zip file download. Then upload settings in a fresh session and see if settings are applied. Try with dif selections and slope selections.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.1.0.9015
+Version: 0.1.0.9016
 Authors@R: c(
     person("Ercan", "Suekuer", email = "ercan.suekuer@roche.com", role = "aut",
            comment = c(ORCID = "0009-0001-1626-1526")),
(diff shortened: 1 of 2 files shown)


Gero1999 (approved):
LGTM, make sure it works after the new parameter selection

PR #851, opened 2025-12-22
Feat: Perform in-app NCA exclusions manually

### LINKED ISSUE #816: Enhancement: Add exclusions manually on the data
### Description
In case the user does not want to just use the mapping of the exclusions but do their own exclusions at the `Review` or `Filtering` stage, they should be able

### Definition of Done
- [ ] Exclusions can be done manually within the App

Gero1999:
## Issue

Closes #816

## Description
The PR allows users to manually add and manage exclusion reasons for concentration data rows in the PKNCA workflow, with full integration into the Shiny UI. It adds a general exclusions module to the Shiny interface, updates the data processing pipeline to handle user-specified in-App exclusions (PKNCA_update_data_object), and documents the new functionality. 

## Definition of Done
- [x] Exclusions can be done manually within the App

## How to test
1. Map Data (with or without NCAwXRS columns)
2. NCA > General Exclusions
3. Play with the addition of exclusions, make sure the table works fine
4. Run NCA
5. Check records were not used for results (e.g, CMAX)
6. Check reasons appear in ADNCA (CDISC) along with the other previously mapped (NCAwXRS)  

## Contributor checklist
- [x] New logic is documented

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,12 +16,10 @@
 
 ## Bugs fixed
 * ZIP folder with results will now include the exploration tab outputs: individual plots, mean plots (#794)
-
 * Updated TMAX label from Time of CMAX to Time of CMAX Observation (#787)
-
 * Bug fix for box/violin plots that were crashing when violin option selected (#786)
+* Summary statistic table changes in grouping variables won't remove previous summary settings (#840)
 * No longer offering direct PK calculations for renal clearance (RENALCL) parameters. PKNCA is currently inaccurate. Instead, the ratios table should be used (#781)
-
 * Bug fix for settings upload via zip file output (#832)
 
 * Bug fix for plotting section where if there is >1 unit the axis label will contain both unique units. (#818)
--- a/inst/shiny/modules/tab_nca/descriptive_statistics.R
+++ b/inst/shiny/modules/tab_nca/descriptive_statistics.R
@@ -30,22 +30,13 @@ descriptive_statistics_ui <- function(id) {
       multiple = TRUE,
       options = list(`actions-box` = TRUE)
     ),
-    card(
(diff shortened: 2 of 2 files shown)


Gero1999:
I agree would be positive to get their feedback, specially from the person that requested it. But this being the first PR we should focus more on the feature working well, the UX can be further improved in the future after holidays. Nevertheless I will try to add everything you said, although I would not classify it as required for merge

js3110:
> @Shaakon35 @js3110 I don't manage to replicate this. Please, make sure that you select all points of the interval: `start` to `end`. The `DOSE 2` predose sample must also be removed. If the problem persists, please feel free to include a screenshot of the exclusions page or a exact reference of the rows you selected

exclusions page:
image

results:
image
image

js3110:
@Gero1999 I see, once I added the DOSE 2 predose sample, then it goes away.
Im not sure this is a good method though, maybe we need to think if there is a way to avoid this or to ensure the user knows that they need to do this to exclude the full profile...

Shaakon35:
> @Shaakon35 @js3110 I don't manage to replicate this. Please, make sure that you select all points of the interval: `start` to `end`. The `DOSE 2` predose sample must also be removed. If the problem persists, please feel free to include a screenshot of the exclusions page or a exact reference of the rows you selected

All good now, it worked when I excluded the patient 1. I don't know why it works now.

Please check the comments from Copilot, there are a few valid things to **change.**
e.g. "Typo" @Gero1999

Gero1999:
Hello @Shaakon35 please consider letting me know which ones you specifically refer to. I already addressed those that I considered relevant/correct. If you see a message saying `Outdated` means that the code of the original suggestion was already modified

PR #855, opened 2025-12-23
fix: ZIP meanplots crash when <3 patients per cohort

### LINKED ISSUE #854: Bug: ZIP meanplots crash < 3 patients
### Description
Meanplots do not handle empty row cases out of the reactive, which produces the crash when saving "All results" with the next error:

```
Download Error:
Faceting variables must have at least one value.
Warning: Error in combine_vars: Faceting variables must have at least one value.
```

### To reproduce
Map data > Filter 2 subjects > Run NCA > Save All results (ZIP)

### Expected behaviour
- [ ] No crash when saving results with < 3 patients

Gero1999:
## Issue

Closes #854 

## Description
This pull request improves the robustness and user feedback of the `g_lineplot` function by handling cases where the input data frame is empty. When no data is available, the function now returns a plot with an informative error message instead of failing.

## Definition of Done
- [X] No crash when saving results with < 3 patients

## How to test
Map data > Filter 2 subjects > Run NCA > Save All results (ZIP)

## Contributor checklist
- [x] New logic is documented


Shaakon35 (approved):
Perfect! that will prevent a lot of bugs

js3110 (commented):
LGTM! Happy to approve but also I would suggest we integrate the same into the boxplots function too? I think its within the same scope so can be handled in this PR

js3110:
> @js3110 the thing is that boxplot were originally already designed to handle empty datasets so they won't crash. But perhaps we can also include the error message on them I guess

I think they handle NA data but the empty dataset is the same as before with validate outside of the function? maybe im wrong. But either way I would say its good to harmonise so they all use the same functions

js3110 (approved):
LGTM :) added a non-blocking comment

PR #860, opened 2025-12-31
Enhancement: Integrate settings upload in Data tab

### LINKED ISSUE #719: Enhancement: Integrate settings upload button with data upload
### Description
Currently, the settings button is not integrated with the data upload step, requiring users to go through additional remapping steps. This enhancement proposes to place the settings upload button together with the data upload, allowing users to upload a dataset with or without settings in a single streamlined step.

### Definition of Done
- [ ] The settings button is now together with the data upload feature.
- [ ] Users can upload a dataset with or without associated settings.
- [ ] The data upload process is streamlined to minimize redundant mapping steps.

js3110:
## Issue

Closes #719 

## Description

Updated data tab upload so that settings files can also be uploaded and read correctly, and then carried into the NCA tab.

## Definition of Done

- [x] Settings can be uploaded in data tab

## How to test

Try uploading variations of datasets / settings etc to see what works and what doesn't work. If something doesn't work, it should give a clear error message to the user.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

### DIFF
--- a/inst/shiny/modules/tab_data/data_upload.R
+++ b/inst/shiny/modules/tab_data/data_upload.R
@@ -155,11 +155,10 @@ data_upload_server <- function(id) {
       }) %>%
         bindEvent(input$data_upload, ignoreNULL = FALSE)
     )
-  
+
     observeEvent(raw_data(), {
       session$userData$raw_data <- raw_data()
-      }
-    )
+    })
 
     reactable_server(
       "data_display",


Shaakon35 (approved):

inst/shiny/modules/tab_data/data_upload.R
```diff
@@ -71,27 +72,69 @@ data_upload_server <- function(id) {
...
+                list(status = "success", type = "settings", content = final_settings, name = name)
+              } else {
+                # Not a settings file either, return original PK error
+                stop(e_pk$message)
+              }
+            }, error = function(e_rds) {
+              # Return the original read_pk error to the user
+              list(status = "error", msg = e_pk$message, name = name)
+            })
           })
```
m-kolomanski: suggestion: Clever approach, but I think we can clean up the code a little bit to make it more robust with specified assumptions and early returns. For one, we are expecting a very specific error to be thrown if `read_pk` tries to load settings file - it will read the RDS fine, but then detect it is a list and not a data.frame, so we can check for that error to see if it makes sense to even try loading it again.

```R
         tryCatch({
            # Attempt to read
            data <- read_pk(path)
            list(status = "success", data = data, name = name, type = "data")
          }, error = function(e_pk) {
            # If read_pk fails check if settings file is loaded and then create settings override
            tryCatch({
              # check if error aligns with what we expect for setttings file
              if (conditionMessage(e_pk) != "Invalid data format. Data frame was expected, but received list.") {
                return(list(status = "error", msg = conditionMessage(e_pk), name = name))
              }

              obj <- readRDS(path)
              is_settings <- is.list(obj) && "settings" %in% names(obj)

              if (!is_settings) {
                stop(conditionMessage(e_pk))
              }
              
              list(status = "success", type = "settings", content = obj, name = name)
            }, error = function(e_rds) {
              list(status = "error", msg = conditionMessage(e_pk), name = name)
            })
          })

```

suggestion(more like idea): I would also consider abstracting away this logic into separate `read_settings` function, so that we can easily add new checks if we wanted (to see if uploaded settings file is truly valid, has all required elements etc). But since this is very little code at the moment, I think we can hold out on that for when we decide to migrate to `yaml` format.

inst/shiny/modules/tab_data/data_upload.R
```diff
@@ -71,27 +72,69 @@ data_upload_server <- function(id) {
...
         })
 
         # Process results
-        successful_loads <- purrr::keep(read_results, ~ .x$status == "success")
+        successful_loads <- purrr::keep(read_results, \(x) x$status == "success")
```
m-kolomanski: :+1:

inst/shiny/modules/tab_data/data_upload.R
```diff
@@ -71,27 +72,69 @@ data_upload_server <- function(id) {
...
+          # Success: Single settings file
+          latest <- found_settings[[1]]
+          settings_override(latest$content)
+          log_success("Settings successfully loaded from ", latest$name)
+          showNotification(paste("Settings successfully loaded."), type = "message")
+        }
+
         loaded_data <- DUMMY_DATA
 
+        found_data <- purrr::keep(successful_loads, ~ .x$type == "data")
```
m-kolomanski: issue, style:
```suggestion
        found_data <- purrr::keep(successful_loads, \(x) x$type == "data")
```

PR #861, opened 2026-01-02
Bug fix: allow unknown units in AVALU and DOSEU

### LINKED ISSUE #836: Bug: App crashes if unknown units in AVALU or DOSEU
### Description
If units in AVALU or DOSEU are not standard, eg "cells" the app crashes:

### To reproduce
Try changing dummy data units to "cells" or other character strings

### Expected behaviour
No crash, app can handle it alone.

js3110:
## Issue

Closes #836 

## Description

If units are not known by the `units` package, the app was crashing due to the `simplify_units` function for simplifying the volume units. I have updated this function to return a warning rather than stop if the units are not recognised, so that more non-traditional units (eg cells) can still be used.

## Definition of Done

- [x] App does not crash if non traditional units are used
- [x] Gives warning that simplification was not possible

## How to test

Using `dummy_RO_cells` in `testthat/data` folder, run through and check that the app doesn't crash with the units, or try your own dataset with different units.

## Contributor checklist
- [x] New logic is documented

## Notes to the Reviewer

I have added the dummy dataset to make it easier to test, we can remove it before I merge or keep it for testing purposes, let me know what you think is best :)
The dataset is the same as the dummy duplicates data, but with AVALU and DOSEU changed to copies/ug and cells respectively.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -24,7 +24,7 @@
 * Bug fix for settings upload via zip file output (#832)
 * Bug fix for plotting section where if there is >1 unit the axis label will contain both unique units. (#818)
 * Units table bugs fixed, so it is filtered based on NCA setup and the table is searchable for each column (#870)
-* Bug fix to allow for unrecognised units to be used in AVALU and DOSEU (#861)
+* Bug fix to allow for unrecognized units to be used in AVALU and DOSEU (#861)
 
 
 # aNCA 0.1.0


js3110:
> It works well, when I put an invalid unit, it still uses the original unit. I guess you wanted to code it this way

yes. You can manually add a conversion if you want to

js3110:
> Better to confirm with the user case, but the solution and code-wise LGTM! As a little note, if possible indicate a small description on how did you created the testing dataset in the PR

Should we keep the testing set? I added a note to say we can remove it when I merge but idk what is the best option... thoughts?

Issue #729, opened 2025-10-21
Enhancement: Simplify the way units_table user changes reflect in the server


Gero1999:
### Description
Currently changes to the units table seem to be independent from `PKNCA_update_data_object` and also they don't detail the non-default changes made by the user (the changes are done on the table and then this shows up...). Perhaps this is something to think about on how to simplify it so it looks better in the R script.

### Definition of Done
The requirements for the feature to be complete:
- [ ] consider integrating units_table as an argument in PKNCA_update_data_object
- [ ] consider a cleaner way to show in the input non-default unit changes made by the user

PR #870, opened 2026-01-05
Bug fix:  recover units table functionalities

### LINKED ISSUE #615: Bug: Recover units table functionalities
### Description
Some functionalities from the old DT table have been missing, probably during the reactable transition (#528). Some aspects that were useful and if possible would be great to recover are:

- [ ] Parameters not selected being filtered
- [ ] Column-wise search, if possible (although unlikely) with selection options
- [ ] `PPTEST` instead of `PPTESTCD` for the parameter names

Here are some comparing images in case something else potentially relevant could be missing

Image

Image

### To reproduce
NCA > Settings > Parameter Units

js3110:
## Issue

Closes #615 

## Description

Units table functionalities have been restored after the switch to reactable

## Definition of Done

- [x] Parameters not selected being filtered
- [x] Column-wise search, if possible (although unlikely) with selection options
- [x] PPTEST instead of PPTESTCD for the parameter names

## How to test

Try out units table and ensure everything works as expected (Note: units table changes are not applied in results section, but this is an issue in main also so I think it should be a separate bug issue, or maybe is linked with #729 ?)

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

@Gero1999 I am not entirely sure what you mean by "with selection options" in the column search, so I have just left it without. Feel free to elaborate and I can see if its something to add...

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.1.0.9031
+Version: 0.1.0.9032
 Authors@R: c(
     person("Ercan", "Suekuer", email = "ercan.suekuer@roche.com", role = "aut",
            comment = c(ORCID = "0009-0001-1626-1526")),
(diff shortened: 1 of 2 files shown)


Gero1999:
@js3110  regarding the question in the Reviewer notes:

 if you check the old version of the App you can see that in the past each column had its own search-box, and each search-box allowed you to even select choices (instead of needing to type them completely)

js3110:
> @js3110 regarding the question in the Reviewer notes:
> 
> if you check the old version of the App you can see that in the past each column had its own search-box, and each search-box allowed you to even select choices (instead of needing to type them completely)

Ahh i see what you mean, thanks!

IMO this is not necessary, and would likely be more difficult to code than it would bring benefit to the user... lmk if you disagree!

Issue #900, opened 2026-01-15
Enhancement: unit settings only manual changes and not entire units table


js3110:
### Description
For better implementation of the settings file in yaml form, and the R script implementation, it would be good to only track the manual unit changes (aka if conversion factor is not 1) and store those, so then the units are generated automatically and conversions are applied afterwards

### Definition of Done
The requirements for the feature to be complete:
- [ ] unit changes (conversion factor != 1) are tracked and stored
- [ ] Manual unit changes stored in settings
- [ ] Manual unit changes stored in R script
- [ ] Unit changes applied to default units table after generation

Issue #891, opened 2026-01-13
Enhancement: Allow flexible settings upload ( not matching to data)


js3110:
### Description
Currently, the settings uploaded must be created from analysis of a dataset. Then when uploaded, it will match based on the data e.g. params and pcspec selected. We need to ensure that:
- Settings can be uploaded that do not contain analytes, pcspec, slopes etc but the rest of the settings will be populated
- Settings file can be easily created with only default settings that are not specifically related to data (eg parameter selections, rule sets)

### Definition of Done
The requirements for the feature to be complete:
- [ ] Created Settings file template with only the default settings not related to data
- [ ] Settings uploaded and defaults widgets are populated
- [ ] If settings are specific to  data, user is given warning if they don't match the current data. If settings are not specific to data, user will not be given warning.

PR #901, opened 2026-01-15
Enhancement/Refactor:  move settings from rds to .yaml

### LINKED ISSUE #892: Enhancement: Move from .rds to yaml file for settings
### Description
To make it easier for users to write their own settings files and templates, rather than having to create it in the app, we should move the settings from rds to yaml, as its easily readable and editable.

### Definition of Done
- [ ] Settings file as yaml
- [ ] Settings uploaded and applied as before

js3110:
## Issue

Closes #892 

## Description

Settings file refactored to .yaml for better readability and user configuration. 
This PR is the most basic implementation, directly changing the import and export from rds to yaml, but maintaining all other features. Later PRs can develop the yaml file to make it more user friendly (#900 and #891 )

## Definition of Done

- [x] settings file yaml form
- [x] settings uploaded and applied as before

## How to test

Try different settings, export and upload and test that it works.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

For this PR, I have implemented a Hard change from .rds to yaml. Not sure if that is the best approach though, as we may want to keep rds in case anyone has saved their settings previously... but eventually we will only need yaml

### DIFF
--- a/inst/WORDLIST
+++ b/inst/WORDLIST
@@ -37,7 +37,6 @@ LAMZ
 LAMZLL
 LAMZNPT
 LAMZSPN
-Lineplot
 METABFL
 MRTMDO
 MRTMDP
@@ -144,6 +143,7 @@ pptest
 pre
 pred
 qmd
+rds
 reproducibility
 reupload
 roadmap
@@ -151,7 +151,6 @@ rpptx
 ruleset
 rulesets
 runnable
-SBS
 signif
 src
 tibble


R/read_uploads.R
```diff
@@ -85,3 +85,81 @@ readers <- list(
...
+      if (!is.null(settings)) {
+        list(status = "success", content = settings, name = name, type = "settings")
+      } else {
+        stop(e_pk$message)
+      }
+    }, error = function(e_yaml) {
+      list(status = "error", msg = e_pk$message, name = name)
+    })
+  })
+}
```
m-kolomanski: issue: This logic should be part of the application, not the package. If user does this programatically, we want them to specifically load data and settings with tailored functions, not guess which file it is.

I get that we have this problem due to how the app is structured (we want to have a single file input, instead of two), but this should remain an app-problem. With this we are making it also a package-problem.

I am really not enjoying the further coupling of package and application. IMHO we should have:

- `read_pk` function for reading raw data
- `read_settings` function for reading setting
- either the application or the direct user should decide which one to apply

js3110:
@m-kolomanski thanks for the comments, makes sense. I have refactored so theres now read_pk and read_settings, and the rest is handled in the app rather than the package. PTAL! :)

Issue #896, opened 2026-01-14
Bug: FLAG column not correct in edge cases


js3110:
### Description
FLAGGED column is incorrect in edge cases, when:
- Rule sets are selected, but parameter was unable to be calculated and no reason is given in Exclude column.

### To reproduce
In dummy data, add manual selection to Subj 1, Range 6:7. Deselect LAMZSPN as a rule and then Run NCA and view that row is "ACCEPTED". This is incorrect as R2ADJ is NA, so the FLAG should be set to "MISSING"

### Expected behaviour
If rule is applied, and Param is NA, should be set to missing
Reason ideally should be given in Exclude, eg "R2ADJ not calculated"

PR #904, opened 2026-01-19
fix: HL parameters being NA for no reason & flags with no rules (#896)

Gero1999:
## Issue

Closes

## Description
This pull request updates the logic for determining which pharmacokinetic parameters should be imputed in the `apply_imputation` function. The main change is to expand the set of parameters considered dependent on half-life or lambda.z, ensuring that imputation is skipped for parameters that depend on these as well as AUC/AUMC.

**Imputation logic improvements:**

* Expanded the set of parameters (`params_hl_dep`) to include those related to `lambda.z` and `half.life` in addition to `auc` and `aumc`, both in the parameter name and dependencies, to more accurately identify parameters that should not be imputed.
* Updated the filtering logic for `params_not_to_impute` to use the new `params_hl_dep` set, ensuring that any parameter depending on AUC, AUMC, lambda.z, or half-life is excluded from imputation.

## Definition of Done
- [ ] If rules are not selected, they should not be applied to the NCA calculation.

## How to test
In dummy data, add manual selection to Subj 1, Range 6:7. Run NCA and view that row is "MISSING" this is expected. Then deselect all rule sets and rerun, row is still "MISSING".

## Contributor checklist
- [ ] New logic is documented

## Notes to reviewer

### DIFF
--- a/R/intervals.R
+++ b/R/intervals.R
@@ -209,13 +209,13 @@ apply_imputation <- function(data, nca_parameters = metadata_nca_parameters) {
   data <- create_start_impute(data)
 
   # Don't impute parameters that are not AUC dependent
-  params_auc_dep <- nca_parameters %>%
-    filter(grepl("auc|aumc", PKNCA) | grepl("auc", Depends)) %>%
+  params_hl_dep <- nca_parameters %>%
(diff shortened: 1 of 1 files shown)


Gero1999:
The original issue was reformulated, then spitted (#896) and now is another diff issue in another location. For simplicity I will close this PR

PR #907, opened 2026-01-19
Bug Fix: na units ignored in units table creation

### LINKED ISSUE #837: Bug: App gives error message if some units are NA
### Description
According to CDISC standards, if AVAL is NA then AVALU should also be NA, but this causes an error after mapping in the app:

`ERROR [2025-12-16 14:10:55] Units should be uniform at least across concentration groups.Review the units for the next group(s):`

### To reproduce
Set one row in dummy data to have AVAL = NA and AVALU = NA

### Expected behaviour
The PKNCA object will not use any NA AVAL values anyway, so having AVALU as NA should not be a problem. The issue arises in the creation of the units table, where unique values are checked. 
We may need to filter first to remove NAs, or add a different check? I am not sure the best way to go about it.

js3110:
## Issue

Closes #837 

## Description

If some units are NA in the app, the app was not allowing you to go past data mapping. Now, the NA units are ignored from the units table creation (and therefore from the NCA analysis.) .  The samples are not ignored, but the units will be assumed to match the pre-existing units for the rest of the matching group.

## Definition of Done

- [x] App still works if units NA
- [x] NA units are not included in units table
- [x] NA aval and  units rows are excluded from analysis
- [x] NA units but not NA aval rows are still included

## How to test

I have created temporary dataset DummyROnaunits.csv that can be used for testing. Check that NCA works fine and rows are excluded as expected.

## Contributor checklist
- [x] New logic is documented
- [x] Remove temporary dataset

Notes to Reviewer

I added and then removed a showNotification warning, since itsin a function so didn't pass the checks. However not sure if its actually needed since the NA units are simply removed from the units table, not from anything else.


js3110:
> This seems a dangerous approach. The best imo would be to provide an error message to the user telling them to fill in NAs units. Making this assumption can derive in invisible errors.
> 
> Otherwise, if you are keen on following this path, make sure we only "ignore" sample units when the NAs are in both its units and their respective variable

NA units are CDISC standard if AVAL is NA, so we shouldn't ask them to fill them in.
The samples are only ignored if AVAL is NA, otherwise it just takes the unit that is used for the rest of the subject/sample (there is an example row of this in the dummy data, where AVAL is a value and AVALU is NA. You can see that this time point is still used in slope selector)

In the case of AVALU being NA for the entire subject, I would have to test, but I think this is an unlikely scenario

Issue #878, opened 2026-01-07
Bug: If bioavailability calculated, Flag always set to MISSING


js3110:
### Description
Since the merge of #702 , the logic was changed so if Exclude column is not empty, the FLag is set to Missing (unless its flagged). However, at the moment the bioavailability information is added to the Exclude column.

Ideally, we would move the Bioavailability information to a new column: maybe "Notes" as it is not related to exclusions, and then the Exclude/Flagging logic will remain the same.

May also be solved with #728 , but depends how this is implemented. To be discussed..

### To reproduce
Calculate bioavailability and see that all rows are set to MISSING

### Expected behaviour
Only rows that are actually missing should be labelled missing.

PR #934, opened 2026-01-26
Bug fix: flag column not correct in edge cases

### LINKED ISSUE #896: Bug: FLAG column not correct in edge cases

js3110:
## Issue

Closes #896 , and also inadvertently #878

## Description

Flag column was not assigning MISSING in cases where the parameter was missing but the Exclude column was empty. Here, the logic now takes into account the parameter selections, and if the parameter is NA, but was selected, the Flag is set to MISSING.

## Definition of Done

- [x] Flags assigned correctly

## How to test

Try various combinations of parameter selections, flags and slope selectors to try and create FLAGGED, MISSING and ACCEPTED. Check if they are applied correctly.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.1.0.9054
+Version: 0.1.0.9055
 Authors@R: c(
     person("Ercan", "Suekuer", email = "ercan.suekuer@roche.com", role = "aut",
            comment = c(ORCID = "0009-0001-1626-1526")),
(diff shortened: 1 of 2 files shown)


Gero1999 (approved):
LGTM, good job!

PR #939, opened 2026-01-27
fix PPSTRESN 

### LINKED ISSUE #561: Bug: PPSTRESN != PPSTRESC

Shaakon35:
## Issue

Closes #561 

## Description

I fixed three columns so that the numeric always match the character variable 

 >(@Shaakon35 if possible explain here more what is the rationale behind the solution; why this works and/or what was the part of the code producing the problem)

## Definition of Done

For PP/ADPP, it is expected based on CDISC rules that the result variables in numeric (`PPSTRESN`) and character (`PPSTRESC`) matches in length and share exactly the same values for the PP/ADPP objects. Therefore: `PPSTRESN = PPSTRESC`.

When no custom units are specified, is also expected that ``PPORRES = PPSTRESN = PPSTRESC`.

## How to test

Run NCA with dummy data, download ADPP and PP to check all these variables are exactly the same value/length:  PPORRES = PPSTRESN = PPSTRESC

## Contributor checklist
- [ ] New logic is documented


Shaakon35:
@m-kolomanski Tests added.

@Gero1999 @js3110 
There is still one issue that does not happen often but does not work:
if PPORRESU and PPSTRESU don't have the same unit, the rounding of the numeric and character columns is not exactly perfect, but I guess we can leave it like that for now.

PR #946, opened 2026-01-30
Enhancement: settings file saves only edited units

### LINKED ISSUE #900: Enhancement: unit settings only manual changes and not entire units table

m-kolomanski:
## Issue

Closes #900 

## Description

- Settings are now saved with only units that were edited by the user.
- Units table handles the distinction between default and user-defined units.
- If there is no manual definitions for specific parameters, default units are implemented.

## Definition of Done
- [x] unit changes (conversion factor != 1) are tracked and stored
- [x] Manual unit changes stored in settings
- [x] Manual unit changes stored in R script
- [x] Unit changes applied to default units table after generation

## How to test

1. Run the application, go to NCA tab with dummy data. Open units table and change some units.
2. Download settings file. Check the `yaml` file, see if only changed units are present.
3. Refresh the application and upload settings file. Go through the same steps and check if manual changes to the units are applied. Check if defaults are kept.
4. Check if the changes are reflected in the template script.

## Contributor checklist
- [x] ~New logic covered by unit tests~
- [x] ~New logic is documented~
- [x] ~App or package changes are reflected in NEWS~


Shaakon35:
> This works nice when I download the settings from the `NCA Results` page, but not when I top-right use `Save ZIP`

Same here. Otherwise everything works well :)

m-kolomanski:
Hey @Gero1999 @js3110 @Shaakon35 thank you for spotting the issue with zip export, I have completely missed that. The feature is now included in the zip export, PTAL :pray:

Gero1999 (approved):
LGTM!

PR #958, opened 2026-02-03
Feat: Allow choosing other INT parameters than AUCINT for area estimations

### LINKED ISSUE #524: Enhancement: AUC intervals with Parameter selection
### Description
Currently when the user requests in `NCA > Setup > Partial AUC` is automatically assumed they want AUCINT. However, is possible that they want instead AUCINTD. So we need to add a new column so they can choose between the two.

Additionally, we don't want AUCINT / AUCINTD to be selected in the `Parameter Selection` table, as they should not be used for anything else than partial intervals. 

### Definition of Done
- [ ] remove AUCINT / AUCINTD from `Parameter Selection` table
- [ ] include a new column in `Partial AUCs` with default AUCINT, but also as an option AUCINTD or CAVGINT

### LINKED ISSUE #935: Enhancement: cumulative intervals for excretion parameters
### Description
Requested by user:

"Our urine TL templates typically include cumulative intervals (e.g., 0–6, 0–12, 0–24) aswell, and we find these cumulative values to be operationally quite valuable. They’re particularly helpful for diagnosing data issues early and can sometimes allow us to salvage subjects when one of the intermediate intervals is missing or incomplete"

This could be implemented in the excretion tab, as an additional option ("Cumulative") so that each of the samples are cumulated and give parameter results.

### Definition of Done
- [ ] Cumulative urine intervals can be calculated
- [ ] Cumulate each sample for a single profile

Gero1999:
## Issue

Closes #524 
Might also close #935

## Description
This PR:

- Allow users to choose in the renamed `Partial Interval Calculations` (previously "Partial AUCs") the INT parameter they want to calculate

- Replaced all uses of `auc_data`/`partial_aucs` with `int_parameters` in function arguments and internal logic, notably in `update_main_intervals` and related processing. The new `int_parameters` data frame now includes a `parameter` column to specify which parameter the interval applies to.

- Updated the Shiny module UI for the dropdown card to rename "Partial AUCs" to "Partial Interval Calculations".

**Tests:**
- Refactored tests to use `int_parameters` in place of `auc_data`, including updating test cases to check for the new structure and parameter handling.

**Other affected files:**
- Updated related function calls and template scripts to use `int_parameters` throughout, ensuring consistency across the codebase. Also documentation.

## How to test

- First, make sure the default NCA run produces the same results as in `main` (we changed something in `create_start_impute`)

- Second, check the implementation works as expecterd. Try different combinations of interval parameters by following the next schema:

Map data > NCA > Setup > Parameter Area Calculations > Add row(s) > Select different parameter/times

Please, check `Notes to reviewer` for the second check. If you notice a specific parameter is not working, check also if it does not work in `NCA > Parameter Selection`, as this will need a different issue to solve it in both.

- Third, Save a ZIP file of results and run the R script to make sure there are no issues

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

Please, notice that `FREXINT` won't work. This also happens when you select the parameter in the regular `Parameter Selection`. Something must be done for it apart

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -21,6 +21,8 @@
 * WTBL and WTBLU columns added to the data mapping, for optional conversion of dose to adjust to body weight for excretion calculations (#959)
 * Slope selector table for half life adjustments uses time to choose the point of interest. Also it is aesthetics have been polished (#956)
 * Partial interval parameters section now allows other calculations than `AUCINT`, such as `RCAMINT`, `AUCINTD` or `CAVGINT` among others (#524)
+* Slope selector plots count with grouping options (#333)
+* Add x/y axis limits for the exploration plots (#817)
 
 ## Bugs fixed
 * ZIP folder with results will now include the exploration tab outputs: individual plots, mean plots (#794)
--- a/R/exploration_plots.R
+++ b/R/exploration_plots.R
@@ -9,6 +9,10 @@
 #' @param ylog_scale Logical; whether to use a logarithmic scale for the y-axis. Default is `FALSE`.
 #' @param threshold_value Numeric; y-intercept for a horizontal threshold line.
 #' Default is `NULL` (no threshold).
+#' @param x_limits Numeric vector of length 2 for x-axis limits (min, max).
+#' Default is `NULL` (no limits).
+#' @param y_limits Numeric vector of length 2 for y-axis limits (min, max).
+#' Default is `NULL` (no limits).
 #' @param show_dose Logical; if `TRUE`, vertical lines for dose times are shown. Default is `FALSE`.
 #' @param palette Optional palette name or named character vector of colors for the plot.
 #' Default is "default" color palette.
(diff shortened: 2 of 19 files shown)


inst/shiny/modules/tab_nca/setup/settings.R
```diff
@@ -473,3 +500,11 @@ settings_server <- function(id, data, adnca_data, settings_override) {
...
   available_choices[1]
 }
+
+# Define the parameters that can be used for partial area calculations
+INT_PARAMS <- metadata_nca_parameters %>%
+  filter(
+    grepl("INT", PPTESTCD),
+    TYPE != "PKNCA-not-covered"
+  ) %>%
+  arrange(PPTESTCD)
```
m-kolomanski: suggestion: Include this part in module server to avoid polluting the global environment with constants.

Gero1999:
> LGTM.
> 
> suggestion, nitpick: `int` is a popular shorthand for `integer` - seeing `int_params` is kinda confusing to my brain, even though I know perfectly well it stands for `interval`. `auc_*` or `aucint_*` would be better, but I can see that is not quite accurate based on the params choices. If we have any better name, than I would suggest changing it - if not, `int` ok.


@m-kolomanski yep you are right it can indeed cause confusion also with standard intervals. I will perhaps change it for `ppint` (Pharmacokinetic parameters) or `param_int`

> comment: The PR contains some unrelated formatting changes - these should probably be submitted separately, as they pull the focus away from the actual logic changes.

sorry about that, you are right as well

js3110 (approved):
In that case I'll approve, but lets make sure we don't forget to discuss / think about this potential problem

PR #959, opened 2026-02-03
Bug fix / Enhancement: crash if weight column not numeric

### LINKED ISSUE #957: Bug/enhancement:  crash if WEIGHT column not numeric -> rename WEIGHT to WTBL and add to mapping
### Description

1. Weight is not CDISC standard, instead it should be WTBL (and WTBLU for unit)
2. Crash during excretion calculations because WEIGHT is not in metadata, so its not converted to numeric after mapping

### To reproduce
Use dataset attached (multispec data adapted to include WEIGHT column)
Go to excretion tab, and try to calculate excretion for URINE. Should get the following message:

```
Warning: Error in mutate: ℹ In argument: `weight_with_unit = pmap(list(.w = WEIGHT, .wu = WEIGHTU), ~set_units(..1, ..2, mode = "standard"))`.
Caused by error in `pmap()`:
ℹ In index: 1.
Caused by error in `UseMethod()`:
! no applicable method for 'set_units' applied to an object of class "character"
```

[test-multispec-ADNCA.csv](https://github.com/user-attachments/files/25048873/test-multispec-ADNCA.csv)

### Expected behaviour
WEIGHT and WEIGHTU should be renamed to WTBL and WTBLU
Ideally, we should add these columns to the supplementary variables in the mapping section.

js3110:
## Issue

Closes #957 

## Description

For excretion calculations where Dose is adjusted for body weight, we were previously using WEIGHT and WEIGHTU columns, with no user mapping. However, CDISC weight column is WTBL. To fix the bug (due to WEIGHT not being in metadata and not specifying numeric) we renamed and added info in metadata for conversion of class.

In this PR I have 
1. Changed WEIGHT and WEIGHTU in excretion tab to WTBL and WTBLU
2. Added WTBL and WTBLU to the column mapping via the metadata

In addition, I updated the metadata handling so that the column mapping uses `mapping_order` to define the order of the select inputs.

## Definition of Done

- [x] WEIGHT and WEIGHTU renamed to WTBL and WTBLU
- [x] WTBL and WTBLU added to column mapping

## How to test

Try with test multispec data and map weight column, check that excretion calculations with body weight adjustment work well.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer


Gero1999:
hey it is indeed not crashing (olee). But in the testing which parameters are supposed to be adjusted by the `Body Weigth`? I don't manage to get different results. Also FREXINT might be dissapearing (?):

Screenshot 2026-02-06 114506

Screenshot 2026-02-06 114451

js3110:
@Gero1999 only parameters that use dose are adjusted by BW (in this case, I believe its only FREXINT), therefore none of the other results should change. FREXINT often cannot be calculated if the dose units do not match the amount excreted units (which is why we need to convert using body weight)

For example, in the multispec data:
- DOSEU = mg/kg
- AMOUNTU = g
- FREXINT = AE / DOSE, so units = g / (mg/kg) which doesn't work with PKNCA (see below)
image

If you adjust BW:
- DOSEU = mg
so then FREXINT is g/mg which then works (with simple conversion)

If FREXINT can't be calculated due to incorrect units, it doesn't appear. We could update this to give the user a warning but that would be a separate issue...

inst/shiny/modules/tab_nca/excretion_analysis.R
```diff
@@ -86,14 +86,14 @@ excretion_server <- function(id, input_pknca_data) {
...
+      weight_col <- "WTBL"
+      weightu <- "WTBLU"
 
       # Adjust dose by bodyweight if selected
       if (input$adjust_bw) {
         # Check if weight columns exist
         if (!(weight_col %in% names(data$dose$data)) || !(weightu %in% names(data$dose$data))) {
-          showNotification("Please ensure WEIGHT and WEIGHTU columns exist in the dose data.
+          showNotification("Please ensure WTBL and WTBLU columns exist in the dose data.
                            No adjustments can be made.", type = "warning")
```
m-kolomanski: nitpick: We assign the column names to the variables, but then hard-code them here.

```suggestion
          showNotification(glue::glue("Please ensure {weight_col} and {weightu} columns exist in the dose data.
                          No adjustments can be made."), type = "warning")
```

m-kolomanski (approved):

PR #961, opened 2026-02-06
Bug fix: Crash if units not included in settings

### LINKED ISSUE #960: Bug: settings upload crash if no units in settings
### Description
Linked with #891 , currently if there are no units table in the settings file, the app crashes when you try to run  NCA
App crashes, but also units table does not appear with error messages

### To reproduce
Upload settings file with no units -> run NCA

### Expected behaviour
Default units should be applied.

js3110:
## Issue

Closes #960 

## Description

App was crashing and units table not rendering if the settings were uploaded but without units included. Fix to check that the units table exists and is also nrow > 1 to ensure that the settings aren't just empty.

## Definition of Done

- [x] Units table renders
- [x] NCA runs well

## How to test

Create settings with and without units -> upload and check that everything runs smoothly

## Contributor checklist
- [x] New logic is documented

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.1.0.9058
+Version: 0.1.0.9059
 Authors@R: c(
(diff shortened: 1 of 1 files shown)


Shaakon35 (approved):

Gero1999 (approved):
LGTM!

PR #970, opened 2026-02-10
By default display more half life slope plots in `Slope Selector` & shorten titles

### LINKED ISSUE #969: Enhancement: Default for more half life plots displayed in App
### Description
Now that the server side is much faster after loading the plots we can afford showing by default more plots in this tab. Most users will be interested in quickly seeing all their plots once they map the data.

### Definition of Done
- [ ] Display more than 1 plot by default in the `Slope Selector` section

Gero1999:
## Issue

Closes #969

## Description

This pull request makes several improvements to the half-life plotting functionality and the related Shiny UI. The main focus is on enhancing the plot titles to be more concise and relevant, and updating the plot selection options in the UI.

**Enhancements to plot titles:**

* The plot title logic in `get_halflife_plots.R` was updated to only include grouping columns with more than one unique value, making titles shorter while fully  informative. 

* The `plots_per_page` selection in the slope selector UI was updated to provide a more practical set of options and to default to showing 8 plots per page instead of 1.

## Definition of Done

- [x] Display more than 1 plot by default in the Slope Selector section

## How to test

Check how the plots look on:

Map data > NCA > Slope Selector

By default many are displayed for the user to review.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

I tried making the title/subtitle as tables but unfortunately plotly does not support html table tags... perhaps @m-kolomanski has a technical alternative idea for this

@Shaakon35 In relation with using variables to indicate the profile number in the title (such as ATPTREF), which I think you may think about it, this discussion is better to have it in a different issue because is more complicated than what it seems

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -19,8 +19,9 @@
 * Settings are now uploaded on initial opening of the app in the data tab, and applied to the next steps (#860)
 * Settings file has been converted from rds to yaml, allowing better readability and editing for users. (#901)
 * WTBL and WTBLU columns added to the data mapping, for optional conversion of dose to adjust to body weight for excretion calculations (#959)
-* Slope selector table for half life adjustments uses time to choose the point of interest. Also it is aesthetics have been polished (#956).
-* Slope selector plots count with grouping options (#333).
+* Slope selector table for half life adjustments uses time to choose the point of interest. Also it is aesthetics have been polished (#956)
+* Slope selector plots count with grouping options (#333)
+* Add x/y axis limits for the exploration plots (#817)
 
 ## Bugs fixed
 * ZIP folder with results will now include the exploration tab outputs: individual plots, mean plots (#794)
--- a/R/exploration_plots.R
+++ b/R/exploration_plots.R
@@ -9,6 +9,10 @@
 #' @param ylog_scale Logical; whether to use a logarithmic scale for the y-axis. Default is `FALSE`.
 #' @param threshold_value Numeric; y-intercept for a horizontal threshold line.
 #' Default is `NULL` (no threshold).
+#' @param x_limits Numeric vector of length 2 for x-axis limits (min, max).
+#' Default is `NULL` (no limits).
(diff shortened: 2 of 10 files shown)


R/get_halflife_plots.R
```diff
@@ -24,6 +24,11 @@ get_halflife_plots <- function(pknca_data, add_annotations = TRUE) {
   concu_col <- pknca_data$conc$columns$concu
   exclude_hl_col <- pknca_data$conc$columns$exclude_half.life
 
+  # Define which columns use for the title to keep it short
+  grp_cols_int <- group_vars(pknca_data)
```
m-kolomanski: question: What does `_int` stand for here? Is it needed?

nitpick: I don't see the benefit of shortening the name, could as well be `groups_*` as opposed to `grp_*`

m-kolomanski (commented):
Plotly does have [functionality of generating tables](https://plotly.com/r/table/) with some support for html tags, although you cannnot just pass html table. It might be possible to combine that using [subplots](https://plotly.com/r/subplots/), but I have never done that with plotly before.

R/get_halflife_plots.R
m-kolomanski: Yeah, I am aware that it stands for intervals, yet the programming-focused part of my brain keeps firing like _what, that is not an integer???_ :sweat_smile: If we can't find anything more fitting (and INT is official CDSICS nomenclature), we should probably stay with that. Especially that _historically_ R is more stats-oriented language, rather than pure programming language.

Gero1999:
I implemented the changes requested by Mateusz (thanks a lot!). Also I re-checked the branch as an aggressive user and realized that 16 plots might kill a bit of memory and sometimes reactions are not as fluid as I like. So I changed the default display to 8, which I think would be enough for the users

js3110 (changes requested):
Looks nice, except the titles could be shortened further (see my comment)

Gero1999:
> As discussed, lets open a new issue to filter unique based on intervals

https://github.com/pharmaverse/aNCA/issues/990

PR #983, opened 2026-02-11
fix: same session ZIP tree changes not updating exported outputs

### LINKED ISSUE #955: Bug: zip download selection doesn't work in same session
### Description
If you export the zip with all exports, and then in the same session try to export only one item (eg exploration), the output will be with all exports again.

### To reproduce
1. Run NCA
2. Export zip file with ALL files
3. export again, with limited files
4. View zip folders (they are the same)

### Expected behaviour
Zip output should match the selection made.

Gero1999:
## Issue

Closes #955 

## Description

If you export the zip with all exports, and then in the same session try to export only one item (eg exploration), the output will be with all exports again.

## Definition of Done

creates only what is requested in the zip file

## How to test

download multiple zip files by changing the selection

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer


Shaakon35 (changes requested):
@Gero1999 
It's not working,

With dummy data, I export the zip file with PP only

Then I export with ADPP and the slidedeck and the settings, but only the ADPP is there

Issue #867, opened 2026-01-05
Enhancement: Add N to legend/footnote for groups


js3110:
### Description
When grouping and faceting, it would be useful for the user to see the N for each group, added to the legend.
Eg DOSE: 50 (N = 10), or added to the title of the facets.

### Definition of Done
The requirements for the feature to be complete:
- [ ] N added to legend for each group
- [ ] N added to title for each facet.

PR #984, opened 2026-02-11
feat: add N subjects in facet_wide titles

Gero1999:
## Issue

Contributes to #867

## Description

This pull request introduces enhancements to the faceting and subject counting features in the plotting functions, allowing for more informative facet labels and improved handling of subject counts per facet. The changes add support for facet-specific subject counts, update data processing and plotting functions to use this information, and introduce a helper for generating facet labels with subject counts.

Faceting and subject count enhancements:

* Added a `facet_by` parameter to `process_data_mean` and updated its usage in `exploration_meanplot` to support facet-specific subject counts.
* Updated `g_lineplot` to accept a `facet_count_n` parameter, defaulting to `"USUBJID_COUNT"`, and to use a new helper function for generating facet labels with subject counts.
* Added the `.build_facet_labels` function to generate facet labels that include the subject count, and integrated it into the plotting workflow.
* Modified faceting in `g_lineplot` to use the new `facet_label_var`, ensuring facet labels display both grouping values and subject counts. (`R/g_lineplot.R`)

## Definition of Done

- [x] N added to title for each facet.

## How to test

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

Please, for meanplots notice that if you subset/facet too much, groups will start dissapearing for representation due to the lack of points per time point (<3 points to do an average)

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.1.0.9072
+Version: 0.1.0.9073
 Authors@R: c(
     person("Ercan", "Suekuer", email = "ercan.suekuer@roche.com", role = "aut",
            comment = c(ORCID = "0009-0001-1626-1526")),


Shaakon35:
The issue https://github.com/pharmaverse/aNCA/issues/867 mentions adding this (n=) in the legend as well. Could we do this, please? only for the mean plot

js3110 (commented):
I don't think its worth adding this as a widget and letting the user customise. I think we should just include the N all the time for each group/facet. That is what the original issue was describing

PR #991, opened 2026-02-16
Fix:  slope selector rules in settings upload

### LINKED ISSUE #966: Bug: Slope selector rules do not update in App settings & get_settings_code
### Description
When you upload the settings with some specified slope selector rules the settings won't update 

### To reproduce
Map data > Make exclusions/selection in `Slope Selector` > Run NCA > Save settings

Then start a new session and: Upload settings > Check `Slope Selector`. Rules are not there

### Expected behavior
- [ ] After uploading settings, slope selector rules should appear appropriately

js3110:
## Issue

Closes #966 

## Description

Following update of slope selector in #956 , settings upload didn't work with slope selectors. This has been fixed to ensure that the UI and the results are upadted if the settings are uploaded with slopes.

## Definition of Done

- [x] Slope selector works with settings upload
- [x] Create settings code also works 

## How to test

Try with different data and different slope selections in slope selector tab.
Export both settings via Setup tab and settings via ZIP export, upload them and check that slopes have been applied correctly.
Try to generate settings code and check that it matches the R script generated in the export.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

There were also some inconsistencies in creating the `userData$slope_rules` so this has been cleaned in this PR

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.1.0.9070
+Version: 0.1.0.9071
 Authors@R: c(
     person("Ercan", "Suekuer", email = "ercan.suekuer@roche.com", role = "aut",
            comment = c(ORCID = "0009-0001-1626-1526")),


Gero1999 (approved):
It works nice for me! 😉  A non-blocking comment:

- I think we need to strongly think if it won't be better to decouple data upload and settings upload. For me it still feels weird to do both at the same time, and more than once I am bothered by the fact that I need to put both files within the exact same folder

Shaakon35:
> It works nice for me! 😉 A non-blocking comment:
> 
> * I think we need to strongly think if it won't be better to decouple data upload and settings upload. For me it still feels weird to do both at the same time, and more than once I am bothered by the fact that I need to put both files within the exact same folder

Agree. We will change this

Shaakon35 (approved):

PR #1040, opened 2026-03-04
feat: parameter exclusion from TLG summaries via PPSUMFL/PPSUMRSN in ADPP (#1019)

### LINKED ISSUE #1019: Enhancement: New tab: pk parameter exclusion from TLGs
### Description
Similarly to the manual exclusion tab, we need to have ADPP shown, where we can select in red the pk parameters that we exclude from the summary tables and mean plots in the TLGs.

once the exclusion is done, the variable ANL01FL (row excluded from summary tables and mean plots) shoudl be created in ADPP (therefore ADPP is updated)

### Definition of Done
- [ ] function that performs...
- [ ] button that triggers calculating...
- [ ] application should display...

Shaakon35:
## Issue

Closes #1019

## Description

Adds a **Parameter Exclusions** tab to NCA > Results. Users can exclude individual PK parameter rows from descriptive statistics, mean plots, and ADPP export.

### What it does

- New **Parameter Exclusions** tab in the Results sidebar.
- Displays NCA results in a selectable reactable with PPSUMFL/PPSUMRSN columns.
- PPSUMFL/PPSUMRSN are auto-populated from PKNCA flag rules (e.g., R2ADJ < 0.8). Users can add manual exclusions on top.
- Excluded rows (PPSUMFL = "Y") are highlighted yellow and filtered from descriptive statistics and mean plots.
- "Show excluded points" checkbox in Parameter Plots overlays excluded rows as cross-shaped (✗) points.
- ADPP export includes PPSUMFL/PPSUMRSN flags for both auto and manual exclusions.

### Flag semantics

- `PPSUMFL = "Y"` → record is **excluded** from summary statistics
- `PPSUMFL = ""` → record is **included** in summary statistics
- `PPSUMRSN` → exclusion reason (concatenated with "; " when multiple reasons apply)

### Data flow

1. User selects rows and provides a reason in the Parameter Exclusions tab.
2. `parameter_exclusions_server` returns exclusion indices/reasons as a reactive.
3. `tab_nca.R` computes two views via `.apply_param_exclusions()`:
   - **Tagged**: full data with `.pp_excl`/`.pp_excl_reason` markers (for CDISC export and show-excluded toggle).
   - **Filtered**: excluded rows removed (for descriptive stats).
4. `parameter_plots_server` receives both base results (for picker initialization) and tagged results (for the plot). Pickers don't re-initialize on exclusion changes.
5. `export_cdisc()` merges `.pp_excl`/`.pp_excl_reason` into the `exclude` column, then derives PPSUMFL/PPSUMRSN.

### Shared helpers

Exclusion helpers shared between general exclusions and parameter exclusions live in `inst/shiny/functions/utils-exclusions.R`:
- `.build_exclusion_reasons` — reason-per-index builder
- `.render_exclusion_table` — unified reactable renderer (with optional Type column)
- `.register_remove_observers` — remove-button observer pattern
- `.apply_param_exclusions` — tagged/filtered view builder
- Color constants: `EXCL_COLOR_PARAM`, `EXCL_COLOR_NCA`, `EXCL_COLOR_TLG`, `EXCL_COLOR_BOTH`

## Definition of Done

- Parameter Exclusions tab available in NCA > Results.
- PPSUMFL/PPSUMRSN columns shown in the exclusions table, auto-populated from flag rules.
- Manual exclusions layered on top of auto-populated values.
- Excluded rows highlighted yellow, filtered from descriptive statistics and mean plots.
- Show excluded points toggle works for both flag-rule and manual exclusions.
- ADPP export includes PPSUMFL/PPSUMRSN flags.

## How to test

1. Upload a dataset and run NCA with at least one flag rule enabled (e.g., R2ADJ < 0.8).
2. Go to **NCA > Results > Parameter Exclusions**.
3. Verify rows that were flagged by the rule already show `PPSUMFL = "Y"` and the reason in `PPSUMRSN`, and are highlighted yellow.
4. Hover over column headers (PPORRES, PPSUMFL, PPSUMRSN, etc.) — tooltips should appear.
5. Select additional rows, enter a reason, click "Add". Verify the manual reason is appended to `PPSUMRSN` (separated by `"; "` if the row already had an auto-populated reason).
6. Check **Descriptive Statistics** and **Parameter Plots** — manually excluded LAMZ rows should be filtered out. check only LAMZ
7. Check **Parameter Datasets > ADPP** — `PPSUMFL` and `PPSUMRSN` columns should reflect both auto and manual exclusions.
8. Export ZIP and verify ADPP contains the flags. 

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

- The raw `exclude` column from PKNCA is not shown in the parameter exclusions table. Instead, `PPSUMFL` and `PPSUMRSN` are derived from it (same logic as `.add_crit_flags()` in `export_cdisc.R`), giving users a preview that matches the final ADPP output.
- Exclusion helpers are shared between general and parameter exclusion modules via `utils-exclusions.R` to avoid code duplication.


js3110 (commented):
Looks better!

Either as sub-issues or within this issue please add:
- PP exclusions in settings file download & upload
- PP exclusions carried into R script (I just ran it and while the PPSUMFL column is created, its not populated with the exclusions

I would recommend doing it as sub issues that branch from this branch, and we merge them into this one once complete before merging to main

js3110 (commented):
It looks good, I am happy. Jsut a couple of nitpicks/questions and then I will approve

Gero1999 (approved):
LGTM

Gero1999 (changes requested):
wrong PR

Shaakon35:
@Gero1999 if the raw is MISSING (purple), is that expected to have : 

PPSUMFL = Y
PPSUMRSN  = Too few points for half-life 

?

Gero1999:
> @Gero1999 if the raw is MISSING (purple), is that expected to have :
> 
> PPSUMFL = Y PPSUMRSN = Too few points for half-life
> 
> ?

`exclude = "Too few points for half-life"` or whatever. Then `export_cdisc` uses the `exclude` column to create `PKSUMRSN`

And yes we better do the change `PPSUMRSN > PKSUMRSN`. We can do it in a different PR though if it is too much

Shaakon35:
> And yes we better do the change `PPSUMRSN > PKSUMRSN`. We can do it in a different PR though if it is too much
 
Sounds good, I would like to ask Merck or other companies about this.

inst/shiny/modules/tab_nca.R
```diff
@@ -77,6 +81,36 @@ tab_nca_ui <- function(id) {
   )
 }
 
+# Build tagged and filtered views of NCA results based on parameter exclusions.
+# Tagged view adds .pp_excl/.pp_excl_reason markers for CDISC export.
+# Filtered view removes excluded rows for summary tables and plots.
+.apply_param_exclusions <- function(res, excl_info) {
```
js3110: helper functions can be added to utils-exclusions.R in functions folder

js3110 (commented):
Functionally it seems to work well, nice work!

Coding wise, if possible try to harmonise as much as possible between the Parameter selections module and the general exclusions module: as these are very similar. Check if any helper functions can be shared between the two, or if any additional refactoring can be done to share the functions and set up of both to prevent duplicating code

inst/shiny/modules/tab_nca.R
Shaakon35: @js3110 I did a change with ONA, could you have a look please?

Gero1999 (changes requested):
I exclude points in Parameter Exclusions for most CMAX and then:

1) Check Parameter Plots, press checkbox `Show excluded points (x)` and the parameters excluded still don't appear
2) In Parameter Exclusions `PPSUMFL = ""Y"` for the CMAXs that have been excluded, instead of the opposite way
3) Descriptive statistics seems to work well

js3110:
@Gero1999 @Shaakon35 I wonder if the CRITFL is wrong, and should be "Y" for flagged instead of "N"

As PPSUMFL and PKSUMFL are "Y" for flagged and empty for not....

image

If CRIT1 is R2ADJ < 0.7 then I think it should be "Y" if its < 0.7 and empty/ "N" if its not flagged, so I think that currently that is incorrect. We could change the logic in this PR or open a separate issue, I don't mind

js3110 (approved):
Works well for me, the flags seem correct and the plots work well with exclusion.

Only thing to keep in mind is we need to integrate settings as well!

Shaakon35:
### PPSUMFL / PPSUMRSN flag strategy

The exclusion flags in ADPP follow this convention:

| PPSUMFL | Meaning |
|---------|---------|
| `"Y"` | Record is **excluded** from summary statistics |
| `""` (empty) | Record is **included** in summary statistics |

`PPSUMRSN` contains the exclusion reason. When multiple exclusions apply to the same row (e.g., a flag rule + a manual exclusion), reasons are concatenated with `"; "`.

**Sources of exclusions (layered):**
1. **PKNCA flag rules** (auto) — derived from the `exclude` column in PKNCA results (e.g., `"R2ADJ < 0.8"`, `"AUCPEO > 20"`). These also generate CRITy/CRITyFL column pairs.
2. **Manual parameter exclusions** — user-selected rows in the Parameter Exclusions tab. Stored as `.pp_excl` / `.pp_excl_reason` markers on the tagged result, merged into `exclude` at export time.

Both sources feed into the same `PPSUMFL`/`PPSUMRSN` derivation in `.derive_ppsum_flags()` (`R/export_cdisc.R`), so the ADPP output is consistent regardless of exclusion source.

js3110:
currently CRITFL is Y if **not excluded** though, which is the opposite to the others. hence why i suggest maybe changing it

Shaakon35:
> currently CRITFL is Y if **not excluded** though, which is the opposite to the others. hence why i suggest maybe changing it

Issue created:
https://github.com/pharmaverse/aNCA/issues/1278

Gero1999 (approved):
LGTM

PR #1123, opened 2026-03-26
fix: handle missing lambda z and empty data in slope selector (#1122)

### LINKED ISSUE #1122: Slope selector: unhandled errors when plot data is empty or lambda z is not selected
## Problem

The slope selector module has two related error-handling gaps:

### 1. Plot crashes with bare "Error:" when data is empty

When the slope selector plot has no data to display, it crashes and shows a generic `"Error:"` message with no useful information.

**Expected:** Show a text message instead of the broken plot, e.g. _"Plot cannot be displayed: no data available."_

### 2. Missing lambda z parameter causes crash

When lambda z (`half.life`) is not selected in any PK parameters, the slope selector still tries to render but fails because there are no intervals with `half.life == TRUE` (see `slope_selector.R` line ~144: `filter(type_interval == "main", half.life)`). This results in an error since the slope selector cannot function without lambda z.

**Expected:** Show a message in the slope selector UI, e.g. _"Lambda z has not been added to the selected parameters. Add lambda z to use the slope selector."_

## Location

- `inst/shiny/modules/tab_nca/setup/slope_selector.R` — main module
- `R/get_halflife_plots.R` — plot generation

## Suggested approach

- Wrap plot rendering in a `tryCatch` or add a `validate(need(...))` guard so that empty data shows a user-friendly message instead of a crash.
- Add a check for `half.life` presence in the selected intervals before entering the slope selector logic. If absent, render an informational message instead of the plots UI.

Shaakon35:
Blocked by https://github.com/pharmaverse/aNCA/issues/1153

## Description

The slope selector module crashed with a bare `"Error:"` in two scenarios:
1. When lambda z (`half.life`) was not selected in any PK parameters, the interval filter produced zero rows and downstream code failed.
2. When plot data was empty, the UI attempted to render plots on a NULL/empty list.

### Changes

- `inst/shiny/modules/tab_nca/setup/slope_selector.R`:
  - Added early return when filtered intervals are empty (lambda z not selected), resetting state and displaying: *"Lambda z has not been added to the selected parameters. Add lambda z to use the slope selector."*
  - Replaced `req(plot_outputs())` with explicit NULL/empty checks in `renderUI`, displaying: *"Plot cannot be displayed: no data available."*
- Bumped version to 0.1.0.9114
- Added NEWS entry

## Related Issue(s)

Fixes #1122

## How to test

1. Upload a dataset and run NCA **without** selecting lambda z in the PK parameters
2. Navigate to the Slope Selector tab — verify the informational message appears instead of a crash
3. Upload a dataset where concentration data leads to empty plot data — verify a message is shown instead of `"Error:"`

## Contributor checklist

- [x] New logic covered by unit tests (N/A — Shiny UI guards, no testable R/ logic added)
- [x] New logic is documented
- [ ] Settings upload works with the new implementation (if applicable)">


Gero1999 (changes requested):
when i try to do a HL selection (3 points, 2nd profile) the App crashes:


```
Warning: The 'plotly_click' event tied a source ID of 'A' is not registered. In order to obtain this event data, please add `event_register(p, 'plotly_click')` to the plot (`p`) that you wish to obtain event data from.
Warning: The 'plotly_click' event tied a source ID of 'A' is not registered. In order to obtain this event data, please add `event_register(p, 'plotly_click')` to the plot (`p`) that you wish to obtain event data from.
Warning: Error in purrr::pmap: ℹ In index: 2.
Caused by error in `.f()`:
! Cannot both include and exclude half-life points for the same interval
  2: shiny::runApp
```

Gero1999 (approved):
I think is just me who has the error. I tried researching, but I did not manage to find why. The code here does not seem the root anyways, so we can proceed merging

js3110 (commented):
It works fine, but I think the current logic/ fix is overcomplicated for no reason. I have made some suggestions

inst/shiny/modules/tab_nca/setup/slope_selector.R
```diff
@@ -205,10 +214,46 @@ slope_selector_server <- function( # nolint
...
+          if (nrow(main_intervals) == 0) {
+            return(tags$p(
+              class = "text-muted",
+              "Plot cannot be displayed: no data available."
+            ))
+          } else {
+            return(tags$p(
+              class = "text-muted",
+              "Lambda z has not been added to the selected parameters.",
+              "Add lambda z to use the slope selector."
```
js3110:
```suggestion
              "Lambda z half-life has not been added to the selected parameters.",
              "Add LAMZHL to use the slope selector."
```

js3110 (commented):
A few comments:

1. The issue should LAMZHL not LAMZ, however with the current implementation it doesn't work with either, but only gives the LAMZ warning for LAMZHL. IMO it should only be LAMZHL because LAMZ is not required for the plot (LAMZHL can be calculated without LAMZ being selected), however, the plots don't seem to generate. this may need additional debugging and fixing
2. If I deselect ATPTREF, the no data message appears, but if i deselect PCSPEC or PARAM its just blank (no error message though so its not the end of the world)


I have added some suggestions and questions. I haven't run the code or debugged it myself so the suggestions might not be exactly correct, its just somethin to at least look into...

Shaakon35:
Re: @js3110's question on the `plots` NULL check (lines 248-252):

Yes — on initial render, `renderUI` can fire before the `observeEvent` has run. At that point `hl_intervals` has rows (so the interval checks pass), but `plot_outputs()` is still `NULL` (its initial value). Without this guard, the downstream `.[page_search$is_plot_searched()]` would error on `NULL`.

It can also happen if none of the three change conditions (`in_data`, `in_hl_adj`, `in_selected_intervals`) are true — `plot_outputs()` stays at its previous value, which could be `NULL` if this is the first pass.

js3110:
CLosing as handled by #1171

PR #1125, opened 2026-03-26
feat: persist PP exclusions in settings YAML download and upload

### LINKED ISSUE #1110: PP exclusions in settings file download & upload
## Parent

Sub-issue of #1019 (PR #1040)

## Problem

Parameter exclusions (PPSUM1F/PPSUM1R) are not persisted in the settings YAML file. When a user downloads settings and re-uploads them, the PP exclusions are lost.

## Expected behavior

- **Download**: The settings YAML should include the list of excluded parameter rows (indices and reasons).
- **Upload**: When a settings file containing PP exclusions is uploaded, the exclusions should be restored in the Parameter Exclusions panel.

## Branch

Branch from `1019-feat/parameter-exclusions-anl01fl` and merge back into it before merging to main.

js3110:
## Issue

Closes #1110

## Description

Persists parameter exclusions (PPSUM1F/PPSUM1R) in the settings YAML file so they survive download/upload round-trips.

Exclusions are stored by key (grouping columns + PPTESTCD + PPORRES) rather than row index, so they can be validated against the current NCA results on restore. If a result value has changed since the settings were saved, the exclusion is skipped and a warning banner is shown in the Parameter Exclusions panel.

### Flow

1. User uploads settings — exclusions are queued in `pending_override`
2. User clicks Run NCA — `res_nca()` fires, pending exclusions are matched against results by key
3. Matched exclusions are restored with resolved row indices
4. Unmatched exclusions produce a dismissible alert banner listing the entries that couldn't be matched

### Changes

- **`parameter_exclusions.R`**: Accept `param_excl_override` reactive. Store `keys` (data frame of identifying columns) alongside row indices. Deferred restore via `pending_override` + `res_nca()` observer. Key matching via `.match_keys_to_rows()`. Alert banner for mismatches.
- **`tab_nca.R`**: Extract `param_exclusions` from `settings_override`, pass to module, store in `session$userData`.
- **`nca_setup.R`** / **`zip-utils.R`**: Include `param_exclusions` in settings YAML.
- **`exclusion-utils.R`**: Shared utilities for both exclusion modules.
- **`test-parameter_exclusions.R`**: Tests for key extraction, matching, restore, and shared utilities.

## Definition of Done

- Settings YAML download includes PP exclusion entries with key columns
- Uploading a settings file restores exclusions only when key values match current results
- Unmatched exclusions show a warning banner in the Parameter Exclusions panel
- Shared utilities are used by `parameter_exclusions.R`; `general_exclusions.R` migration tracked in #1126

## How to test

1. Upload data and run NCA
2. Go to Parameter Exclusions, exclude some rows with a reason
3. Download settings YAML — verify it contains a `param_exclusions` section with `reason` and `keys` (not row indices)
4. Reload the app, upload the same data, upload the saved settings YAML, run NCA
5. Verify the Parameter Exclusions panel shows the previously excluded rows
6. Edit the YAML to change a `PPORRES` value, re-upload, run NCA — verify a warning banner appears listing the unmatched exclusion
7. Repeat steps 3–5 using the ZIP export instead of standalone settings download

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

- The return type of `parameter_exclusions_server` changed: it now returns a clean list with `reason` + `keys` (no `rows` or `xbtn_id`). The caller (`tab_nca.R`) computes `excl_info` via `.build_exclusion_reasons()`.
- Key matching uses `as.character()` comparison to handle type differences between YAML-parsed values and data frame columns.
- The `pending_override` pattern ensures exclusions are only applied after NCA results exist, avoiding the race condition where the fingerprint observer would immediately clear them."

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.1.0.9114
+Version: 0.1.0.9115
 Authors@R: c(
     person("Ercan", "Suekuer", email = "ercan.suekuer@roche.com", role = "aut",
            comment = c(ORCID = "0009-0001-1626-1526")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 ## Features added
 
+* PP exclusions (PPSUM1F/PPSUM1R) are now included in settings YAML download and restored on upload (#1110)
 * Parameter Exclusions: exclude individual PK parameter rows from descriptive statistics and ADPP export via PPSUM1F/PPSUM1R flags (#1040)
 * Ratio calculations table is now included in settings YAML export and restored on upload, with validation against available parameters and groups (#1091)
 * Data tab filters are now included in the settings YAML file and restored on upload, for both standalone settings download and ZIP export (#1082)
--- a/inst/shiny/functions/exclusion-utils.R
+++ b/inst/shiny/functions/exclusion-utils.R
@@ -1,0 +1,58 @@
+# Shared utilities for exclusion list management.
+# Used by both general_exclusions.R and parameter_exclusions.R.
+
+#' Rehydrate an exclusion override list with fresh button IDs.
+#'
+#' Uploaded settings store exclusions without xbtn_id. This function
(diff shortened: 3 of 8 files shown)


js3110:
Closed because implementation has changed

PR #1147, opened 2026-03-27
fix: units table hides PCSPEC groups with no selected parameters

Gero1999:
## Issue

Closes #1144

## Description

The Parameter Units table hid all rows for a PCSPEC group (e.g., URINE) when no NCA parameters were flagged `TRUE` in the intervals for that specimen. This happened because `rows_to_hide_units_table` used an `inner_join` that coupled group visibility to parameter flags — if no excretion parameters were selected, URINE groups had no matching rows and were hidden entirely.

The fix changes the group filtering to use a `semi_join` against all interval groups, independent of parameter flags. Specimen types now remain visible in the units table as long as they are present in the active intervals.

## Definition of Done

- Units table shows rows for all PCSPEC values present in the filtered intervals, regardless of whether any parameters are currently selected for that specimen type.

## How to test

1. Upload a dataset with both PLASMA and URINE specimens (URINE having only 1 time point per subject).
2. Go to NCA Setup → Settings.
3. Add URINE to the PCSPEC selector.
4. Click "Parameter Units" button.
5. Verify that URINE rows appear in the units table (previously they were hidden).

## Contributor checklist
- [ ] New logic is documented

## Notes to reviewer

Single-file change in `inst/shiny/modules/tab_nca/units_table.R`. The `rows_to_hide_units_table` reactive now derives visible groups from all intervals via `semi_join` instead of coupling them to parameter flags via `inner_join`. No new dependencies or behavioral changes for the PLASMA-only case — the change only affects scenarios where a specimen type has no parameters flagged TRUE in its intervals.

### DIFF
--- a/inst/shiny/modules/tab_nca/units_table.R
+++ b/inst/shiny/modules/tab_nca/units_table.R
@@ -57,21 +57,29 @@ units_table_server <- function(id, mydata) {
       ))
     })
 
-    # Define rows from units table not of interest for the user
+    # Define rows from units table not of interest for the user.
+    # Group filtering (e.g., by PCSPEC) is independent of parameter flags so
+    # that specimen types like URINE remain visible even when no excretion
+    # parameters are currently selected for that specimen.
     rows_to_hide_units_table <- reactive({
       group_cols <- intersect(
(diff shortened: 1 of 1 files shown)


Gero1999:
This seems already fixed in `main`

PR #1171, opened 2026-04-01
fix: slope selector crashes when PK parameters are removed (#1153)

Gero1999:
## Issue

Closes #1153

## Description

The slope selector crashed when PK parameters were deselected because `get_halflife_plots()` and `slope_selector_server` filtered intervals with `filter(type_interval == "main", half.life)`. When the user deselected `half.life` or all its dependent parameters, no intervals passed the filter, causing downstream errors.

### Changes

**`R/get_halflife_plots.R`**
- Broadened the interval filter: keeps any main interval where `half.life` or any half.life-dependent parameter (`PKNCA::get.parameter.deps("half.life")`) is selected.
- Reduces kept intervals to compute **only** `half.life` (sets all other params to `FALSE`, clears `impute` to `NA`). This avoids `PKNCA_impute_method_blq` errors and `tidyr::spread` duplicate-key issues.
- Returns an empty list early when no qualifying intervals remain.

**`inst/shiny/modules/tab_nca/setup/slope_selector.R`**
- Mirrors the broadened filter for change-detection (half.life OR dependent params).
- Shows an informational empty state message instead of crashing when no half-life parameters are selected.

**`inst/shiny/functions/utils-slope_selector.R`**
- `handle_interval_change()`: joins on identity columns only (`start`, `end`, group vars) instead of all columns, so parameter selection changes are treated as updates rather than add+remove.
- `arrange_plots_by_groups()`: early return for empty lists.

**`inst/shiny/www/main.css` + `_slope_selector.scss`**
- Added `.slope-selector-empty-state` styling for the empty state message.

**`tests/testthat/test-get_halflife_plots.R`**
- Updated expected marker colors for the exclusion test (PKNCA may not find a valid lambda.z fit with excluded points).
- Added test: plots are generated when `half.life = FALSE` but a dependent param like `lambda.z` is `TRUE`.
- Added test: returns empty list when no half.life-related parameter is selected.

**`DESCRIPTION` / `NAMESPACE`**
- Removed unused `methods` import; added `PKNCA::get.parameter.deps` and `dplyr::if_any`.

## How to test

1. Load the app with default dummy data
2. Select **Drug A** only
3. Go to Parameter Selection and remove parameters one at a time from: R2ADJ, CLST, CLSTD, CMAX, CMAXD, LAMZHL, LAMZ, LAMZNPT, LAMZLL, R2, LAMZSPN, TLST, TMAX, VSSO
4. Check the slope selector after each removal — it should not crash
5. Remove **all** PK parameters for a study type — slope selector should show an informational message without crashing
6. Keep only a half-life dependent parameter (e.g., LAMZ) without half.life itself — slope plots should still render

## Contributor checklist

- [x] New logic is documented

## Notes to reviewer

- The `pk.nca()` call inside `get_halflife_plots` now runs with `impute = NA` and only `half.life = TRUE` to avoid side-effects from BLQ imputation or other parameter computations.
- `handle_interval_change()` now joins on identity columns only — this is intentional so that toggling a parameter flag triggers a re-render for affected groups rather than being invisible to the diff.


js3110 (commented):
@Gero1999 I don't notice any change in the funcitonality of the app -> the slope selector plots still don't work if I remove the params (I tried R2ADJ, LAMZHL, LAMZ)

js3110 (approved):
'Works for me :)

Shaakon35 (approved):
PERFECT

PR #1195, opened 2026-04-08
feat: store and restore duplicate exclusions in settings YAML (#873)

### LINKED ISSUE #873: Enhancement: Add duplication handling of data to settings
### Description
In order for #722 to be implemented, we need handling of duplicates in data to also be processed in settings file, so that upon upload the user can jump to NCA tab.

### Definition of Done
- [ ] duplicate handling in settings file
- [ ] everything else works as expected

js3110:
## Issue

Closes #873

## Description

Duplicate exclusion resolutions are now persisted in the settings YAML and restored on re-upload. Instead of storing positional row indices (which break if the dataset changes), the implementation stores the key columns that identify each excluded row (`AFRLT`, `STUDYID`, `PCSPEC`, `DOSETRT`, `USUBJID`, `PARAM`, `AVAL`). On re-import, these keys are matched against the current dataset to recover row indices.

**Key changes:**
- `R/apply_mapping.R`: Two new internal helpers — `extract_time_dup_keys()` (dataset → key data.frame) and `match_time_dup_keys()` (key data.frame → row indices)
- `R/readers.R`: `read_settings()` now parses `time_duplicate_keys` from YAML as a data.frame
- `inst/shiny/functions/zip-utils.R` + `inst/shiny/modules/tab_nca/nca_setup.R`: Both export paths include `time_duplicate_keys` in the payload
- `inst/shiny/modules/tab_data.R`: `imported_mapping` reactive carries both `mapping` and `time_duplicate_keys`
- `inst/shiny/modules/tab_data/data_mapping.R`: On import, matches stored keys against `mapped_data()` synchronously inside `processed_data` before `annotate_duplicates()` runs. Warns if stored keys don't match the current dataset. New duplicates not in stored keys trigger the manual resolution modal.

## Definition of Done

- [x] Duplicate exclusion resolutions are included in the exported settings YAML
- [x] `read_settings()` parses the duplicate resolution data
- [x] On settings re-upload, previously resolved duplicates are auto-applied
- [x] New duplicates (not in the stored resolutions) still trigger the manual resolution modal
- [x] Existing tests pass; new tests cover round-trip of duplicate resolutions

## How to test

1. Upload a dataset that contains time duplicates
2. Resolve duplicates in the modal (select rows to keep)
3. Proceed to NCA tab, save settings YAML
4. Reload the app, upload the same dataset + saved settings
5. Verify: no duplicate modal appears, duplicates are auto-resolved
6. Upload a modified dataset (different rows) + same settings
7. Verify: warning notification about unmatched exclusions, modal appears for new/unmatched duplicates

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

- The key-based approach was chosen over storing raw row indices because indices are fragile — they break if the dataset is re-sorted, rows are added/removed, or the file is re-exported. Keys are matched against the actual data on import.


R/apply_mapping.R
```diff
@@ -205,3 +205,58 @@ annotate_duplicates <- function(dataset, time_duplicate_rows = NULL) {
...
+
+  # Build a match key for both sides
+  build_key <- function(df) {
+    do.call(paste, c(df[, cols, drop = FALSE], list(sep = "\x1f")))
+  }
+
+  data_keys <- build_key(dataset)
+  stored_keys <- build_key(keys_df)
+
+  matched <- which(data_keys %in% stored_keys)
```
Shaakon35: `%in%` matches **all** dataset rows whose composite key appears in `stored_keys`. Two issues:

1. If two dataset rows share all 7 key columns (including `AVAL`), a single stored key will match both — potentially excluding a row that wasn't originally excluded.
2. The matched rows aren't validated as actual time duplicates in the current dataset. A stored key could match a non-duplicate row if the dataset changed.

Suggested fix — use a row-level semi-join and validate against current duplicates:

```r
match_time_dup_keys <- function(dataset, keys_df) {
  if (is.null(keys_df) || nrow(keys_df) == 0) return(NULL)
  cols <- intersect(names(keys_df), names(dataset))
  if (length(cols) == 0) return(NULL)

  build_key <- function(df) {
    do.call(paste, c(df[, cols, drop = FALSE], list(sep = "\x1f")))
  }

  data_keys <- build_key(dataset)
  stored_keys <- build_key(keys_df)

  # Match each stored key to at most one dataset row (first occurrence)
  matched <- integer(0)
  remaining_indices <- seq_along(data_keys)
  for (sk in stored_keys) {
    hit <- remaining_indices[data_keys[remaining_indices] == sk]
    if (length(hit) > 0) {
      matched <- c(matched, hit[1])
      remaining_indices <- setdiff(remaining_indices, hit[1])
    }
  }

  if (length(matched) == 0) return(NULL)
  sort(matched)
}
```

This ensures each stored key consumes exactly one dataset row, preventing over-matching when rows share identical key values.

inst/shiny/modules/tab_data/data_mapping.R
```diff
@@ -377,14 +377,48 @@ data_mapping_server <- function(id, adnca_data, imported_mapping, trigger) {
...
+                n_stored - n_matched, n_stored
+              ),
+              type = "warning",
+              duration = 10
+            )
+          }
+          if (!is.null(matched_indices)) {
+            resolved_time_duplicate_rows(matched_indices)
+            dup_rows <- matched_indices
+          }
```
Shaakon35, line 416: Writing to `resolved_time_duplicate_rows()` inside a `reactive()` is a side effect in what should be a pure computation. This can cause re-invalidation issues and is fragile — it only works because of the `if (is.null(dup_rows))` guard.

Suggested fix — extract the import-matching logic into a dedicated `observeEvent` that fires once when settings are loaded, and keep `processed_data` side-effect-free:

```r
# Add this observer BEFORE the processed_data reactive
observeEvent(imported_mapping()$time_duplicate_keys, {
  keys_df <- imported_mapping()$time_duplicate_keys
  if (is.null(keys_df) || nrow(keys_df) == 0) return()
  # Only apply if no manual resolution has happened yet
  if (!is.null(resolved_time_duplicate_rows())) return()
  req(mapped_data())

  matched_indices <- match_time_dup_keys(mapped_data(), keys_df)
  n_stored <- nrow(keys_df)
  n_matched <- length(matched_indices %||% integer(0))
  if (n_matched < n_stored) {
    showNotification(
      sprintf(
        "%d of %d stored duplicate exclusions could not be matched to the current dataset.",
        n_stored - n_matched, n_stored
      ),
      type = "warning",
      duration = 10
    )
  }
  if (!is.null(matched_indices)) {
    resolved_time_duplicate_rows(matched_indices)
  }
}, ignoreNULL = TRUE, once = FALSE)

# Then processed_data stays clean:
processed_data <- reactive({
  req(mapped_data())
  tryCatch(
    {
      result <- annotate_duplicates(mapped_data(), resolved_time_duplicate_rows())
      select(result, any_of(c(names(mapped_data()), "DTYPE")))
    },
    time_duplicate_error = function(e) {
      df_duplicates(e$duplicate_data)
      NULL
    }
  )
}) %>%
  bindEvent(
    list(mapped_data(), resolved_time_duplicate_rows()),
    ignoreInit = FALSE
  )
```

This separates the side effect (setting resolved rows) from the computation (producing processed data).

inst/shiny/modules/tab_data/data_mapping.R
```diff
@@ -377,14 +377,48 @@ data_mapping_server <- function(id, adnca_data, imported_mapping, trigger) {
     resolved_time_duplicate_rows <- reactiveVal(NULL)
     observe({
       session$userData$time_duplicate_rows <- resolved_time_duplicate_rows()
+      # Store key-based representation for settings export
+      session$userData$time_duplicate_keys <- extract_time_dup_keys(
+        mapped_data(), resolved_time_duplicate_rows()
+      )
```
Shaakon35, line 383: `extract_time_dup_keys(mapped_data(), ...)` is called here without guarding against `mapped_data()` being `NULL`. If `resolved_time_duplicate_rows()` fires before `mapped_data()` is available, `extract_time_dup_keys` will receive `NULL` as the dataset and error on `names(NULL)`.

Suggested fix — add `req(mapped_data())` and guard the extraction:

```r
observe({
  session$userData$time_duplicate_rows <- resolved_time_duplicate_rows()
  # Store key-based representation for settings export
  if (!is.null(mapped_data()) && !is.null(resolved_time_duplicate_rows())) {
    session$userData$time_duplicate_keys <- extract_time_dup_keys(
      mapped_data(), resolved_time_duplicate_rows()
    )
  }
})
```

Alternatively, add a `NULL` dataset guard at the top of `extract_time_dup_keys`:

```r
extract_time_dup_keys <- function(dataset, row_indices) {
  if (is.null(dataset) || is.null(row_indices) || length(row_indices) == 0) return(NULL)
  # ...
}
```

R/apply_mapping.R
```diff
@@ -205,3 +205,58 @@ annotate_duplicates <- function(dataset, time_duplicate_rows = NULL) {
 
   dataset[, !names(dataset) %in% c(".is_time_dup", ".dup_group", "ROWID")]
 }
+
+# Columns that identify a time duplicate row
+TIME_DUP_KEY_COLS <- c("AFRLT", "STUDYID", "PCSPEC", "DOSETRT", # nolint: object_name_linter
+                       "USUBJID", "PARAM", "AVAL")
```
Shaakon35: These columns duplicate the grouping columns hardcoded in `annotate_duplicates()` (line ~183: `group_by(AFRLT, STUDYID, PCSPEC, DOSETRT, USUBJID, PARAM)`), plus `AVAL`. If the grouping columns change in `annotate_duplicates`, this list will silently go out of sync.

Suggested fix — define the grouping columns once and derive both:

```r
# Columns that define a time duplicate group
TIME_DUP_GROUP_COLS <- c("AFRLT", "STUDYID", "PCSPEC", "DOSETRT", # nolint: object_name_linter
                         "USUBJID", "PARAM")

# Key columns include AVAL to distinguish rows within a group
TIME_DUP_KEY_COLS <- c(TIME_DUP_GROUP_COLS, "AVAL") # nolint: object_name_linter
```

Then update `annotate_duplicates` to use `TIME_DUP_GROUP_COLS`:

```r
dataset <- dataset %>%
  group_by(across(all_of(TIME_DUP_GROUP_COLS))) %>%
  mutate(
    .is_time_dup = (n() - sum(DTYPE != "")) > 1,
    .dup_group = cur_group_id()
  ) %>%
  ungroup()
```

This keeps the two in sync by construction.

R/apply_mapping.R
```diff
@@ -205,3 +205,58 @@ annotate_duplicates <- function(dataset, time_duplicate_rows = NULL) {
...
+#'
+#' @returns A data.frame with one row per excluded duplicate, containing
+#'   only the key columns. Returns NULL if `row_indices` is NULL or empty.
+#'
+#' @keywords internal
+#' @noRd
+extract_time_dup_keys <- function(dataset, row_indices) {
+  if (is.null(row_indices) || length(row_indices) == 0) return(NULL)
+  cols <- intersect(TIME_DUP_KEY_COLS, names(dataset))
+  dataset[row_indices, cols, drop = FALSE]
```
Shaakon35: `dataset[row_indices, cols, drop = FALSE]` preserves the original row names (e.g., `"3"`, `"7"`). After a YAML round-trip via `.convert_list_to_df`, row names reset to `"1"`, `"2"`, etc. This doesn't affect correctness since `match_time_dup_keys` only uses column values, but it makes debugging and testing round-trip equality harder.

Suggested fix — reset row names before returning:

```r
extract_time_dup_keys <- function(dataset, row_indices) {
  if (is.null(row_indices) || length(row_indices) == 0) return(NULL)
  cols <- intersect(TIME_DUP_KEY_COLS, names(dataset))
  result <- dataset[row_indices, cols, drop = FALSE]
  rownames(result) <- NULL
  result
}
```

Shaakon35 (approved):
I'll approve in advance as I may not have much time next week.
It looks good, and it works :)

R/apply_mapping.R
Gero1999: +1 this is also my concern

Gero1999: but I think it still works when I test, at least for the case I self-made

Gero1999: maybe something that can be done to make sure this never is changed is add a test for this edge-case particularly @js3110

Gero1999 (approved):
Overall LGTM and works, good job! 

I have some similar concerns with what Valentin said regarding the code: 
- match_time_dup_keys uses `which(data_keys %in% stored_keys)` can create problems, if not now in the future. Adding a test at least to control the case would be nice

Gero1999 (commented):
LGTM

PR #1197, opened 2026-04-08
fix: use NA_character_ for impute column to prevent type mismatch

Shaakon35:
## Issue

Closes #1196
Closes #1192

## Description

Four fixes for crashes and warnings when NCA intervals have no matching concentration data:

1. **impute type mismatch** (`intervals.R`) — `format_pkncadata_intervals` set `data$impute <- NA` (logical). When `create_start_impute` returned early due to empty intervals (e.g. all ATPTREFs deselected), `impute` stayed logical. The slope selector's `anti_join` then failed comparing character vs logical `impute`. Fix: `data$impute <- NA_character_`.

2. **dropped intervals in `create_start_impute`** (`create_start_impute.R`) — Intervals with no matching concentration data were silently dropped, causing downstream mismatches when the slope selector compared old vs new interval sets. Fix: preserve unmatched intervals with `impute = NA_character_` instead of dropping them.

3. **missing PPORRES in half-life plots** (`get_halflife_plots.R`) — When `pk.nca()` returns a 0-row result with units enabled, PKNCA's `pknca_unit_conversion` creates `PPSTRES` via `ret$PPORRES * ret$conversion_factor` but `PPORRES` was never in the empty data frame (it's `NULL`), so `PPSTRES` is created while `PPORRES` is not. The subsequent `tidyr::spread(key="PPTESTCD", value="PPORRES")` then crashes. Fix: add a symmetric guard creating `PPORRES` from `PPSTRES` when missing, plus early returns for 0-row results at two levels (inside `.merge_conc_with_nca_results` and in `get_halflife_plots`).

4. **merge crash on extra_vars** (`get_halflife_plots.R`) — `.merge_conc_with_nca_results` added `extra_vars` (e.g. `ATPTREF`) to `merge_by` after checking they exist in `conc$data`, but never checked they exist in the pivoted wide NCA output. When `ATPTREF` was missing from `wide_output`, `merge()` crashed with `'by' must specify a uniquely valid column`. Fix: intersect `extra` with `names(wide_output)` before using it in `merge_by`.

## Definition of Done

- Deselecting all ATPTREFs in NCA setup no longer crashes with impute type mismatch
- `get_halflife_plots` no longer crashes when `pk.nca()` returns a 0-row result with units
- Intervals with no matching concentration data are preserved, not silently dropped
- Half-life merge no longer crashes when extra_vars columns are absent from the wide NCA output

## How to test

1. Upload data and map columns
2. Go to NCA Setup, deselect all ATPTREFs → no `anti_join` error
3. Upload data with an analyte group where all `AVAL` values are `NA`
4. Navigate to NCA > Slope Selector → no `PPORRES` error, no `fix.by` error
5. Upload data where some profiles have no concentration data within their interval range → warnings are suppressed, no crash

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

Four independent defensive fixes. The impute fix (`intervals.R`) is the core fix for #1196. The `create_start_impute` change preserves intervals that previously vanished. The three guards in `get_halflife_plots.R` form a chain protecting against 0-row PKNCA results: PPORRES guard → 0-row early return → merge_by guard.


js3110:
The `NA_character_` fix on line 284 of `intervals.R` is correct and should be kept — that's a genuine type inconsistency.

However, the defensive coercion guards in `update_main_intervals` and `handle_interval_change` are treating a symptom rather than the root cause. The reason `impute` can end up as logical in the first place is a design problem in `create_start_impute`: when the merge+filter of concentration data against intervals produces matches for **some** intervals but not others, the function replaces `pknca_data$intervals` entirely with the processed merge result (line 71). Any original interval that didn't match the merge+filter is **silently dropped**. The intervals that survive get a character `impute` column, but the ones that don't match simply disappear.

This matters in practice. For example, in a multi-subject dataset where one subject's concentration samples all fall outside the interval range (e.g., all samples beyond TRTRINT), that subject's interval is silently dropped while other subjects' intervals survive with a character `impute`. The dropped interval is the actual bug — the type mismatch you're seeing downstream is a consequence of the interval being lost and then recreated without `impute` in a later reactive cycle.

The fix should be in `create_start_impute` itself: instead of replacing the intervals table with the merge result, it should compute the `impute` value from matched concentration data and join it back onto the **original** intervals. Intervals with no matching conc data get `impute = NA_character_` and are preserved. This way `create_start_impute` only ever adds the `impute` column — it never filters intervals.

Concretely:
1. Tag original intervals with `INT_ROWID` before the merge
2. After computing impute, track which `INT_ROWID`s were matched
3. Unmatched intervals get `impute = NA_character_`
4. `bind_rows` the matched (with computed impute) and unmatched (with `NA_character_`) sets
5. Drop `INT_ROWID`

With that fix, the `as.character()` guards in `update_main_intervals` and `handle_interval_change` become unnecessary and should be removed — they mask the real problem and add maintenance burden.

The `NA_character_` change on line 284 and the PPORRES guard in `get_halflife_plots.R` are fine to keep as-is.

js3110:
@Shaakon35 I updated it, should ne working now and addressing the root cause not just downstream guards

R/get_halflife_plots.R
```diff
@@ -215,6 +220,12 @@ get_halflife_plots <- function(pknca_data, add_annotations = TRUE,
       o_nca$result$PPSTRESU <- o_nca$result$PPORRESU
     }
   }
+  # pk.nca() with units can return 0-row results containing PPSTRES but not
```
Shaakon35, line 218: Nit: stray blank line splitting this comment in two. Should be a single block:

```r
# pk.nca() with units can return 0-row results containing PPSTRES but not
# PPORRES. Ensure PPORRES exists so as.data.frame(out_format="wide") works.
```

R/get_halflife_plots.R
```diff
@@ -225,13 +236,26 @@ get_halflife_plots <- function(pknca_data, add_annotations = TRUE,
     select(-any_of(c("PPORRESU", "PPSTRESU", "PPSTRES"))) %>%
     mutate(exclude = paste0(na.omit(unique(exclude)), collapse = ". "))
 
+  # If no half-life results were produced (e.g. no conc data in interval range),
```
Shaakon35, line 233: This early return produces a 0-row data frame that lacks the NCA result columns (`lambda.z`, `adj.r.squared`, etc.) the caller expects. It works because `get_halflife_plots` has its own `nrow == 0` guard at line 85 that returns before accessing those columns.

Worth adding a note here like `# Coupled with the nrow(d_conc_with_res) == 0 guard in get_halflife_plots` so a future reader doesn't remove the outer guard without realizing this return type is incomplete.

Shaakon35:
**Optional:** Consider adding a smoke test for `get_halflife_plots` with empty/no-match results. Even a minimal test that constructs a `pknca_data` where all concentrations are `NA` (or intervals don't overlap any samples) and asserts `get_halflife_plots()` returns `list(plots = list(), data = list())` would cover the three new guards in that file and prevent regressions.

Shaakon35:
As I am the author of the PR, I cannot approve.

But I approve 👍

R/create_start_impute.R
```diff
@@ -111,8 +113,22 @@ create_start_impute <- function(pknca_data) {
...
     ungroup() %>%
-    # Select only the columns of interest
-    select(any_of(c("start", "end", group_vars(pknca_data), "impute", names(pknca_data$intervals))))
+    select(any_of(c("start", "end", group_vars(pknca_data), "impute",
+                     names(pknca_data$intervals), "INT_ROWID")))
+
+  # Preserve intervals that had no matching concentration data, with impute = NA
+  matched_rowids <- unique(imputed_intervals$INT_ROWID)
+  unmatched_intervals <- original_intervals %>%
+    filter(!INT_ROWID %in% matched_rowids) %>%
```
Gero1999, line 122: question: for which case is this line, is it needed?
```suggestion
```

R/get_halflife_plots.R
```diff
@@ -225,13 +235,28 @@ get_halflife_plots <- function(pknca_data, add_annotations = TRUE,
...
+  # return an empty data frame so callers get zero plots instead of a merge error.
+  # NOTE: the returned frame lacks NCA columns (lambda.z, adj.r.squared, etc.);
+  # the nrow(d_conc_with_res) == 0 guard in get_halflife_plots handles this.
+  if (nrow(wide_output$result) == 0) {
+    conc_select_cols <- c(group_vars(pknca_data), time_col, conc_col,
+                          timeu_col, concu_col, exclude_hl_col, "ROWID")
```
Gero1999, line 238: Rigth now we have two redundant changes, one inside (not doing anything) and the other outside (doing it all)

I would prefer everything is handled intenrally by this function so in the future the R script can handle it by itself the empty case scenario

Gero1999 (approved):
LGTM, some concerning code aspects mentioned

R/create_start_impute.R
js3110: to filter for the intervals that are not in the imputed intervals (because they were filtered out as they didn't have conc data)

js3110: yes, its needed otherwise all intervals will then be set to NA

R/get_halflife_plots.R
js3110: @Gero1999 fixed, I think. Check out the changes and let me know if this is what you had in mind? Otherwise can you explain a bit mroe detail (reference whic functions/files you are specifically referring to) and I can take a closer look !

Gero1999: That will work, thanks!

Gero1999 (approved):
LGTM!

js3110 (approved):

PR #1201, opened 2026-04-08
Fix: Custom unit conversions not applied to NCA results

Gero1999:
## Issue

Closes #1199

## Description

Three bugs in `nca_results.R` caused custom unit conversions to fail:

1. **Variable shadowing**: The code applied unit conversions to `res` (lines 56-70), but then re-read `res_nca()` into a fresh `results` variable (line 75), discarding the adjustments. Fixed by assigning `results <- res` instead.

2. **Wrong null check**: `!is.null(conversion_factor)` inside `ifelse()` was always `TRUE` because `conversion_factor` is a column vector, never `NULL`. Changed to `!is.na(conversion_factor)`.

3. **Duplicate rows from units join**: The `left_join` used `intersect(names(.), names(units_table))` as join keys, which included value columns like `PPORRESU`. This caused many-to-many joins and duplicate `PPSTRES` values, triggering `pivot_wider` errors. Fixed by explicitly excluding value/unit columns (`PPSTRESU`, `PPSTRES`, `PPORRESU`, `conversion_factor`, `default`) from the join keys and selecting only the needed columns from the units table.

## Definition of Done

- [x] Line 75 uses the unit-adjusted `res` instead of re-reading `res_nca()`
- [x] `!is.null(conversion_factor)` changed to `!is.na(conversion_factor)`
- [x] Units join no longer produces duplicate rows
- [ ] Custom unit conversions are reflected in the NCA results table after saving

## How to test

1. Upload data and run NCA.
2. Go to **NCA Results** → click **Parameter Units**.
3. Change a unit (e.g., Cmax from `ng/mL` to `µg/mL`) and set a conversion factor (e.g., `0.001`).
4. Click **Save**.
5. Verify the NCA results table now shows converted values without errors.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

The three fixes are in separate commits for clarity. The duplicate-rows fix (commit 3) addresses the `pivot_wider` warning about non-unique `PPSTRES` values that appeared when custom units were applied.


Gero1999:
This was already solved in `main`

PR #1181, opened 2026-04-07
fix: export modal missing comment field and download error without NCA

Shaakon35:
## Issue

Closes #1180

## Description

Fixes two regressions in the export modal (`zip.R`) introduced in #1163:

1. **Missing comment field** — The `settings_comment_container` div with a `textInput` was never added to `.show_export_modal`. The `shinyjs::show/hide` observer already toggled it based on `settings_file` selection, but the element didn't exist. Added the hidden container and wired the comment value to `session$userData$settings_save_comment` on confirm.

2. **Download error without NCA** — `res_nca()` was called directly in `.run_export`, throwing a `shiny.silent.error` (empty message) when NCA hasn't been run. Wrapped in `tryCatch(..., error = function(e) NULL)` to match the original pattern on `main`.

## Definition of Done

- Comment field appears in export modal when `settings_file` is selected
- Export from exploration tab works without NCA results

## How to test

1. Upload data, go to Exploration tab, render an individual plot
2. Click Save (top-right), verify the export modal opens
3. Check `settings_file` in the tree → comment field should appear
4. Uncheck `settings_file` → comment field should hide
5. Click Export → download should succeed without error

## Contributor checklist
- [x] Code passes lintr checks
- [x] Code passes all unit tests
- [ ] New logic covered by unit tests
- [x] New logic is documented
- [ ] App or package changes are reflected in NEWS
- [ ] Package version is incremented
- [ ] R script works with the new implementation (if applicable)
- [ ] Settings upload works with the new implementation (if applicable)
- [ ] If any `.scss` change was done, run `data-raw/compile_css.R`

## Notes to reviewer

Both fixes restore behavior that existed on `main` but was lost during the zip.R refactor in #1163.

PR #1214, opened 2026-04-09
fix: wrap res_nca() in tryCatch to prevent download error before NCA

### LINKED ISSUE #1213: Bug: ZIP download fails with .htm error page when exporting before NCA is run
### Description

Clicking **Save** → **Export** after data mapping (before running NCA) produces a file named `zip_modal-download_zip.htm` instead of a ZIP archive. The browser console shows `Download Error:` with an empty message.

The cause is the unprotected `res_nca()` call in `.run_export()` (`inst/shiny/modules/tab_nca/zip.R`). When NCA hasn't been run, `res_nca()` throws a `shiny.silent.error` (via `req()`), which has an empty message. The `tryCatch` in `.run_export` catches and re-throws it, causing the `downloadHandler` to serve an HTML error page instead of the ZIP file.

`prepare_export_files` already handles `res_nca = NULL` correctly — it skips all NCA-dependent exports. The call just needs to be wrapped so the silent error is caught and converted to `NULL`.

### To reproduce

1. Launch the app
2. Upload any dataset and complete data mapping
3. Click the **Save** button (top-right)
4. In the export modal, click **Export ZIP with Results** → then **Export**
5. The downloaded file is `zip_modal-download_zip.htm` instead of a `.zip`

### Expected behaviour

The ZIP file should download successfully, containing only the items available before NCA (settings file, session info).

Gero1999:
## Issue

Closes #1213

## Description

`res_nca()` in `.run_export()` throws a `shiny.silent.error` (empty message) when NCA hasn't been run, causing the `downloadHandler` to serve an HTML error page (`zip_modal-download_zip.htm`) instead of a ZIP file. Wrapping the call in `tryCatch(..., error = function(e) NULL)` converts the error to `NULL`, which `prepare_export_files` already handles by skipping NCA-dependent exports.

## Definition of Done

- Clicking Save → Export after data mapping (before running NCA) downloads a valid ZIP file containing settings and session info.

## How to test

1. Launch the app
2. Upload any dataset and complete data mapping
3. Click **Save** (top-right)
4. In the export modal, click **Export ZIP with Results** → then **Export**
5. Verify a `.zip` file is downloaded (not `.htm`)
6. Verify the ZIP contains `settings.yaml` and `session_info.txt`

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

One-line fix. `prepare_export_files` already has `if (!is.null(res_nca)) { ... }` guarding all NCA-dependent exports, so passing `NULL` is the intended fallback path.">

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,7 @@
 
 ## Features added
 
+* `run_app()` now accepts a `settings` parameter to pre-load a YAML settings file on startup (#514)
 * Exploration sidebars: "View Exports" button opens a scrollable gallery modal showing all saved plots inline with name, type, timestamp headers and a remove option (#1137)
 * Added "Min. Points for Half-life" setting in NCA > Settings > General Settings, allowing users to configure PKNCA's `min.hl.points` option (range 2–10, default 3) (#1155)
 * Settings version control: single YAML file stores multiple versions with metadata (timestamp, comment, dataset, aNCA version, active tab). Save button in header, version selection modal on upload, version delete support (#1103)
--- a/NEWS.md
+++ b/NEWS.md
@@ -54,6 +54,7 @@
   PPTX and HTML exports; box plot parameters are also configurable (#972)
 
 ## Bugs fixed
+* ZIP export before running NCA no longer fails with an `.htm` error page; `res_nca()` is now wrapped so the silent error is caught (#1213)
 * SASS compilation moved from runtime (`app.R`) to a `data-raw/compile_css.R` script, fixing startup crashes on read-only deployments (#1107)
 * ZIP folder with results will now include the exploration tab outputs: individual plots, mean plots (#794)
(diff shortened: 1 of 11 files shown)


Shaakon35 (approved):
Works

NEWS.md
```diff
@@ -55,6 +54,7 @@
   PPTX and HTML exports; box plot parameters are also configurable (#972)
 
 ## Bugs fixed
+* ZIP export before running NCA no longer fails with an `.htm` error page; `res_nca()` is now wrapped so the silent error is caught (#1213)
```
js3110: probably don't need the NEWs since this bug didn't exist in last version

js3110 (approved):
Works for me :)

js3110:
Closing as this issue was solved in #1181

PR #1169, opened 2026-04-01
Fix: use dose-relative times for manual interval parameter renaming

Gero1999:
## Issue

Closes #1168

## Description

Two bugs in the descriptive statistics module:

**1. Parameter columns not displayed in the table.** The parameter selector was refactored from `updatePickerInput` (raw column names like `CMAX[ng/mL]`) to `selector_label` (cleaned names like `CMAX`). But `summary_stats_filtered` still used `select(any_of(input$select_display_parameters))` — the clean names didn't match the actual column names with units, so `any_of()` silently dropped all parameter columns.

**2. Interval parameter renaming mismatch.** Commit `742ca2434` changed `signif(start_dose)` / `signif(end_dose)` to `start` / `end` in `descriptive_statistics.R`. This produced absolute interval times instead of dose-relative times, mismatching `pivot_wider_pknca_results` which uses `signif(start_dose)` / `signif(end_dose)`. The same incorrect convention was propagated to `parameter_plots.R` and `flexible_violinboxplot.R`.

## Changes

- **`descriptive_statistics.R`** — map clean parameter names back to actual column names (with units) before selecting; revert interval renaming to `signif(start_dose)` / `signif(end_dose)`
- **`parameter_plots.R`** — align interval renaming with `pivot_wider_pknca_results`
- **`flexible_violinboxplot.R`** — align interval renaming with `pivot_wider_pknca_results`

## How to test

1. Upload data and run NCA, if possible with multiple profiles
2. Go to **Results > Descriptive Statistics** — verify parameter columns (CMAX, AUC, etc.) show numeric values
3. If manual AUC intervals are configured, verify interval parameters appear correctly (e.g., `AUCINT_0-12`)
4. Go to **Results > Parameter Plots** — verify interval parameters appear in the selector

## Contributor checklist
- [x] Code passes lintr checks
- [x] Code passes all unit tests
- [x] New logic covered by unit tests
- [x] New logic is documented
- [x] App or package changes are reflected in NEWS
- [x] Package version is incremented
- [x] R script works with the new implementation (if applicable)
- [x] Settings upload works with the new implementation (if applicable)

PR #1217, opened 2026-04-10
fix: handle all-NA PPSTRESU groups in descriptive statistics

### LINKED ISSUE #1216: Bug: Descriptive statistics crash when PPSTRESU is all NA for a parameter group
### Description

When `PPSTRESU` is `NA` for all rows in a parameter group (e.g., partial AUC intervals like `RCAMINT` where no result was computed), multiple things break:

**1. Descriptive statistics crash** — `calculate_summary_stats()` crashes at line 67:

```r
ModeUnit = names(sort(table(PPSTRESU), decreasing = TRUE, useNA = "ifany")[1])[1]
```

`table()` drops NAs by default, returning an empty table. `names(sort(empty_table, ...)[1])[1]` returns `NULL`, and `dplyr::mutate` cannot combine `NULL` with character results across groups:

```
Error in mutate:
! `ModeUnit` must return compatible vectors across groups.
✖ Can't combine NULL and non NULL results.
```

**2. Column names get `NA` suffix** — In `pivot_wider_pknca_results.R` and `label_operators.R`, the pattern `ifelse(PPSTRESU != "", ...)` returns `NA` when `PPSTRESU` is `NA` (since `NA != ""` evaluates to `NA`). This produces column names like `RCAMINT_0-20NA` instead of `RCAMINT_0-20`.

Both issues stem from `export_cdisc.R` line 394 setting `PPSTRESU = ifelse(is.na(PPSTRES), NA_character_, PPSTRESU)`, which is correct CDISC behavior but wasn't handled downstream.

### To reproduce

1. Upload a dataset with both SERUM and URINE specimens.
2. In NCA settings, add partial AUC intervals (e.g., AUCINT and RCAMINT).
3. Run NCA computation.
4. In **NCA Results > Select Parameters**, observe `RCAMINT_0-20NA` instead of `RCAMINT_0-20`.
5. Navigate to **Descriptive Statistics** — app crashes.

The issue is on `main`.

First reported in https://github.com/pharmaverse/aNCA/pull/1169#pullrequestreview-4081159606.

### Expected behaviour

- Descriptive statistics should handle groups where all unit values are `NA` without crashing.
- Manual interval parameters should appear as `RCAMINT_0-20` (without `NA` suffix) in NCA results.

Gero1999:
## Issue

Closes #1216

## Description

When `PPSTRESU` is `NA` for all rows in a parameter group (common with partial AUC intervals like RCAMINT where no result was computed), two things break:

1. **Descriptive statistics crash** — `table(PPSTRESU)` drops NAs, returning an empty table. `names(sort(empty_table, ...)[1])[1]` returns `NULL`, which `dplyr::mutate` can't combine with character results from other groups.

2. **Column names get `NA` suffix** — `ifelse(PPSTRESU != "", ...)` returns `NA` when `PPSTRESU` is `NA`, producing names like `RCAMINT_0-20NA`.

**Fixes:**
- Extract `.mode_unit()` helper in `calculate_summary_stats.R` that returns `NA_character_` for empty/all-NA input
- Guard `PPTESTCD` formatting against `NA` `ModeUnit`
- Add `!is.na(PPSTRESU)` guard to all `ifelse(PPSTRESU != "", ...)` patterns in `pivot_wider_pknca_results.R` (4 locations) and `label_operators.R` (2 locations)

## Definition of Done

- [x] `calculate_summary_stats()` no longer crashes with all-NA unit groups
- [x] Manual interval parameters appear as `RCAMINT_0-20` (not `RCAMINT_0-20NA`)
- [x] Parameters with no units appear without a unit suffix in the output
- [x] Existing behavior unchanged for groups with valid units

## How to test

1. Upload a dataset with both SERUM and URINE specimens
2. In NCA settings, add partial AUC intervals (AUCINT and RCAMINT)
3. Run NCA computation
4. In **NCA Results > Select Parameters** — verify RCAMINT appears as `RCAMINT_0-20` (no `NA` suffix)
5. Navigate to **Descriptive Statistics** — should render without error
6. Verify that RCAMINT columns appear without unit brackets, while other parameters retain their units

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

First reported in https://github.com/pharmaverse/aNCA/pull/1169#pullrequestreview-4081159606. The root cause is that `export_cdisc.R` correctly sets `PPSTRESU = NA` when `PPSTRES` is `NA` (CDISC SD0027), but downstream code assumed `PPSTRESU` is always a non-NA string. The fix adds NA guards at all 7 affected locations across 3 files."


Shaakon35:
image

RCAMINT does not appear.

Also can you fix the typo in tab_nca.R:

```
msg <- paste0(

"Urine Parameters (FREXINT, RCAMINT) calculated for non-urine samples",
"will return NA values and units"
```

-> use **paste** to add a " "

Gero1999:
@Shaakon35 did you include `Urine` in the `NCA setup`?

js3110 (approved):
LGTM

Shaakon35:
I tried again, added Plasma and Urine.

Descriptive stats:
image

NCA results:
RCAMINT is there two times, but I asked it only once:
image

Can you fix that please? :)
Except if that is correct? but I doubt

Gero1999:
hey @Shaakon35, it is "correct". The problem is that right now when you request an interval parameter you do it for all `PCSPEC` (SERUM, URINE) even if you request urine interval parameters (i.e, `RCAMINT`). Therefore, the results show two columns, one for the URINE, which has the expected units, and one for the SERUM, which is not possible to calculate and has NA units.

I changed it so now the NA units show in a clean way:

image


However, notice that this is merely aesthetic. Let me know if you think something else should be done

Shaakon35 (approved):
Works well!

Is it normal to have two RCAMINT when I only asked for one 0 20?

Image

Gero1999:
@Shaakon35 `RCAMINT` should be there if you left it checked in `Selection Parameters`. It is by default calculated

PR #1223, opened 2026-04-14
Parameter selection matrix UI

Gero1999:
## Issue

Related to the collapsible parameter sections discussion — while working on that, I realised that with our current capabilities (especially the PK parameters reference modal) we could revisit the parameter selection UX that we hadn't managed to land before.

## Description

Replaces the accordion-based `virtualSelectInput` parameter selectors and the read-only summary table with a single interactive **checkbox matrix**:

- **Rows** = study types, **columns** = parameters (PPTESTCD), grouped by TYPE
- Hovering a column header shows the full parameter name (PPTEST)
- Clicking a cell toggles whether a parameter is calculated for that study type
- TYPE group headers are color-coded: Standard (blue), Urine (gold), Sparse (purple)
- **Select all** / **Clear all** buttons to bulk-toggle all parameters
- The study types summary table is moved into a collapsed accordion at the bottom

The module return interface (`selections`, `types_df`) is unchanged — no modifications needed in `nca_setup.R` or any downstream consumer.

## Known limitation (out of scope)

Selecting **all** parameters and running NCA can crash the app with `Can't recycle input of size N to size M` in `pivot_wider_pknca_results()`. This is a **pre-existing server-side issue** in how results are pivoted — not caused by this PR. The old UI simply didn't expose a way to select all parameters at once. Tracked in #1231.

## Definition of Done

- Parameter matrix renders correctly with study types as rows and parameters as columns
- Checkbox clicks update the correct parameter for the correct study type
- Scroll position is preserved when toggling checkboxes
- Settings upload correctly overrides checkbox state
- Clear all unchecks everything and NCA results reflect the empty selection
- Default parameter selections on fresh data load match previous behavior

## How to test

1. **Checkbox accuracy**: Clear all → check one parameter (e.g., CMAX) for one study type → Run NCA → verify results contain exactly that parameter
2. **Cross-study isolation**: Select CMAX for type A, CAVG for type B → Run NCA → verify each type only has its selected parameter
3. **Settings round-trip**: Make selections → Download settings → Reload app → Upload settings → verify matrix matches → Run NCA → verify results match
4. **Clear all**: Make several selections → Click "Clear all" → verify all checkboxes are unchecked visually → re-select a few → Run NCA → verify only re-selected params appear
5. **Scroll stability**: Scroll to the right of the matrix → click several checkboxes quickly → scroll should not jump
6. **Defaults**: Load fresh data (no settings file) → verify default params are pre-checked and respect study type rules (e.g., CAVG not checked for single-dose)

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

The diff looks large but most changes are **purely aesthetic** (CSS/SCSS). Here's what actually matters:

| File | What changed |
|------|-------------|
| `parameter_selection.R` | **The only functional change.** UI switched from accordion + virtualSelectInput to an HTML checkbox table. Server logic simplified — removed sub-module orchestration, added single click handler. All helper functions (`.apply_parameter_selections`, `.build_param_ref_data`, etc.) are unchanged. |
| `parameter_selector_panel.R` | **Deleted.** The virtualSelectInput sub-module is no longer needed. |
| `_parameter_matrix.scss` | New SCSS partial — purely styling. |
| `main.css` | Compiled CSS from the SCSS partial. Removed unused `.nca-pill-*` classes. |
| `main.scss` | One import line added. |

**TL;DR**: Review `parameter_selection.R` for logic. Everything else is CSS.


js3110 (commented):
It looks really good! and works super well :D

Some small suggestions:
- Sometimes the table takes a while to load, but there is no indication that it would be there- could you add a loading spinner in the space so that the user can tell that its rendering?
- I like the clear all button, would also be nice to have a select all button!

Gero1999:
> Sometimes the table takes a while to load, but there is no indication that it would be there- could you add a loading spinner in the space so that the user can tell that its rendering?

Yes this is a good idea!

> I like the clear all button, would also be nice to have a select all button!

@js3110 This one you mean all parameters for all studies? I feel like it might generate easily problems in calculations no? There has to be something that could be nice in this regard but I can't think of it

js3110:
> > Sometimes the table takes a while to load, but there is no indication that it would be there- could you add a loading spinner in the space so that the user can tell that its rendering?
> 
> Yes this is a good idea!
> 
> > I like the clear all button, would also be nice to have a select all button!
> 
> @js3110 This one you mean all parameters for all studies? I feel like it might generate easily problems in calculations no? There has to be something that could be nice in this regard but I can't think of it

I think the user would be aware that they can't calculate all params for all (and if they try, they will just get error messages but no crash) , but I am thinking more that its nice if they know they want to calculate a lot of parameters, its good to have the option to either clear all or select all, to then only click a few more boxes afterwards- if that makes sense? then its up to the user whether they use it or not- and we are responsible for ensuring that the app doesnt crash with this

js3110 (approved):
LGTM! thanks :)

Shaakon35 (approved):
All good for me :)

Can you please add a button Reset? (next to Select all, and unselect all).
the button reset would reset according to the study type

Issue #721, opened 2025-10-20
Enhancement: Add mapping used during session to settings


Shaakon35:
### Description
Enhance the settings functionality to include the mapping information used during the user's session. This will allow for easier reapplication of mapping if the settings are reused, reducing redundant mapping efforts for users.

### Definition of Done
- [ ] Add in the settings the mapping used during the session.
- [ ] Ensure the mapping information is saved and correctly restored when settings are uploaded.

Issue #718, opened 2025-10-20
Enhancement: Settings upload in the tab "Data"


Shaakon35:
### Description

Essentially, the current settings upload is not the best as you still have to re-map your dataset to go through the app.
I would like to rethink the way it's currently done.

Open an issue to put the settings file at the beggening, skip 1 2 3 4 directly etc.

### Definition of Done
The requirements for the feature to be complete:
- [ ] the settings button is now together with the data upload, you can have a dataset with or without settings #719 
- [ ] add in the settings the mapping used during the session #721 
- [ ] if dataset and settings are uploaded, then: #722 

1. Best option: run the code to apply the mapping, and move the user to the NCA tab directly
OR
2. Alternative: pre-populated the select input with the mapping done during the previous session

#722 
Blocked by automation of R script: https://github.com/pharmaverse/aNCA/issues/467

PR #1226, opened 2026-04-15
fix: read mapping, filters, ratio_table, and units from YAML in get_settings_code()

Gero1999:
## Issue

Closes #1189
Contributes to #1187
- [x] R script template derives `extra_vars_to_keep` from `mapping$Grouping_Variables` instead of reading a separate field
- [x] `get_settings_code()` produces scripts where `extra_vars_to_keep` is correctly populated from the YAML mapping

## Description

`get_settings_code()` used a hardcoded `default_mapping` parameter and ignored `filters`, `ratio_table`, `units_table`, and `extra_vars_to_keep` stored in the settings YAML. Generated R scripts would use wrong mapping and skip filters, ratios, and unit overrides.

**Changes:**
- Read `mapping`, `filters`, `ratio_table`, `units_table` from the YAML when present
- Derive `extra_vars_to_keep` from `mapping$select_Grouping_Variables` (consistent with the UI path in `nca_results.R`)
- Read `time_duplicate_keys` from the YAML instead of hardcoding `NULL`
- Rename `default_mapping` to `.default_mapping` (internal, not exported) and use it as fallback for legacy YAML files
- Remove the `mapping` parameter from the public API

## Definition of Done

- [x] `get_settings_code()` reads mapping from YAML, falls back to `.default_mapping` for legacy files
- [x] `get_settings_code()` includes `applied_filters` from YAML in the session object
- [x] `get_settings_code()` includes `ratio_table` from YAML in the session object
- [x] `get_settings_code()` includes `units_table` from YAML in the session object
- [x] `get_settings_code()` derives `extra_vars_to_keep` from mapping
- [x] Generated R scripts correctly apply mapping, filters, units, and ratio calculations

## How to test

1. Generate a settings YAML from the app (with custom mapping, filters, ratio table, and units)
2. Call `get_settings_code(settings_file_path, data_path)` — no `mapping` argument needed
3. Inspect the generated R script: it should contain the custom mapping values, filters, ratio table, and units from the YAML
4. Test with a legacy YAML (no `mapping` key) — should fall back to CDISC defaults

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

- This is a **breaking change** for anyone calling `get_settings_code(mapping = ...)` explicitly. The `mapping` parameter has been removed since the YAML already contains the mapping (since #721). Legacy YAML files without a `mapping` key still work via `.default_mapping` fallback.
- R is not available in the dev container, so tests could not be run locally. CI will validate.
- Related to #718 (parent), #1187 (extra_vars_to_keep), #1188 (units_table path). This PR addresses the `get_settings_code()` side; the script template side (#1187) and units format (#1188) may need separate follow-ups.


js3110 (commented):
extra vars to keep are different- in settings code:

Image

vs in session code:

Image

js3110 (approved):
Thanks, looks good now :)

Shaakon35:
Looks good, but I need to test a bit more

Shaakon35 (approved):

PR #1232, opened 2026-04-16
Fix prevent pivot_wider crash when all parameters are selected

### LINKED ISSUE #1231: App crashes when all NCA parameters are selected
### Description

Selecting all NCA parameters and running NCA causes the app to crash with:

```
Error in [<-: Can't recycle input of size 996 to size 83.
```

preceded by:

```
Warning: Values from `PPSTRES` are not uniquely identified; output will contain list-cols.
```

### Root Cause

When all parameters are selected, PKNCA produces result rows with `PPTESTCD = NA` and `PPSTRES = NA`. These are internal dependency calculations that PKNCA runs but doesn't expose as named parameters.

In `pivot_wider_pknca_results()`, these NA-keyed rows create duplicate keys for `pivot_wider()` (multiple NA rows per subject/interval), which produces list-columns instead of atomic vectors. Downstream code then crashes trying to assign list-columns into a regular data frame.

### Additional finding: missing metadata mappings

Comparing `PKNCA::get.interval.cols()` with `metadata_nca_parameters`, **4 PKNCA parameters have no CDISC mapping**:

| PKNCA parameter | Description |
|---|---|
| `clr.last` | Renal clearance (to Tlast) |
| `clr.obs` | Renal clearance (observed, using AUCinf.obs) |
| `clr.pred` | Renal clearance (predicted, using AUCinf.pred) |
| `count_conc_measured` | Count of measured (non-imputed) concentrations |

These are not the direct cause of the crash (the NA rows are), but they should be added to the metadata for completeness. When selected, `translate_terms()` returns the raw PKNCA name unchanged, which could cause issues in CDISC exports.

### Debugging evidence

```r
# 24 duplicate rows, all with PPTESTCD = NA
dupes %>% select(USUBJID, PPTESTCD, PPSTRES, start, end, DOSNOA) %>% print(n = 10)
# A tibble: 24 x 6
   USUBJID PPTESTCD PPSTRES start   end DOSNOA
   <chr>   <chr>      <dbl> <dbl> <dbl>  <int>
 1 S1-01   NA            NA     0  24.1      1
 2 S1-01   NA            NA     0  24.1      1
 3 S1-01   NA            NA     0  24.1      1
 4 S1-01   NA            NA     0  24.1      1
 ...
```

### Steps to Reproduce

1. Upload data with multiple study types
2. Go to Parameter Selection -> click \"Select all\"
3. Click \"Run NCA\"
4. App crashes

### Fix

Filter out `NA` PPTESTCD rows before pivoting in `pivot_wider_pknca_results()`. See PR #1232.

Separately, consider adding the 4 missing parameters to `metadata_nca_parameters` for completeness."

Gero1999:
## Issue

Closes #1231

## Description

When all NCA parameters are selected, PKNCA calculates internal dependency parameters that have no CDISC mapping. These appear in the results as rows with `PPTESTCD = NA` and `PPSTRES = NA`. When `pivot_wider()` encounters multiple NA-keyed rows per subject/interval, it produces list-columns, and downstream code crashes with `Can't recycle input of size 996 to size 83`.

**Root cause:** PKNCA dependency parameters with no CDISC mapping produce `NA` PPTESTCD values that are not filtered before pivoting.

**Fix:** Filter out `NA` PPTESTCD rows from `myres$result` once, before any `pivot_wider()` call. These rows carry no useful output (both key and value are NA).

## How to test

1. Upload data → Parameter Selection → \"Select all\" → Run NCA
2. Should complete without crashing
3. Verify results table shows all selected parameters with valid PPTESTCD codes

## Contributor checklist
- [ ] New logic is documented

## Notes to reviewer

Single-line fix in `R/pivot_wider_pknca_results.R`: `filter(!is.na(PPTESTCD))` applied once before any pivoting. 6 lines added total (including comment)."


Gero1999:
this works but I still don't like the solution, I will keep checking

Gero1999:
@js3110 this one seems that it has indeed been solved already in `main`, so I will close it as well as the issue

PR #1235, opened 2026-04-17
Fix: set end time of interval to last time point in last dose scenario

### LINKED ISSUE #1234: Bug: Final dose interval should include all time points not just cut off at TAU
### Description
For the final dose of a study, if TRTRINT is available, the interval is set to start = Dose time, end = TRTRINT. But actually it should be end = last time point in order to get the full time profile and accurate clast etc. The only parameter where end should = TAU is AUCTAU for this scenario

### To reproduce
Check with a dataset where the final profile has datapoints that are > TAU

### Expected behaviour
End should = last time point

js3110:
## Issue

Closes #1234 

## Description

Previously, if TRTRINT was available, the end time for a final dose was set to TRTRINT. But this sometimes cuts off the final points in cases where the samples are taken for longer. Feedback from scientists advised to set the end time to be the last time point, except for the calculation of AUCTAU (currently not possible in aNCA anyway).

## Definition of Done

- [x] end time set to last time point for last dose
- [x] all other intervals remain the same

## How to test

Check the intervals, especially for last doses. Use a dataset where the samples are taken after the end of the TRTRINT.

## Contributor checklist
- [x] New logic is documented


R/intervals.R
```diff
@@ -102,13 +102,12 @@ format_pkncadata_intervals <- function(pknca_conc,
...
-    # Make end based on next dose time (if no more, TRTRINT or last NFRLT)
+    # Make end based on next dose time (if no more, last NFRLT)
     mutate(end = if (has_trtrint) {
       case_when(
         !is.na(lead(!!sym(time_column))) ~ lead(!!sym(time_column)),
-        is.na(TRTRINT) & is_one_dose ~ Inf,
-        is.na(TRTRINT) ~ start + max_end,
-        TRUE ~ start + TRTRINT
+        is.na(TRTRINT) & is_one_dose ~ Inf, #TAU = NA assumes single dose
+        TRUE ~ start + max_end
```
Gero1999, line 111: question: isn't `start + max_end` = `max(AFRLT, na.rm = TRUE)` ?

Gero1999 (approved):
LGTM!

tests/testthat/test-intervals.R
```diff
@@ -92,7 +92,7 @@ describe("format_pkncadata_intervals", {
 
     result_tau <- format_pkncadata_intervals(pknca_conc_tau, pknca_dose)
 
-    expect_equal(result_tau$end[4], 10)
+    expect_equal(result_tau$end[4], 9)
```
Shaakon35, line 95: This test has `TRTRINT = 5` and `max(ARRLT) = 4`, so `max_end < TRTRINT`. The assertion `end == 9` (i.e., `start + max_end`) confirms the interval stops at the last sample, not at tau. But the relationship between `max_end` and `TRTRINT` is implicit — a reader can't tell whether this is the interesting case or just a coincidence.

Consider adding an explicit test that makes the intent clear:

```r
it("ends at last sample time even when samples stop before tau", {
  # TRTRINT = 10 but last sample at ARRLT = 4
  # → end should be start + max_end = 5 + 4 = 9, not start + TRTRINT = 5 + 10 = 15
  df_conc_long_tau <- df_conc %>%
    mutate(TRTRINT = 10)

  pknca_conc_long_tau <- PKNCA::PKNCAconc(
    df_conc_long_tau,
    formula = AVAL ~ AFRLT | STUDYID + PCSPEC + DOSETRT + USUBJID / PARAM,
    exclude_half.life = "exclude_half.life",
    time.nominal = "NFRLT"
  )

  result <- format_pkncadata_intervals(pknca_conc_long_tau, pknca_dose)

  # Last dose starts at 5, last sample at ARRLT = 4 → end = 9
  expect_equal(result$end[4], 9)
})
```

This documents the design decision: the interval always extends to the last observed sample, regardless of tau.

R/intervals.R
```diff
@@ -102,13 +102,12 @@ format_pkncadata_intervals <- function(pknca_conc,
...
     arrange(start) %>%
-    # Make end based on next dose time (if no more, TRTRINT or last NFRLT)
+    # Make end based on next dose time (if no more, last NFRLT)
     mutate(end = if (has_trtrint) {
       case_when(
         !is.na(lead(!!sym(time_column))) ~ lead(!!sym(time_column)),
-        is.na(TRTRINT) & is_one_dose ~ Inf,
-        is.na(TRTRINT) ~ start + max_end,
-        TRUE ~ start + TRTRINT
+        is.na(TRTRINT) & is_one_dose ~ Inf, #TAU = NA assumes single dose
```
Shaakon35: Nit: missing space before `#`.

```r
# Before:
is.na(TRTRINT) & is_one_dose ~ Inf, #TAU = NA assumes single dose

# After:
is.na(TRTRINT) & is_one_dose ~ Inf, # TAU = NA assumes single dose
```

Shaakon35 (approved):

R/intervals.R
js3110: in many cases yes, but not always, so its safer to keep it this way incase of different sampling windows.

PR #1239, opened 2026-04-20
UI: Consolidate parameter-related settings into Parameter Selection tab

### LINKED ISSUE #1238: Consolidate parameter-related settings into Parameter Selection tab
### Current state

Parameter-related configuration is spread across two tabs:
- **Parameter Selection** — parameter matrix (study types × parameters)
- **Settings** — Partial Interval Calculations, Ratio Calculations, Units button

This forces users to switch tabs to configure related settings.

### Proposed change

Move all parameter-related settings into the **Parameter Selection** tab:

1. **Partial Interval Calculations** → collapsible accordion below the matrix
2. **Ratio Calculations** → collapsible accordion below intervals
3. **Units button** → header row next to the PK parameter details button

Order below the matrix: Intervals → Ratios → Detected Study Types (existing).

Rename the **Settings** tab to **General Settings** (it will only contain General Settings, Data Imputation, and Flagging Rules).

### Motivation

All parameter-level configuration lives in one place. The user picks which parameters, which intervals, which ratios, and which units without switching tabs.

Gero1999:
## Issue

Closes #1238

## Description

Moves all parameter-related configuration into the **Parameter Selection** tab so users don't need to switch tabs to configure what gets calculated:

- **Partial Interval Calculations** → collapsible accordion below the parameter matrix
- **Ratio Calculations** → collapsible accordion below intervals
- **Units button** → header row, next to the PK parameter details button
- **Detected Study Types** → remains at the bottom (collapsible, unchanged)

The **Settings** tab is renamed to **General Settings** and now only contains: General Settings, Data Imputation, Flag Rule Sets, and the Download settings button.

All accordion sections in Parameter Selection are collapsed by default to keep the page clean.

## How to test

1. **Layout**: Open the app → NCA Setup → verify \"General Settings\" and \"Parameter Selection\" tabs exist
2. **Parameter Selection tab**: Verify the matrix is at the top, followed by collapsed accordions for Partial Interval Calculations, Ratio Calculations, and Detected Study Types. Units and PK parameter details buttons should be in the header row.
3. **Partial Intervals**: Expand the accordion → Add Row → fill in an interval → Run NCA → verify the interval appears in results
4. **Ratio Calculations**: Expand the accordion → configure a ratio → Run NCA → verify ratio appears in results
5. **Units button**: Click it → verify the units modal opens and works as before
6. **General Settings tab**: Verify it contains General Settings, Data Imputation, Flag Rule Sets, and Download settings button — no Partial Intervals or Ratios
7. **Settings round-trip**: Download settings → reload → upload → verify all settings (including intervals and ratios) are restored

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

**No server logic was changed.** All changes are UI reorganization:

| File | What changed |
|------|-------------|
| `settings.R` | Extracted `partial_intervals_ui()` as standalone function. Removed the accordion panel from `settings_ui()`. Fixed stale accordion `open` parameter. |
| `parameter_selection.R` | Added `units_ui`, `intervals_ui`, `ratios_ui` parameters. Renders them in the header row and as accordion panels. Updated help dropdown. |
| `nca_setup.R` | Renamed tab to \"General Settings\". Passes UI elements to `parameter_selection_ui()` via args. Removed ratios/units from Settings tab. |

Server wiring is preserved because all UI elements keep their original namespace IDs (e.g., `partial_intervals_ui(ns(\"nca_settings\"))` matches `settings_server(\"nca_settings\", ...)`)."


js3110 (approved):
Great idea! makes much more sense, and looks good and works well :D

AngelaManginelli:
Hi Gero,
first, this is a huge improvement! Looks very good. I have a few points: 
1. Minor. I do not like much that you have the Tab "General Setting" and then within it again the collapsible accordion with same name "General Setting". But this is just my taste and do not have a better idea, perhaps "Parameter Setting" but dunno.
2. Minor. The download setting button is ONLY in the "General Setting" Tab but if I understand correctly you can set other stuff in the other tabs, and then you have to go back to "General Setting" to download everything. One idea could be to put the "Download Setting" as header but above all accordion.
3. Issue!! I downloaded a setting in which I set 
Choose the NCA Profile: Dose 2 
but when I re-uploaded it it was back to Dose 1 as default.
Cheers
Angela.

Gero1999:
@AngelaManginelli thanks for the comments! Remember that to provide a formal review you always need to do `Files changed > Submit Review > Approve/Request-changes`. Regarding the points you raised:


> 1. Minor. I do not like much that you have the Tab "General Setting" and then within it again the collapsible accordion with same name "General Setting". But this is just my taste and do not have a better idea, perhaps "Parameter Setting" but dunno.

I agree, I will change `General Settings` > `NCA Settings`

> 2. Minor. The download setting button is ONLY in the "General Setting" Tab but if I understand correctly you can set other stuff in the other tabs, and then you have to go back to "General Setting" to download everything. One idea could be to put the "Download Setting" as header but above all accordion.

This is very sharp of you, we recently had a UX discussion on this topic. Right now this button is a bit redundant, because you can anytime in the App do `Save > settings_file > Export` so we will probably remove it. But this will be done in a different branch once decided.

> 3. Issue!! I downloaded a setting in which I set
>    Choose the NCA Profile: Dose 2
>    but when I re-uploaded it it was back to Dose 1 as default.

This is a good catch, the bug is coming from `main` (not related with these changes) and already being addressed apart in #1244

Gero1999:
image

AngelaManginelli (approved):
Hi Gero,
thanks for your answer.
1. Very good!
2. I see, but this bring me to the follow: I have seen the button "Save" and in fact I wondered if this would also save the setting, but if you go with the mouse on the button, the help says "Export all selected results" and you only see the option if you click on it. Suggest to add "and setting", then probably fine to remove the Download button for setting, even though I still find it a bit more straight forward to have them separated (more practical). But your decision.
3. Yes, you are right, in fact I should have asked to open it myself but I have seen you already opened it. 
For me, we can approve this. 
Great job!
Angela.

Gero1999:
Thanks @AngelaManginelli, I opened an issue for 2. (#1253). Feel free to populate it, we can also use it for discussion threads. Later on you if you want you can take on it once you finish the website task 💪🏽

Shaakon35 (commented):
Would be great to rename variable: 
PCSPEC -> Specimen 

and add "Short Parameter" , so that they can relate with the short parameters names in the table below

image

Like here:
image

Shaakon35 (approved):

PR #1244, opened 2026-04-22
fix: prevent pcspec and profile from reverting after settings upload

### LINKED ISSUE #1227: Bug: pcspec widget reverts to default after settings upload
## Description

When uploading settings that include a specific pcspec (specimen) selection, the pcspec picker briefly shows the correct imported value but then reverts to the default (e.g., "PLASMA").

## Root Cause

The bidirectional cascade between the analyte and pcspec observers in `inst/shiny/modules/tab_nca/setup/settings.R` causes a double-trigger that loses the imported pcspec value.

The sequence:

1. `settings_override()` fires → `pending_settings` is set with imported settings
2. Analyte observer fires → calls `.consume_settings()` (reads **and clears** `pending_settings`) → sets pcspec to the imported value via `updatePickerInput` → sets profile ✅
3. Pcspec observer fires (triggered by step 2's `updatePickerInput`) → calls `.consume_settings()` → gets `NULL` (already consumed) → updates analyte picker with current value (no-op selection)
4. Analyte observer fires **again** (triggered by step 3's `updatePickerInput` for analyte) → calls `.consume_settings()` → gets `NULL` → computes pcspec with no override → falls back to `default_logic` (grep for "plasma"/"serum") → **overwrites the imported pcspec** ❌

The `updating_filters` guard doesn't prevent this because each observer resets the flag via `on.exit()` before the next observer fires (they run on separate flush cycles since `updatePickerInput` is async).

## Expected Behavior

The pcspec picker should retain the imported value from settings after the cascade settles.

## Possible Fix

The `.consume_settings()` pattern (read-once-then-clear) doesn't work with bidirectional cascades that span multiple flush cycles. Options:

1. **Don't clear `pending_settings` until the full cascade settles.** Use a debounce or a counter to track how many cascade rounds have occurred, and only clear after the cascade is idle.
2. **Use a flag to skip the second analyte observer fire.** After the analyte observer runs with settings, set a flag that prevents the next analyte observer fire from recomputing pcspec.
3. **Prevent the pcspec observer from re-triggering the analyte observer when the selection hasn't actually changed.** Compare `target_analyte` with `isolate(input$select_analyte)` and skip the `updatePickerInput` call if they match.

Option 3 is the simplest and most targeted fix.

## Steps to Reproduce

1. Open the app and upload a dataset
2. Configure NCA settings with a non-default pcspec (e.g., "URINE" instead of "PLASMA")
3. Save settings to YAML
4. Reload the app, upload the same dataset with the saved settings
5. Observe that the pcspec picker shows the default value instead of the saved one

js3110:
## Issue

Closes #1227

## Description

The bidirectional cascade between the analyte and pcspec observers in `settings.R` was clearing `pending_settings` on first read (via `.consume_settings()`), which meant the pcspec observer — firing on the next flush cycle — had no access to the imported override values. This caused both the specimen and profile pickers to revert to defaults after settings upload.

The fix moves the clear from a consume-on-read helper to the pcspec observer, which is the last observer in the cascade that needs the override values. The analyte observer now reads `pending_settings()` without clearing it.

## Definition of Done

- Specimen picker retains the imported value after settings upload
- Profile picker retains the imported value after settings upload
- Normal user-driven filter changes still work correctly

## How to test

1. Open the app and upload a dataset
2. Configure NCA settings with a non-default specimen (e.g. \"URINE\") and select specific profiles
3. Save settings to YAML
4. Reload the app, upload the same dataset with the saved settings
5. Verify that both the specimen and profile pickers show the saved values

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

The `.consume_settings()` helper has been removed. `pending_settings()` is now read directly by both observers, and only cleared in the pcspec observer after it has applied all overrides (including profile). No timers or debounce logic needed — the fix relies on the known cascade order: analyte observer always fires before pcspec observer.
">


Gero1999 (approved):
LPTM! :)

Shaakon35 (approved):
Great! :)

PR #1229, opened 2026-04-16
feat: add dose-normalised summary slides to PPT/QMD export (#1054)

bachapman:
## Issue

Closes #1054

## Description

Adds opt-in dose-normalised summary slides to the PPT/QMD export. Each dose group gains a **Dose-Normalised Plot & Statistics** slide showing a dose-normalised mean concentration-time plot alongside a summary statistics table (filtered to user-selected dose-norm parameters).

Key changes:
- `get_dose_esc_results()` extended with `dose_norm_parameters` argument; computes `dose_norm_meanplot` and `dose_norm_statistics` per group using existing `exploration_meanplot(..., line_type = "dose-normalized")` and `calculate_summary_stats()` helpers
- New **Dose-normalised parameters** virtual selector in the Customise Slides modal, grouped by parameter type (Concentration / AUC / AUMC), defaulting to CMAXD, AUCLSTD, AUCIFOD when available in the current run
- Validation gates on whether dose-norm parameters are available — no spurious error when dose-norm params were not computed
- Renders through both PPTX (`officer-utils.R`) and QMD (`quarto-utils.R`) export paths

## Definition of Done

- Dose-Normalised Plot & Statistics section appears in the Summary Slides tree of the Customise Slides modal
- Dose-normalised parameters selector shows grouped choices and defaults to CMAXD / AUCLSTD / AUCIFOD
- Exported PPTX and QMD each contain a dose-norm slide per group when the section is selected
- No validation error fires when the NCA run does not include dose-norm parameters

## How to test

1. Run NCA with at least one dose-normalised parameter (e.g. CMAXD, AUCLSTD, AUCIFOD) selected
2. Open **Save Results → Export**, enable slides, click **Customise Slides**
3. Confirm **Dose-Normalised Plot & Statistics** appears under Summary Slides; tick it
4. Confirm the **Dose-normalised parameters** selector appears, defaults to CMAXD/AUCLSTD/AUCIFOD, and groups choices by Concentration / AUC / AUMC
5. Clear all selections from the **Dose-normalised parameters** selector — confirm a validation error appears and the Export button is disabled
6. Export as PPTX — verify a dose-norm plot + stats slide is present for each group
7. Export as QMD — verify the `.qmd` references `dose_norm_meanplot` and `dose_norm_statistics`
8. Deselect the section, re-export — confirm dose-norm slides are absent
9. Run NCA without dose-norm parameters — open Customise Slides and confirm no validation error appears

## Contributor checklist
- [x] Code passes lintr checks
[...]

PR #1245, opened 2026-04-22
fix: dose-aware AUCint parameters — rename PPTESTCDs and populate PPANMETH in CDISC export

### LINKED ISSUE #1242: Bug: dose-aware AUCint parameters have misleading PPTESTCD and PPTEST
### Description

Four dose-aware AUCint parameters in `metadata_nca_parameters.csv` have PPTESTCD codes and PPTEST labels that suggest dose normalization, when they actually represent dose-aware interpolation/extrapolation of concentrations at interval boundaries.

The `D` suffix in their PPTESTCD (e.g., `AUCINTD`, `AUCINTAD`) follows the convention used by actual dose-normalized parameters (`CMAXD`, `AUCLSTD`, `AUCIFOD`), making them indistinguishable. This caused confusion in #1229 where they were incorrectly picked up as dose-normalized parameters for the dose-norm slides feature.

### Current vs Expected

| PKNCA | Current PPTESTCD | Current PPTEST | Issue |
|---|---|---|---|
| `aucint.last.dose` | AUCINTD | AUC from T1 to T2 Normalized by Dose | Not dose-normalized — uses dose-aware interpolation |
| `aucint.all.dose` | AUCINTAD | AUCint (based on AUCall extrapolation, dose-aware) | `D` suffix suggests dose-normalization |
| `aucint.inf.obs.dose` | AUCINTID | AUCint (based on AUCinf,obs extrapolation, dose-aware) | `D` suffix suggests dose-normalization |
| `aucint.inf.pred.dose` | AUCINTPD | AUCint (based on AUCinf,pred extrapolation, dose-aware) | `D` suffix suggests dose-normalization |

These parameters use PKNCA's dose-aware AUC calculation (`pk.calc.aucint.*` with dose timing), which interpolates/extrapolates concentrations using dose information from adjacent intervals. This is different from dose normalization (dividing by dose).

Additionally, `AUCINTD` has an incorrect PPTEST — \"AUC from T1 to T2 Normalized by Dose\" — which is factually wrong for a dose-aware parameter.

### Proposed Changes

Rename the PPTESTCD codes to clearly distinguish them from dose-normalized parameters, and correct the PPTEST labels. The new names should indicate \"dose-aware\" rather than implying \"dose-normalized\".

### Files to Update

- `data-raw/metadata_nca_parameters.csv` — fix PPTESTCD, PPTEST, and input_names columns
- `data/metadata_nca_parameters.rda` — regenerate via `Rscript data-raw/metadata_nca_parameters.R`
- `inst/shiny/modules/tab_nca/setup/settings.R` — update the help table that lists interval parameters

### Related

- Discovered in #1229 (dose-normalised slides PR)
- Same file as #1236 (CAVG interval PPTESTCD fixes)

Gero1999:
## Issue

Closes #1242

## Description

Dose-aware AUCint parameters (`aucint.last.dose`, `aucint.all.dose`, etc.) compute the same AUC as their non-dose-aware counterparts but stop interpolation/extrapolation when the next dose occurs. Previously they had PPTESTCDs with a `D` suffix (e.g. `AUCINTD`) that was indistinguishable from dose-*normalized* parameters (`CMAXD`, `AUCLSTD`). The `AUCINTD` PPTEST label was also wrong — it said \"Normalized by Dose\". They have actually nothing to do with dose-normalized parameters...

### Approach

**Metadata-driven PPANMETH**: A new `PPANMETH` column in `metadata_nca_parameters.csv` declares the analytical method for parameters that need it. During CDISC export, `.apply_metadata_ppanmeth()` looks up this column and for any parameter with a non-empty PPANMETH:
1. Strips the trailing lowercase suffix from PPTESTCD (e.g. `AUCINTda` → `AUCINT`)
2. Populates `PPANMETH = \"Interpolation truncated at next dose time\"` in the PP/ADPP output

This is fully generic — no hardcoded maps. Future parameters with analytical method variants just need a PPANMETH value in the CSV and a lowercase suffix in their PPTESTCD.

### Key changes

| File | Change |
|---|---|
| `data-raw/metadata_nca_parameters.csv` | Add `PPANMETH` column; rename dose-aware PPTESTCDs (`AUCINTD` → `AUCINTda`, etc.); fix PPTEST label |
| `R/export_cdisc.R` | Add `.apply_metadata_ppanmeth()` — metadata-driven PPTESTCD consolidation and PPANMETH population |
| `inst/shiny/modules/tab_nca/setup/settings.R` | Update help table with new PPTESTCDs and labels |
| `inst/www/templates/settings_template.yaml` | Update default partial interval parameter |
| `tests/` | Update test data and assertions for new PPTESTCDs; add PPANMETH consolidation test |

### Internal vs CDISC PPTESTCDs

| Internal (app/pivoted table) | CDISC export (PP/ADPP) | PPANMETH |
|---|---|---|
| `AUCINTda` | `AUCINT` | `Interpolation truncated at next dose time` |
| `AUCINTAda` | `AUCINTA` | `Interpolation truncated at next dose time` |
| `AUCINTISda` | `AUCINTIS` | `Interpolation truncated at next dose time` |
| `AUCINTIPda` | `AUCINTIP` | `Interpolation truncated at next dose time` |

### ⚠️ Requires manual step

`metadata_nca_parameters.rda` needs regeneration: `Rscript data-raw/metadata_nca_parameters.R`

## How to test

1. Upload data, configure partial intervals with both `AUCINT` and `AUCINTda` for the same range
2. Run NCA → verify pivoted results show `AUCINT_0-24` and `AUCINTda_0-24` as separate columns
3. Export ZIP → check PP CSV: both should have `PPTESTCD = AUCINT`, dose-aware rows have `PPANMETH = Interpolation truncated at next dose time`
4. Verify ratio calculations work with dose-aware interval parameters

## Contributor checklist
- [x] New logic is documented
- [x] If a package dependency was added/changed, run `data-raw/test_suggests_hidden.R`"


js3110:
@Gero1999 question: what dose Dose-aware mean?

Regarding point 4- these parameters were not incorrectly picked up in #1229 , as the unit type column is correct, so this point is redundant from DoD and How to Test

Gero1999:
We can ask @billdenney, but I think it is just to make dose interpolations/extrapolations to provide more accurate AUCINTs

Point 4 was removed

js3110:
@Gero1999 is the DESCRIPTION update in this branch intentional or accidentally carried over from another branch? seems not relevant to this PR

Gero1999:
Had to be carried over accidentally while bumping, I will correct

js3110 (approved):
LGTM

Shaakon35:
I looked at this issue, and I am not sure if the New PPTESTCD exists at all

Gero1999:
@Shaakon35  Neither the PPTESTCD nor the PPTEST mentioned exist (CDISC = FALSE), they are just meant for the user to know what they refer to

Shaakon35:
Okay I understand now.

I won't accept this PR, we are changing the CDISC standard to fit the app, it's not good.

Instead I would just filter out the 4 PPTESTCD that are annoying from the button to export the pdf.
What do you think about that?

Gero1999:
@Shaakon35 we have more parameters that are not CDISC standard in the App, but I am happy removing these specifically from the App completely. I will proceed

billdenney:
I'm jumping in to give some context.

The CDISC standard does not have all the parameters which PKNCA calculates. One simple example is that I don't think that CDISC has a "number of points above the limit of quantification" parameter, but it is used extensively for validation that an AUC is reasonable to calculate. We keep it in PKNCA output so that people can use it for quality checks (similar to span ratio), and it improves the traceability for exclusion rationale rather than forcing the reviewer to go back to the original concentration-time profiles.

When we generate PP or ADPP within Human Predictions, we add it with the extensible option for PPTESTCD/PPTEST. Ideally, we would have it and others added to the controlled terminology. It is usually reasonable to exclude those parameters from output, but the context they provide is the reason they exist.

Shaakon35 (approved):

js3110:
@Gero1999 @Shaakon35 I wouldn't remove the parameters from the app completely, I would just remove them from the CDISC sections if thats possible?
Some users may want to calculate these parameters for exploratory purposes, just not reporting

Maybe its a separate issue but at least in this PR I would keep the parameters, not remove them

Shaakon35:
> @Gero1999 @Shaakon35 I wouldn't remove the parameters from the app completely, I would just remove them from the CDISC sections if thats possible? Some users may want to calculate these parameters for exploratory purposes, just not reporting
> 
> Maybe its a separate issue but at least in this PR I would keep the parameters, not remove them

I agree, let's do that, it's cleaner and allow more exploration of these 4 parameters

Gero1999:
@js3110 @Shaakon35 I will then proceed removing them from `Selection Parameter` and include them in the `Interval Parameters` which is where they belong. I will also specify more about their calculation in the guidelines and results so users can have a grasp of what they are

Gero1999:
## Diagnosis: "Restoring session…" hang on settings YAML upload

I traced the full auto-replay pipeline (`data_upload → .start_auto_replay → mapping → filtering → preview → pknca_data → auto_replay_ready → app.R dismiss / NCA auto-run`) and reviewed every code change in this PR that affects runtime behavior. Below are the findings.

---

### 1. Extra reactive cycle from `manual_slopes` NULL → empty-df change

**Files:** `manual_slopes_table.R`, `readers.R`, `utils-slope_selector.R`

On `main`, `manual_slopes` stays `NULL` until the user adds a rule or a non-empty override is applied. On this branch, the `observeEvent(pknca_data(), ...)` in `manual_slopes_table_server` now **always initializes `manual_slopes` to an empty data frame** before checking the override:

```r
# NEW (this PR) — lines 62-71 of manual_slopes_table.R
manual_slopes(initial_manual_slopes)   # ← sets empty df BEFORE checking override

if (!is.null(manual_slopes_override())) {
  if (...nrow == 0) return(NULL)        # ← returns, but manual_slopes is already set
  ...
}
```

This is compounded by the `.normalize_optional_df` → `.convert_list_to_df` change in `readers.R`: empty `slope_rules` from the YAML are now an empty data frame instead of `NULL`, so `!is.null(manual_slopes_override())` is `TRUE` (it was `FALSE` on `main`).

**Effect:** `slope_rules()` (which IS `manual_slopes`) changes from `NULL` → empty df. Since `processed_pknca_data` depends on `slope_rules()`, it re-evaluates. This triggers `slope_selector_server` → `detect_pknca_data_changes` → plot regeneration (because `old` pknca_data was just set). The cycle should stabilize after one extra round (the `req(is.null(manual_slopes()), ...)` guard prevents re-initialization), but it doubles the initial computation cost and, for large datasets, could push the pipeline past the 15-second safety timeout window.

### 2. Backward-incompatible PPTESTCD renames in `int_parameters`

**Files:** `metadata_nca_parameters.csv`, `settings_template.yaml`

Old YAMLs (generated before this PR) contain `AUCINTD`, `AUCINTAD`, `AUCINTID`, `AUCINTPD`. These PPTESTCDs no longer exist in the metadata. When the YAML is restored:

1. `.restore_non_filter_settings` sets `int_parameters()` to the YAML data (with old names)
2. The `reactable` renders with `dropdown_extra(choices = PARTIAL_INT_PARAMS$PPTESTCD)` — the old names aren't in the choices
3. `update_main_intervals` calls `translate_terms(parameter, "PPTESTCD", "PKNCA")` — old names aren't found, so they pass through unchanged
4. `!!sym("AUCINTD") := TRUE` creates a new column in the intervals that PKNCA doesn't recognize

This won't cause an R-side error (PKNCA ignores unknown columns), but the `dropdown_extra` JS widget receiving a value not in its choices list could trigger repeated edit events as it tries to reconcile the mismatch, causing `int_parameters()` → `settings()` → debounce → `processed_pknca_data` invalidation cycles.

### 3. `int_parameters` always included in settings (even when empty)

**File:** `settings.R` line 483

```r
# main:
int_parameters = if (nrow(int_parameters()) > 0) int_parameters() else NULL,

# this PR:
int_parameters = int_parameters(),
```

Combined with the `readers.R` change (`.normalize_optional_df` removed), empty `int_parameters` is now an empty tibble instead of `NULL` throughout the pipeline. Code that checks `is.null(int_parameters)` (e.g., `update_main_intervals` line 225) now takes the non-NULL branch and calls `translate_terms` on an empty data frame — harmless, but a behavioral change.

### 4. Safety timeout may not fire in time

The 15-second safety timeout in `.setup_auto_replay` (tab_data.R) uses `shinyjs::delay`, which depends on the R process being free to receive the JS callback. If the extra reactive cycle from finding #1 keeps R busy with plot regeneration (`get_halflife_plots` for many subjects), the timeout callback is queued but can't execute until R is free. Meanwhile, the modal stays visible.

The 10-second NCA auto-run timeout in `tab_nca.R` has the same limitation.

---

### Suggested fixes

1. **Guard the `manual_slopes` initialization** — only set it to an empty df if the override is NULL or truly empty. Move the initialization inside the else branch:
   ```r
   observeEvent(pknca_data(), {
     req(is.null(manual_slopes()), slopes_pknca_groups())
     if (!is.null(manual_slopes_override()) &&
           is.data.frame(manual_slopes_override()) &&
           nrow(manual_slopes_override()) > 0) {
       # ... apply override ...
       manual_slopes(manual_slopes_override())
     } else {
       # Initialize empty only when no override applies
       manual_slopes(initial_manual_slopes)
     }
   })
   ```

2. **Add a PPTESTCD migration step** in `read_settings` / `.process_settings_payload` that maps old names to new ones:
   ```r
   rename_map <- c(AUCINTD = "AUCINTda", AUCINTAD = "AUCINTAda",
                    AUCINTID = "AUCINTISda", AUCINTPD = "AUCINTIPda")
   if (!is.null(obj$settings$int_parameters)) {
     obj$settings$int_parameters$parameter <- ifelse(
       obj$settings$int_parameters$parameter %in% names(rename_map),
       rename_map[obj$settings$int_parameters$parameter],
       obj$settings$int_parameters$parameter
     )
   }
   ```

3. **Keep the NULL-coercion for empty collections** — restore the `if (nrow == 0) NULL` behavior for `slope_rules` and `ratio_table` in `readers.R`, or add explicit `nrow > 0` guards at the consumption sites. The `int_parameters` change (always include, even empty) is fine since `update_main_intervals` handles it correctly.

4. **Add `reset_reactable_memory()`** to the `int_parameters` restore path in `.restore_non_filter_settings` (settings.R line 615) to prevent stale JS-side state from firing spurious edit events.

Gero1999:
as @Belubbit realized, the `export_cdisc` was showing the PPANMETH in the `SDTM.PP` but not in the `ADAM.ADPP`. I tried addressing this and I included it in the PR description as well

Gero1999:
### ADPP export fix: `PPANMETH` was missing from the ADPP variable list

`.apply_metadata_ppanmeth()` consolidates dose-aware PPTESTCDs (e.g. `AUCINTda` → `AUCINT`) and populates `PPANMETH` on the shared `cdisc_info` data frame — **before** the PP/ADPP split. However, `PPANMETH` was only listed in the PP variable metadata (`metadata_nca_variables.csv`), not in ADPP. So the `select(any_of(CDISC_COLS$ADPP$Variable))` call silently dropped the column, leaving ADPP with duplicate `PPTESTCD = AUCINT` rows and no way to distinguish dose-aware from non-dose-aware results.

**Fix** (commit a435228): Added `PPANMETH` to the ADPP section of `data-raw/metadata_nca_variables.csv`. Both PP and ADPP now retain the column after export.

Belubbit (approved):
Works well now

daviesmeg (approved):
Looks good to me, works as specified!

Gero1999:
All contributor checklist work as well. I also did a general review and seems to work all fine. I will proceed to merge

PR #1254, opened 2026-04-23
fix: eliminate spurious warnings in rm_impute_obs_params

### LINKED ISSUE #1218: Test warnings in test-PKNCA.R: uninitialised impute column and no matching intervals
## Description

Several tests in `test-PKNCA.R` (lines 277–465) emit warnings during `PKNCA_update_data_object` calls:

1. **`Unknown or uninitialised column: 'impute'`** — When `start_impute = FALSE`, `create_start_impute` is skipped so the `impute` column is never added to `data$intervals`. Then `rm_impute_obs_params` accesses `data$intervals$impute` on a tibble, which warns.

2. **`No intervals found with the specified target parameters, groups and/or impute method. No changes made.`** — When `start_impute = TRUE` but no `parameter_selections` is passed, all NCA parameter flags on the intervals remain `FALSE`. `rm_impute_obs_params` then calls `interval_remove_impute` for each impute method, but `identify_target_rows` finds no matching intervals → warning.

## Affected tests

- `returns a PKNCAdata object` (line 277)
- `includes only selected analytes, dosnos, and pcspecs in intervals` (line 289)
- `sets NCA options correctly` (line 304)
- `uses default min.hl.points of 3 when not specified` (line 318)
- `sets custom min.hl.points when specified` (line 330)
- `calls update_main_intervals internally when parameter_selections is provided` (line 349)
- `applies custom_units_table when provided` (line 374)
- `applies custom_units_table correctly when units table has group columns` (line 415)
- `handles custom_units_table with extra columns not in data$units` (line 449)
- `skips update_main_intervals when neither parameter_selections nor int_parameters is set` (line 465)

## Suggested fix

Either add `parameter_selections` to the test fixtures so the intervals have params set to TRUE (making the data more realistic), or wrap the calls with `suppressWarnings` / `expect_warning` where the warnings are expected behavior for minimal fixtures.

These warnings are pre-existing on `main` and not caused by any recent PR.

js3110:
## Issue

Closes #1218

## Description

Two warnings were emitted during `PKNCA_update_data_object` calls in tests (and production) due to missing guards in `rm_impute_obs_params`:

1. **`Unknown or uninitialised column: 'impute'`** — The `impute` column on `data$intervals` was only created by `create_start_impute()` (when `start_impute = TRUE`) or the BLQ block (when `blq_imputation_rule` is not NULL). When both were skipped, `rm_impute_obs_params` and `remove_pp_not_requested` referenced a non-existent column.

2. **`No intervals found with the specified target parameters...`** — When no `parameter_selections` was provided, all NCA parameter flags remained `FALSE`. The `Reduce` loop called `interval_remove_impute` for each impute method, which found no matching rows and warned on every iteration.

**Fixes:**

- Initialize the `impute` column to `NA_character_` in `update_main_intervals` when it doesn't already exist, so downstream code can always reference it.
- Skip the `Reduce` loop in `rm_impute_obs_params` when none of the target params are `TRUE` in any interval row.

## Definition of Done

- Warnings listed in #1218 no longer appear during test runs or production use.
- Existing tests in `test-intervals_helpers.R` (which test `interval_remove_impute` directly with intentionally mismatched arguments) remain unaffected.

## How to test

Run `devtools::test(filter = "PKNCA")` and verify no warnings are emitted from the affected tests (lines 277–465).

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

Only `R/intervals.R` is changed. No test changes needed — the warnings are eliminated at the source. The `interval_remove_impute` function still warns when called directly with non-existent groups (tested in `test-intervals_helpers.R`), which is the correct behavior for that API.
">


R/intervals.R
```diff
@@ -285,6 +285,12 @@ update_main_intervals <- function(
 
   data$impute <- NA_character_
 
+  # Ensure the impute column exists on intervals so downstream code
+  # (e.g. rm_impute_obs_params, remove_pp_not_requested) can reference it.
+  if (!"impute" %in% names(data$intervals)) {
+    data$intervals$impute <- NA_character_
+  }
+
```
Gero1999: suggestion:

right after this we have `create_start_impute`. Maybe it would make more sense to force the addition of the impute column there (with empty strings probably better than NA)

Otherwise, we might also prefer including a handling missing `impute` action on the `rm_impute_obs_params` and `remove_pp_requested`, to prevent that their future use in other functions or alone can create problems. For these cases we would also need to add tests

R/intervals.R
```diff
@@ -347,6 +353,22 @@ rm_impute_obs_params <- function(data, metadata_nca_parameters = metadata_nca_pa
...
+  if (length(present_params) > 0) {
+    param_data <- data$intervals[, present_params, drop = FALSE]
+    has_active_params <- any(rowSums(
+      replace(param_data, is.na(param_data), FALSE)
+    ) > 0)
+  } else {
+    has_active_params <- FALSE
+  }
+
+  if (!has_active_params) {
```
Gero1999: suggestion:

can we refactor this to a helper function called something like `has_requested_params(PKNCAdata, params) = TRUE/FALSE` and use it in the conditional to just return the data when FALSE

Gero1999 (changes requested):
Good work! Only a couple of changes

R/intervals.R
js3110: it doesn't make sense to do it in create start impute because it only calls if impute = TRUE, so instead I hadnled it in the other functions as you suggested

R/intervals.R
```diff
@@ -347,6 +352,12 @@ rm_impute_obs_params <- function(data, metadata_nca_parameters = metadata_nca_pa
     trimws() %>%
     unique()
 
+  # If none of the target params are TRUE in any interval, there's nothing to
+  # remove imputation from.
+  if (!has_requested_params(data$intervals, params_not_to_impute)) {
+    return(data)
+  }
```
Gero1999: suggestion: maybe move it as an OR with the conditional on line 346

Gero1999 (approved):
LGTM! Non-blocking comment

Shaakon35:
Two suggestions from reviewing the diff:

**1. Unrelated NAMESPACE reordering**

The diff moves 11 `importFrom` lines into alphabetical order. This has no functional effect but inflates the diff and increases merge conflict risk with other in-flight PRs. Consider reverting this change and keeping it for a separate housekeeping commit/PR.

**2. Duplicated `impute` column initialization**

The same guard appears in both `rm_impute_obs_params` (`R/intervals.R`) and `remove_pp_not_requested` (`R/PKNCA.R`):

```r
if (!"impute" %in% names(...)) {
  ...$impute <- NA_character_
}
```

If a third caller needs this, the pattern will need to be copied again. Two options to consolidate:

- Extract a small helper, e.g. `ensure_impute_column(intervals)`, that both functions call.
- Initialize the column once upstream in `update_main_intervals`, which is the shared entry point — this was also suggested by @Gero1999 in the first review round.

Shaakon35 (approved):
I approve in advance as everything works

PR #1255, opened 2026-04-23
Add validation for invalid CONCDUR values in format_pkncaconc_data

js3110:
## Issue

Closes #1152

## Description

When `CONCDUR` (calculated as `time_end_column - time_column`, typically `AEFRLT - AFRLT`) contains negative or infinite values, PKNCA throws a vague error:

> duration must be numeric without missing (NA) or infinite values, and all values must be >= 0

Since `CONCDUR` is created internally, users have no way to understand this message. This PR adds a validation check in `format_pkncaconc_data` that stops with a clear error message telling the user how many records are invalid, how `CONCDUR` is derived, and which columns to check in their data.

The check runs before PKNCA receives the data, so the vague error never appears.

## Definition of Done

- [x] Validation check within `format_pkncaconc_data` that checks if CONCDUR is < 0 or Inf
- [x] Error given to user if yes, detailing that they should check that AFRLT and AEFRLT are correct

## How to test

1. Upload data where `AEFRLT < AFRLT` for some records — should see a clear error in the UI naming the affected columns
2. Upload data without `AEFRLT` — should work normally (no CONCDUR validation triggered)
3. Upload valid data with `AEFRLT >= AFRLT` — should work normally

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

The validation uses `stop()` rather than `warning()` because warnings are not surfaced in the Shiny UI (only errors are caught by `tryCatch` and shown as notifications). This also prevents the vague PKNCA error from ever appearing.

### DIFF
--- a/R/format_data.R
+++ b/R/format_data.R
@@ -87,7 +87,7 @@ format_pkncaconc_data <- function(ADNCA,
       (ADNCA[["CONCDUR"]] < 0 | is.infinite(ADNCA[["CONCDUR"]]))
     if (any(invalid_concdur)) {
       n_invalid <- sum(invalid_concdur)
-      warning(
+      stop(
         n_invalid, " record(s) have invalid sampling duration (CONCDUR): ",
         "values are negative or infinite. ",
         "CONCDUR is calculated as '", time_end_column, "' - '", time_column, "'. ",
--- a/R/format_data.R
+++ b/R/format_data.R
@@ -82,6 +82,20 @@ format_pkncaconc_data <- function(ADNCA,
 
   if (!is.null(time_end_column)) {
     ADNCA[["CONCDUR"]] <- ADNCA[[time_end_column]] - ADNCA[[time_column]]
+
+    invalid_concdur <- !is.na(ADNCA[["CONCDUR"]]) &
+      (ADNCA[["CONCDUR"]] < 0 | is.infinite(ADNCA[["CONCDUR"]]))
+    if (any(invalid_concdur)) {
+      n_invalid <- sum(invalid_concdur)
+      warning(
+        n_invalid, " record(s) have invalid sampling duration (CONCDUR): ",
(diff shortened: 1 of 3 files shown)


Shaakon35 (approved):
Amazing, thank you :)

R/format_data.R
```diff
@@ -82,6 +82,20 @@ format_pkncaconc_data <- function(ADNCA,
 
   if (!is.null(time_end_column)) {
     ADNCA[["CONCDUR"]] <- ADNCA[[time_end_column]] - ADNCA[[time_column]]
+
+    invalid_concdur <- !is.na(ADNCA[["CONCDUR"]]) &
+      (ADNCA[["CONCDUR"]] < 0 | is.infinite(ADNCA[["CONCDUR"]]))
```
Gero1999, line 87:
```suggestion
    invalid_concdur <- !is.finite(ADNCA[["CONCDUR"]]) || ADNCA[["CONCDUR"]] < 0
```

Gero1999 (changes requested):
question: I am a bit concerned with the fact of giving an error. `PCENDTC` will be expected to be missing when it is not an urine sample record, but here we will block this normal scenario, perhaps we need to address it as "when missing, PCENDTC = PCSTDTC"

js3110:
> question: I am a bit concerned with the fact of giving an error. `PCENDTC` will be expected to be missing when it is not an urine sample record, but here we will block this normal scenario, perhaps we need to address it as "when missing, PCENDTC = PCSTDTC"

I am not sure what you mean? CONCDUR is only calculated if AEFRLT is not null, and its only validated if its not missing...

You can test in these scenarios, it still works? @Gero1999

Gero1999 (approved):
I was actually thinking in CDISC standards but I am realizing it does not matter because we don't use dates for duration. This is perfecto!

PR #1256, opened 2026-04-23
Enhancement: Formula-style fraction layout for Ratio Calculations UI

### LINKED ISSUE #1250: Enhancement: Formula-style UI for Ratio Calculations
### Description

The current Ratio Calculations UI (`NCA > Setup > Settings > Ratio Calculations`) uses a flat table with columns `TestParameter`, `RefParameter`, `TestGroups`, `RefGroups`, etc. This layout makes it hard to see how the fields relate — the test parameter and test groups are far apart, and users may miss how they are structured together.

### Proposed changes

Replace the table-row layout with a formula-style visual structure. When the user clicks `(+) Add Row`, instead of appending a table row, a card/block is added that arranges the inputs to mirror the formula shown in the help button:

```
PPTESTCD  =  TestParameter [TestGroups]  /  RefParameter [RefGroups]  ×  AdjustingFactor
```

Concretely, each entry would display:
- **Left side:** The auto-generated PPTESTCD name with an `=` separator
- **Right side:** The parameter dropdowns and group selectors arranged inline, resembling the ratio formula

This makes it immediately clear that each block defines one ratio calculation and how the numerator/denominator map to the result.

### Definition of Done

- [ ] Each ratio entry renders as a formula-style block instead of a table row
- [ ] PPTESTCD is displayed to the left of `=`, auto-updated from inputs
- [ ] Test/Ref parameter and group inputs are arranged inline on the right, visually showing numerator/denominator
- [ ] Add/remove buttons still work as expected
- [ ] Layout is responsive and doesn't break on narrow screens
- [ ] Help button content is updated to match the new visual style

Gero1999:
## Issue

Closes #1250

## Description

Replaces the reactable-based Ratio Calculations table with a visual fraction layout. Each ratio entry now renders as a card showing the mathematical structure:

```
                    TestParam [TestGroup]
☐  PPTESTCD  =  ──────────────────────────────  ×  AdjFactor
                 Σ  RefParam [RefGroup]
```

**Key changes:**
- Numerator (test) inputs on top with green borders, denominator (ref) inputs below with orange borders, separated by a fraction line
- Aggregate Subject shown as a compact dropdown with symbols: `—` (no), `Σ` (yes), `Σ?` (if-needed)
- Hover tooltips on every input explaining its role (e.g. \"Test Parameter (numerator)\", \"Reference Group (denominator level)\")
- Responsive layout: card and fraction use `flex-wrap` so elements reflow on narrow screens instead of getting squeezed
- Row selection via checkboxes for the `(-) Remove Row/s` button
- Help panel restored to original content with LaTeX formula, field descriptions reordered to match the visual layout (left-to-right, top-to-bottom)
- Data model unchanged — same `ratio_table` reactiveVal with 7 columns, same import/export/validation logic

## Definition of Done

- [x] Each ratio entry renders as a fraction-style card instead of a table row
- [x] PPTESTCD is displayed to the left of `=`, auto-updated from inputs
- [x] Test/Ref parameter and group inputs arranged as numerator/denominator
- [x] Add/remove buttons still work as expected
- [x] Layout is responsive and doesn't break on narrow screens
- [x] Help button content reordered to match visual layout

## How to test

1. Open the app, navigate to `NCA > Setup > Settings > Ratio Calculations`
2. Click `(+) Add Row` — verify the fraction card appears with numerator/denominator layout
3. Change Test/Ref parameters and groups — verify PPTESTCD auto-updates
4. Click the Σ dropdown — verify it cycles between `—`, `Σ`, `Σ?` and the tooltip explains each
5. Hover over each input — verify tooltips appear
6. Resize the browser window narrow — verify the layout wraps gracefully
7. Select rows via checkboxes, click `(-) Remove Row/s` — verify removal works
8. Upload a settings YAML with ratio rows — verify they populate correctly
9. Export settings — verify ratio table is preserved in YAML

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

**Known limitations to address before merge:**

1. **Observer accumulation** — each `refresh_cards` re-render creates new `observeEvent` handlers that stack up. With many add/remove cycles this could cause duplicate firings. Consider refactoring to `insertUI`/`removeUI` or a single delegated observer.
2. **Checkbox reset on re-render** — row selection checkboxes reset to `FALSE` when cards re-render (e.g. after editing a parameter). The old reactable managed selection state internally.

Neither is blocking for review/demo but both should be addressed for production use.">


js3110 (approved):
really nice! super clear how it works now :)

i tested settings upload which also works well

Shaakon35:
@Gero1999 the x0.4 appears below, do you think you can have less space between the checkbox on the left, and the formula, it would be better for small screens
image

inst/shiny/modules/tab_nca/setup/ratio_calculations_table.R
```diff
@@ -158,181 +153,276 @@ ratios_table_server <- function(
...
-          "No parameters or group variables available to add a row.",
-          type = "warning"
-        )
-        return()
+    # Update table when any card input changes
+    observe({
+      tbl <- ratio_table()
+      if (nrow(tbl) == 0) return()
+      for (i in seq_len(nrow(tbl))) {
+        .observe_ratio_row(i, input, ratio_table, adnca_data, refresh_cards)
```
Shaakon35, line 200: **Suggestion — observer accumulation**

This `observe` re-runs every time `ratio_table()` changes, and each run calls `.observe_ratio_row()` which creates new `observeEvent` handlers inside `local({})`. Those handlers are never destroyed, so they stack up: after N edits, row 1 has N copies of each observer all firing at once.

This will cause duplicate `ratio_table()` mutations per single user action and could trigger cascading re-renders (observer fires → updates table → observe re-runs → more observers).

One way to fix this: track active observers in a list and call `$destroy()` on them before creating new ones:

```r
active_observers <- reactiveVal(list())

observe({
  tbl <- ratio_table()
  # Destroy previous observers
  purrr::walk(active_observers(), ~ .$destroy())

  if (nrow(tbl) == 0) {
    active_observers(list())
    return()
  }

  obs <- list()
  for (i in seq_len(nrow(tbl))) {
    obs <- c(obs, .observe_ratio_row(i, input, ratio_table, adnca_data, refresh_cards))
  }
  active_observers(obs)
})
```

This would require `.observe_ratio_row()` to return the observer handles instead of discarding them.

inst/shiny/www/main.css
```diff
@@ -213,140 +214,19 @@ h5 {
...
-
-.param-matrix-table tbody tr:nth-child(even) {
-  background: #f0f0f0;
-}
-
-.param-matrix-table tbody tr:hover {
-  background: #edf0fc;
+/* parameter selection gird */
+.nca-pill-grid {
+  display: grid;
```
Shaakon35, line 219: **Suggestion — unrelated CSS changes**

This PR adds `.nca-pill-grid` / `.nca-pill` classes that have no corresponding SCSS partial and aren't referenced by any R code in this PR. Additionally, the param-matrix styles were moved to the bottom of the file with subtle color changes (e.g. `#006aa8` → `#005b8f`, `#f0f0f0` → `#e7e7e7`, `#edf0fc` → `#f4f6fd`).

It looks like `main.css` was fully regenerated from SCSS rather than surgically edited, which pulled in changes from other work. Consider reverting the unrelated hunks and only keeping the new `.ratio-formula-card` block and its dependencies, to keep the diff scoped to the ratio formula feature.

inst/shiny/modules/tab_nca/setup/ratio_calculations_table.R
```diff
@@ -158,181 +153,276 @@ ratios_table_server <- function(
...
+      .remove_selected_ratio_rows(input, ratio_table, refresh_cards)
+    })
 
-      result <- .validate_ratio_table(
-        ratio_df, ratio_param_options(), ratio_reference_options()
-      )
+    # Render formula cards
+    refresh_cards <- reactiveVal(1)
+    output$ratio_cards <- renderUI({
+      refresh_cards()
```
Shaakon35, line 180: **Suggestion — checkbox state lost on re-render**

Every `refresh_cards()` increment rebuilds all cards via `renderUI`, which resets all `select_row_*` checkboxes to `FALSE`. If a user selects rows for removal but then edits a parameter (which also triggers `refresh_cards`), their selection is silently lost.

One option: store checkbox state in a `reactiveVal` and restore it when rebuilding cards:

```r
# In .ratio_formula_card, use the stored state:
checkboxInput(ns(paste0("select_row_", i)), label = NULL,
              value = isTRUE(selected_rows()[[i]]))
```

Alternatively, skip `refresh_cards` for changes that don't affect card structure (the `AggregateSubject` case already does this — the same pattern could apply to other fields that only update `ratio_table()` without needing a visual rebuild).

inst/shiny/www/main.css
```diff
@@ -1,3 +1,4 @@
+@charset "UTF-8";
```
Shaakon35, line 1: **Suggestion — `@charset` directive**

This line wasn't present before and is an artifact of the SCSS compiler emitting it because `_ratio_formula.scss` contains Unicode characters (Σ, ×, —). Functionally harmless, but it confirms `main.css` was fully regenerated rather than hand-edited. If the unrelated CSS changes are reverted (see other comment), this line should go with them.

inst/shiny/modules/tab_nca/setup/ratio_calculations_table.R
Gero1999: this one I was aware but is irrelevant, we can ignore

inst/shiny/www/main.css
Gero1999: this one does not matter neither, we can pass

Shaakon35:
Could you please update the "?" button it's not up-to-date.
Also add information about 
image
-, x and x? because it's hard for me to understand

inst/shiny/www/main.css
Shaakon35: Considering how the special characters are shown, I would consider this no? 
image

Shaakon35:
If you manage, that would be great to make the area in red much smaller, the fomula should start very close to the checkbox.
image

Shaakon35 (approved):
Please review my comments, and merge ;)

Gero1999:
> Could you please update the "?" button it's not up-to-date.

@Shaakon35 If you mean the help-button text I think it should be updated. Check the text or code-changes to see it. But let me know if you were thinking in something else.

> also add information about -, x and x? because it's hard for me to understand

Regarding the symbol, it just means `mean`. You can check the info now also hovering the symbol or alternatively as traditionally reading in the help button:

```
Aggregate Subject : Shown in the denominator. `yes` (x̄) aggregates reference values using the mean of all subjects, `no` (—) does not, and `if-needed` (x̄?) only when ratios cannot be performed within the same subject.
```

Regarding the space I really tried everything ,but it will just break the small screen scenario if I force it

inst/shiny/www/main.css
Gero1999: hmm maybe I am missing something on this but I would say is completely fine. We are not hand-editing the `main.css`, it is auto-generated. We don't expect making direct changes there neither

PR #1262, opened 2026-04-27
refactor: keep slope_rules as NULL until rules are added

### LINKED ISSUE #1261: Allow slope_rules (and settings) to be NULL instead of requiring an empty table
## Enhancement

Originated from [PR #1246 discussion](https://github.com/pharmaverse/aNCA/pull/1246#discussion_r3137573004).

### Current behavior

The core NCA functions (e.g. `PKNCA_update_data_object`) already accept `NULL` for `hl_adj_rules` (slope rules) and handle it correctly — the parameter defaults to `NULL` and is guarded with `if (!is.null(hl_adj_rules))`.

However, the **app and settings pipeline** always materializes an empty data frame instead of `NULL`:

- **App**: `manual_slopes` in `manual_slopes_table.R` is initialized as `NULL` but immediately replaced with a 0-row data frame as soon as `pknca_data()` fires. This is then stored in `session$userData$slope_rules()` and exported to the settings YAML.
- **Settings YAML → script**: `read_settings()` calls `.convert_list_to_df(obj$slope_rules)`, which converts the YAML list back to a data frame. If the key is missing or empty in the YAML, the behavior depends on `.convert_list_to_df` — it may not gracefully return `NULL`.
- **Generated script**: The template always does `slope_rules <- settings_list$slope_rules` and passes it as `hl_adj_rules = slope_rules`, which works but is unnecessary when no rules exist.

### Desired behavior

- If `slope_rules` is missing from the settings YAML, `read_settings()` / `.process_settings_payload()` should return `NULL` (not error or produce an unexpected value).
- The generated R script should omit `hl_adj_rules` (or pass `NULL`) when no slope rules are defined, rather than always passing an empty data frame.
- Same principle for other optional settings that the app always materializes as empty tables (e.g. `int_parameters`, `ratio_table`).

### Context

@js3110 noted in the PR discussion:
> we should really try to remove this so if slope rules don't exist the function can assume that they are empty, rather than the user having to set it themselves (same with settings)

The core functions already handle `NULL`. The improvement is in the settings/app layer — making the pipeline treat missing or empty settings as `NULL` so R script users don't need to define unused settings.

Gero1999:
## Issue

Closes #1261

## Description

`manual_slopes` was eagerly initialized as an empty data frame the moment `pknca_data()` fired, even when no slope rules existed. This meant `session$userData$slope_rules()` was always a data frame (never `NULL`), and exported settings YAML always contained an empty `slope_rules:` key.

This PR keeps `manual_slopes` as `NULL` until the user actually adds a rule (via plotly click, the add button, or a settings override with rows). The same principle is applied to `int_parameters` and `ratio_table` when reading settings from YAML — empty tables are normalized to `NULL`.

### Changes

- **`manual_slopes_table.R`**: Remove eager 0-row DF initialization. `manual_slopes` stays `NULL` until a rule is added or a non-empty settings override is applied. The remove-row observer resets to `NULL` when all rows are deleted.
- **`utils-slope_selector.R`**: `check_slope_rule_overlap()` handles `NULL` existing table (returns the new rule as a single-row DF). Returns `NULL` instead of 0-row DF when all rules are cancelled.
- **`R/readers.R`**: New `.normalize_optional_df()` helper that converts list → DF then coerces 0-row DFs to `NULL`. Applied to `slope_rules`, `int_parameters`, and `ratio_table` in `.process_settings_payload()`. `units` and `time_duplicate_keys` are left unchanged since an empty table has different semantics than no table for those.

## Definition of Done

- `manual_slopes` is `NULL` when no slope rules exist (not an empty DF)
- Adding a rule via plotly click or button works as before
- Removing all rules returns `manual_slopes` to `NULL`
- Settings upload with slope rules still works
- Settings upload without slope rules (or with empty `slope_rules:`) works
- R script pipeline handles `NULL` slope_rules (already did — `PKNCA_update_data_object` has `if (!is.null(hl_adj_rules))` guard)

## How to test

1. **No slope rules flow**: Upload a dataset → go to NCA Setup → do NOT add any slope rules → Run NCA. Verify it completes without errors. Export settings YAML and confirm `slope_rules` is absent or null.

2. **Add/remove slope rules**: Upload a dataset → go to Slope Selector → add an exclusion via plotly double-click on a point → verify it appears in the table. Add a selection by clicking two points in the same interval → verify it appears. Remove all rules using the remove button → verify the table disappears (no empty table shown).

3. **Settings upload with slope rules**: Export settings from a session that has slope rules → re-upload that settings file in a new session → verify slope rules are restored correctly.

4. **Settings upload without slope rules**: Export settings from a session with no slope rules → re-upload → verify no errors and slope rules remain empty/NULL.

5. **R script**: Run the generated R script from a session with no slope rules → verify `hl_adj_rules` is `NULL` and NCA completes.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

This is a minimal refactor — 4 files changed. The core NCA functions (`PKNCA_update_data_object`, `PKNCA_calculate_nca`) already handle `NULL` for `hl_adj_rules`. The change is entirely in the app/settings layer. No new dependencies added. The last two checklist items are N/A (no SCSS or dependency changes) — left unchecked per convention."


Gero1999:
seems to work fine and the changes are logical. I will keep exploring tomorrow with the settings file

js3110:
`check_slope_rule_overlap` now returns `NULL` instead of a 0-row data frame when all rules are removed, but the Shiny test at `inst/shiny/tests/testthat/test-utils-slope-selector.R:96` still asserts `expect_equal(nrow(...), 0)`. Since `nrow(NULL)` returns `NULL`, this will fail.

Suggested fix:

```r
# "should remove full row if full range of rule is removed"
expect_null(check_slope_rule_overlap(EXISTING_FIXTURE, NEW))
```

js3110 (approved):
LGTM! thanks

Shaakon35:
### Suggestion: simplify `new_row[0, ]` fallback in `manual_slopes_table.R` (line 108-112)

```r
old_rows <- if (!is.null(manual_slopes())) {
  filter(manual_slopes(), !is.na(TYPE))
} else {
  new_row[0, ] # empty DF with matching columns
}
```

`new_row[0, ]` is fragile — it depends on `new_row` being successfully constructed earlier in the same observer. If anything above it changes, this breaks silently.

Since `manual_slopes()` is `NULL` here (first rule being added), `old_rows` will be empty and the `rbind` just wraps `new_row`. Consider simplifying:

```r
updated_data <- if (!is.null(manual_slopes())) {
  old_rows <- filter(manual_slopes(), !is.na(TYPE))
  as.data.frame(rbind(old_rows, new_row), stringsAsFactors = FALSE)
} else {
  as.data.frame(new_row, stringsAsFactors = FALSE)
}
```

This avoids the zero-row DF trick and makes the intent clearer: when there are no existing rules, the new row *is* the table.

Shaakon35 (approved):
LGTM

PR #1264, opened 2026-04-27
fix: fall back to default groupby when summary picker is not yet rendered

### LINKED ISSUE #1240: Bug: nca_statistics file does populate groupings unless you view the Descriptive Statistics Results tab
### Description
If you run NCA and go directly to save and include NCA statistics as an output. The resulting NCA statistics does not include the groupings only the "Statistic" and "Parameter" columns.

### To reproduce
1) Run NCA with default data set
2a) Do not navigate to the Descriptive Statistics Results tab
2b) Click save with NCA statistics selected
3) Resulting NCA statistics file will only contain the "Statistic" and "Parameter" columns.

### Expected behaviour
Default grouping should be applied.

### Note
The issues appears to come from input "summary_groupby" and how the choices are dynamically updated and generate. No choices are loaded until the Descriptive Statistics Results tab is viewed.

### LINKED ISSUE #1263: Bug: descriptive statistics ungrouped when tab not yet visited
### Current behavior

The `selector_label` migration replaced the static `pickerInput` for `summary_groupby` with a dynamic `uiOutput`. Because the UI is only rendered when the user visits the Descriptive Statistics tab, `input$summary_groupby` is `NULL` until then. Any code that triggers `summary_stats` before the tab is visited (e.g. auto-replay, settings upload, or downstream reactives) passes `NULL` to `calculate_summary_stats()`, producing ungrouped statistics.

### Expected behavior

When `input$summary_groupby` is `NULL` (picker not yet rendered), the reactive should fall back to the default grouping columns (`group_cols` + `classification_cols`) — the same selection that `selector_label` would initialize the picker with.

### Steps to reproduce

1. Upload a dataset and run NCA
2. Before visiting the Descriptive Statistics tab, trigger any code path that reads `summary_stats` (e.g. export, or the `observeEvent(summary_stats(), ...)` that saves to `session$userData`)
3. The resulting statistics are ungrouped

### Fix

Add a fallback in the `summary_stats` reactive: compute `group_cols` and `classification_cols`, read `input$summary_groupby`, and if `NULL`, use `unique(c(group_cols, classification_cols))` instead.

Gero1999:
## Issue

Closes #1263

## Description

The `selector_label` migration replaced the static `pickerInput` for `summary_groupby` with a dynamic `uiOutput`. Because the UI is only rendered when the user visits the Descriptive Statistics tab, `input$summary_groupby` is `NULL` until then. This caused `calculate_summary_stats()` to receive `NULL` as the grouping argument, producing ungrouped statistics.

The fix computes `group_cols` and `classification_cols` inside the `summary_stats` reactive and falls back to `unique(c(group_cols, classification_cols))` when `input$summary_groupby` is `NULL` — matching the default selection that `selector_label` would initialize the picker with.

## Definition of Done

- Descriptive statistics are correctly grouped even before the user visits the tab
- Once the user visits the tab and changes the picker, the selected grouping is used
- No regression in normal usage flow

## How to test

1. Upload a dataset and run NCA
2. **Without visiting the Descriptive Statistics tab**, export results (or check `session$userData$results$nca_results$nca_statistics`)
3. Verify the statistics are grouped by the default columns (analyte, specimen, dose treatment, profile, etc.) — not ungrouped
4. Visit the Descriptive Statistics tab → verify the picker shows the correct default selection and the table is grouped
5. Change the groupby picker → verify the table updates accordingly

## Contributor checklist
- [ ] New logic is documented

## Notes to reviewer

Single-file, 10-line change. The fallback logic mirrors the default selection computed in the `observeEvent(res_nca(), ...)` block above that initializes the picker via `selector_label`."


Shaakon35 (commented):
It works and I could approve but:

I would like to have pre-selected automatically always:
STUDYID
PCSPEC
DOSETRT
PARAM
ATPTREF
DOSEA

and if available, select:

GROUP (pre-clin)
STRAIN (pre-clin)
TRT01A

so we avoid grouping by AGE or SEX or RACE which makes no sense

js3110 (approved):
works well for me :)

I would suggest the default not include grouping vars and instead is just group cols and dose, but maybe its a personal preference so its non-blocking

Gero1999:
@Shaakon35 I wanted to make it non-company specific, but I agree with what @js3110  proposed and it might probably be enough for us! This will now differ with what Ola did so she probably needs to adapt

Shaakon35:
Almost good, I like the current summary stats.

Just if you run the nca, (do not open the summary stats), go to save, save the summary stats.csv:
it contains the following columns but not ATPTREF:

STUDYID,"PCSPEC","DOSETRT","PARAM","Statistic","AUCLSTD[(Hours*ug/mL)/mg]","AUCLST[Hours*ug/mL]","CLSTD[(ug/mL)/mg]","CLST[ug/mL]","CMAXD[(ug/mL)/mg]","CMAX[ug/mL]","LAMZHL[Hours]","LAMZLL[Hours]","LAMZNPT[count]","LAMZSPN[fraction]","LAMZ[1/Hours]","R2ADJ[unitless]","R2[unitless]","TLST[Hours]","TMAX[Hours]","VSSO[mL]"

can you fix that @Gero1999 ?

Shaakon35:
I approve to not loose time

Shaakon35 (approved):

PR #1269, opened 2026-04-28
feat: automate data-raw rebuild with helper script and CI check

### LINKED ISSUE #1268: Automate data-raw rebuild: helper script + CI staleness check
## Problem

The `data-raw/` scripts (`compile_css.R`, `metadata_nca_parameters.R`, `metadata_nca_variables.R`, `adnca_example.R`, `sdtm_example.R`, `test-multianalyte-ADNCA.R`) must be run manually whenever their source inputs change. Forgetting to run them leads to stale outputs (CSS, `.rda` data files) that only surface as bugs later.

There is no automated way to:
1. Run all relevant scripts in one command
2. Detect stale outputs in CI before merge

### 1. `data-raw/rebuild.R` — single-command local rebuild

A script that sources all `data-raw/*.R` scripts (excluding itself and `test_suggests_hidden.R`). Developers run `Rscript data-raw/rebuild.R` before committing.

### 2. CI workflow — staleness check

A GitHub Actions workflow (similar to the existing `man-pages.yml` pattern) that:
1. Runs `Rscript data-raw/rebuild.R`
2. Checks `git diff` for any changes to tracked files
3. Fails the PR if outputs are stale

This catches forgotten rebuilds before merge without changing the development workflow.

## Out of scope

Running `data-raw/` scripts at app startup (`aNCA::run_app()`) — these are developer-time operations and should not run in production.

Gero1999:
## Issue

Closes #1268

## Description

Adds a single-command rebuild script and a CI staleness check for `data-raw/` outputs.

**`data-raw/rebuild.R`** — Sources all `data-raw/*.R` scripts (except `rebuild.R` and `test_suggests_hidden.R`) in isolated environments. After running all scripts, it writes CSV snapshots of every `.rda` dataset to `data-raw/test/`. Developers run `Rscript data-raw/rebuild.R` before committing when they've changed any source file that feeds into `data-raw/` (SCSS partials, CSV metadata, example data).

**`data-raw/test/`** — Contains CSV exports of every `data/*.rda` dataset, generated automatically by `rebuild.R`. These deterministic text files are the basis for the CI staleness check, since `.rda` files are not reproducible across R versions/platforms (identical data can produce different binary output).

**`.github/workflows/data-raw-check.yml`** — CI workflow that runs `rebuild.R` and checks for changes in tracked files, excluding `data/*.rda` (non-deterministic). The CSV snapshots in `data-raw/test/` serve as the deterministic proxy that catches real data changes. Follows the same pattern as the existing `man-pages.yml` check. Added to `main.yml` alongside the other CI jobs. Can be skipped with `[skip data-raw]` in the commit message.

## Setup after merge

Run `Rscript data-raw/rebuild.R` once to generate and commit the initial `data-raw/test/*.csv` snapshot files. The CI check depends on these files being present.

## Definition of Done

- `Rscript data-raw/rebuild.R` runs all relevant scripts without error
- CSV snapshots in `data-raw/test/` are generated and committed
- CI fails when committed outputs are stale (e.g. SCSS edited but `main.css` not recompiled, or CSV metadata changed but snapshots not regenerated)
- CI passes when outputs are up to date
- CI ignores `.rda` binary differences (non-deterministic serialization)

## How to test

1. Run `Rscript data-raw/rebuild.R` locally — verify all scripts execute, CSV snapshots are written to `data-raw/test/`, and no tracked files change (`git status` is clean after committing the snapshots)
2. Intentionally edit a source file (e.g. change a value in `metadata_nca_parameters.csv`) without running `rebuild.R` — push and verify CI fails with a message about stale outputs

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

- The CI workflow installs `usethis` and `sass` as extra packages since they're needed by the `data-raw/` scripts but aren't in `DESCRIPTION` Imports. If future `data-raw/` scripts need additional packages, they should be added to the `extra-packages` list in `data-raw-check.yml`.
- `.rda` files are excluded from the staleness diff because R's `save()`/`usethis::use_data()` serialization is not deterministic across R versions and platforms. The CSV snapshots in `data-raw/test/` are the deterministic replacement for detecting real changes.


Gero1999:
The workflow seems to be detecting more changes than what we originally expected:


image

Gero1999:
> The workflow seems to be detecting more changes than what we originally expected:
> 
> image

the problem was that different R versions can produce different `rda` byte-sizes. I just changed the strategy directly comparing the final `csv` versions

Gero1999:
Hi @vedhav , thanks for the data updates! Just checking in case you had any comments/review for the implementation

@KOBANAK  and I noticed the solution still can generate locally different .rda checksums for identical files, which could unnecessarily confuse the user and lead them to make updates/pushes. Do you see a workaround, or is this acceptable to move forward with?

vedhav:
Thanks for checking in @Gero1999. I'm still reviewing it. I'll add comments as soon as I wrap my review.

vedhav (approved):
I confirm that it works as expected. This will make sure that the rds files are not stale. I am okay with merging this solution. I just have a small comment that we can talk about before merging.

From the maintainability perspective, it is best to not store multiple csv files with the same names. The csv files are the source of truth and the rds files are just generated from them.
But, in this solution we are generating the csv files from the rds files just for checking, this check can be done in memory by reading the csv file and comparing them (i.e re-run every builder script in memory). This way, we do not store redundant test files in storage. The only caveat of this approach is that when there are mismatches, we can address this by logging the diff after we compare them. WDYT @Gero1999

Gero1999:
@vedhav you are right that it is not super-advantegous to have to keep the extra csv files. I made an alternative solution on #1311, in case you can give it a look

Gero1999:
#1311 will cover in a better way the issue

PR #1273, opened 2026-04-29
Fix ratio calculations with aggregate subject not working

### LINKED ISSUE #1271: Ratio calculation with `aggregate_subject = "yes"` never aggregates reference values
### Description

When using the ratio calculation table with `Aggregate Subject = yes` (e.g., AUCIFOD Oral / AUCIFOD IV to compute bioavailability using mean IV), the aggregation never actually occurs. The ratio result either doesn't appear in the NCA Results, or shows an incorrect value (an arbitrary individual ratio instead of the mean).

### Root cause

In `calculate_ratios.data.frame()` (`R/ratio_calculations.R`, line 166), the `group_by` includes `paste0(group_cols, "_ref")`:

```r
group_by(across(any_of(c(match_cols, group_cols, "PPTESTCD", paste0(group_cols, "_ref")))))
```

`group_cols` is derived as all columns except a small set of value columns (`PPTEST`, `PPORRES`, etc.), so it includes `PPTESTCD`, `USUBJID`, `DOSNOA`, `start_dose`, `end_dose`, and more. After the merge on line 150 (which uses `suffixes = c("", "_ref")`), the `_ref` versions of these columns exist — including `USUBJID_ref`.

Grouping by `USUBJID_ref` means each reference subject gets its own group. The subsequent `mean(PPORRES_ref)` computes the mean of a single value — no aggregation happens.

### What happens downstream

1. `n = n()` is always 1 per group, so `PPTESTCD` becomes `"RAAUCIFOD"` instead of `"RAAUCIFOD (mean)"`
2. The `select(any_of(c(names(df_test), "PPANMETH")))` on line 229 drops the `_ref` columns, producing multiple rows per test subject with different `PPORRES` values (each divided by a different individual ref value)
3. The `distinct()` in `calculate_ratio_app` (lines 410–416) keeps only the first row per `PPTESTCD + group_vars + end` — this silently picks whichever ref subject happened to be first in the merge, producing an arbitrary individual ratio instead of the mean:

```r
all_ratios %>%
    distinct(
      across(
        all_of(c("PPTESTCD", group_vars(res$data), "end"))
      ),
      .keep_all = TRUE
    )
```

4. The ratio column may appear in pivoted results with wrong values, or may not appear at all depending on how the arbitrary values interact with downstream pivoting

### Proposed fix

Remove `_ref` columns from the `group_by` so that the mean is computed across all matched reference rows:

```r
# Current (broken): groups by ref-side columns, preventing aggregation
group_by(across(any_of(c(match_cols, group_cols, "PPTESTCD", paste0(group_cols, "_ref")))))

# Fix: only group by test-side identifiers
group_by(across(any_of(c(match_cols, group_cols, "PPTESTCD"))))
```

After this change, when `USUBJID` is not in `match_cols` (aggregate=yes), all ref subjects for the same test subject fall into one group, and `mean(PPORRES_ref)` computes the actual mean across ref subjects. The existing `distinct()` in `calculate_ratio_app` already handles collapsing the resulting identical rows.

### Impact on other ratio types

- **`aggregate = "no"` (all ratio types)**: `USUBJID` is in `match_cols`, so the merge is 1:1 per subject. `mean()` of a single value is the same regardless of whether `_ref` columns are in the group_by. **No behavioral change.**
- **`aggregate = "if-needed"`**: Individual call (with USUBJID) is unchanged. Aggregated fallback call now correctly computes the mean instead of picking an arbitrary first ref value. **Fixes the fallback path only.**
- **`aggregate = "yes"`**: This is the broken case. **Fixes the bug.**
- **Existing tests**: All tests use `USUBJID` in `match_cols` (effectively `aggregate = "no"`). **No existing test changes behavior.** New tests for `aggregate = "yes"` should be added.

### Edge case: accumulation ratios with multiple DOSNOA per ATPTREF

For accumulation ratios (ATPTREF-based), `start`/`end` are removed from `match_cols`. If a subject has multiple `DOSNOA` values under the same ATPTREF (e.g., BID regimen), the merge cross-joins test profiles with all ref profiles. Currently `DOSNOA_ref` in the group_by keeps them separate and `distinct()` picks the first one (arbitrary — already incorrect). With the fix, they would be averaged instead — arguably more correct, and this scenario is rare in practice. If precise dose-number matching is needed for BID accumulation ratios in the future, `DOSNOA` should be added to `match_cols` for ATPTREF-based ratios as a separate enhancement.

### Additional considerations

- The `PPANMETH` derivation downstream (lines 198–210) uses `ref_cols` values via `c_across`, which would need to reflect that aggregation happened (e.g., noting "mean" rather than a specific ref subject's value).
- The `n = n()` count and the `"(mean)"` suffix logic on line 213 should then work correctly once the grouping is fixed.

js3110:
## Issue

Closes #1271

## Description

Three bugs prevented ratio calculations from working when `Aggregate Subject` is set to `yes` or `if-needed`:

1. **`group_by` included `_ref` columns** (`R/ratio_calculations.R`): The `group_by` in `calculate_ratios.data.frame()` included `paste0(group_cols, "_ref")`, which meant each reference subject got its own group. `mean(PPORRES_ref)` computed the mean of a single value — no aggregation ever occurred. The `distinct()` downstream then silently picked an arbitrary first reference value.

2. **`if-needed` never fell back to aggregation** (`R/ratio_calculations.R`): The `USUBJID %in% match_cols` check was performed after wrapping `match_cols` in `list()`, so `%in%` always returned `FALSE` and the aggregated fallback branch was never appended.

3. **Ratio parameters hidden from NCA Results** (`inst/shiny/modules/tab_nca/nca_results.R`): Ratio PPTESTCDs (e.g., `FABS`, `FREL`, `RAAUCIFOD (mean)`) are not in `metadata_nca_parameters`, so `match()` returned `NA` and they were excluded from the parameter selector widget, hiding the columns from the results table.

## Definition of Done

- Ratio calculations with `Aggregate Subject = yes` correctly compute the mean of reference values across subjects
- Ratio calculations with `Aggregate Subject = if-needed` fall back to aggregation when no individual match exists
- Ratio parameter columns (FABS, FREL, etc.) appear in the NCA Results table
- Existing ratio calculations with `Aggregate Subject = no` are unaffected

## How to test

1. Load a dataset with IV and EX subjects
2. In NCA > Settings > Ratio Calculations, add a ratio: TestParameter = `AUCIFOD`, RefParameter = `AUCIFOD`, RefGroups = `ROUTE: intravascular`, Aggregate Subject = `yes`
3. Run NCA
4. Verify the ratio column (e.g., `FABS`) appears in NCA Results with correct values (test AUCdn / mean of IV AUCdn values)
5. Repeat with Aggregate Subject = `if-needed` — should produce the same results when no subject has both routes

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

- The `group_by` fix removes all `_ref` columns from the grouping. For `aggregate = no`, this has no effect (1:1 join per subject). For accumulation ratios with multiple DOSNOA per ATPTREF (e.g., BID regimens), the ref values would now be averaged instead of arbitrarily picking the first — this is rare and arguably more correct than the previous behavior.
- The `if-needed` bug means this mode has effectively been equivalent to `no` since it was introduced — the aggregated fallback was dead code.
- R is not available in the dev container, so tests were not run locally. CI will validate.


tests/testthat/test-ratio_calculations.R
```diff
@@ -326,6 +326,112 @@ describe("calculate_ratios", {
...
+      PPORRESU = "ng/mL",
+      PPSTRES = c(10, 4),
+      PPSTRESU = "ng/mL",
+      GROUP = c("EX", "IV"),
+      start = 0,
+      end = 24,
+      type_interval = "main",
+      exclude = "",
+      stringsAsFactors = FALSE
+    )
```
Gero1999: issue: Add an input with at least 4 rows (2 subjects) so the test becomes meaningful for the case

tests/testthat/test-ratio_calculations.R
```diff
@@ -326,6 +326,112 @@ describe("calculate_ratios", {
...
+
+    iv_mean <- mean(
+      res_if_needed$result$PPORRES[res_if_needed$result$ROUTE == "intravascular"],
+      na.rm = TRUE
+    )
+    ex_values <- res_if_needed$result$PPORRES[
+      res_if_needed$result$ROUTE == "extravascular"
+    ]
+
+    ratios <- calculate_ratio_app(
```
Gero1999, line 396: This is the only test added using `calculate_ratio_app`. I think it might make sense to make all the added tests with this function, as is the final one we want to test. Check for `if-needed` also for the case when it is not needed and/or there is a mix of cases

Gero1999 (changes requested):
I still don't manage to make the ratios aggregation work... Whenever you don't have same subject for the comparison it seems to always shows the message that the parameter could not be computed. I tried with `RACE`, `SEX`, `DOSNOA`, `ROUTE (Drip/Bolus)`. 

Image

Shaakon35:
While testing:
image


```
[2026-04-29 16:22:30] ERROR: Error calculating NCA results:
ℹ In argument: `across(all_of(c("PPTESTCD",
  group_vars(res$data), "end")))`.
Caused by error in `across()`:
ℹ In argument: `all_of(c("PPTESTCD",
  group_vars(res$data), "end"))`.
Caused by error in `all_of()`:
! Can't subset elements that don't exist.
✖ Elements `PPTESTCD`, `STUDYID`, `PCSPEC`, `DOSETRT`, `USUBJID`, etc. don't exist.
```

Shaakon35:
### Note on `group_by` change in `calculate_ratios.data.frame` (line 166)

Removing `paste0(group_cols, "_ref")` from the `group_by` changes how the mean denominator is computed for **all** ratio calculations, not just the aggregate case.

**Before:** the mean was computed per reference group level (grouped by `match_cols + group_cols + PPTESTCD + group_cols_ref`).
**After:** the mean is computed across all reference rows within the same `match_cols + group_cols + PPTESTCD` combination.

This is correct for the aggregate case (where reference values should be averaged across subjects), but it's a broader change — it also affects `aggregate_subject = "no"` and `aggregate_subject = "if-needed"` with individual matching. The existing non-aggregate tests pass, which is reassuring, but it would be worth confirming this doesn't change results for the non-aggregate path. If the `_ref` columns are always constant within a reference group (which they should be by construction), then the change is safe. Worth calling out explicitly in the PR description either way.

js3110:
> I still don't manage to make the ratios aggregation work... Whenever you don't have same subject for the comparison it seems to always shows the message that the parameter could not be computed. I tried with `RACE`, `SEX`, `DOSNOA`, `ROUTE (Drip/Bolus)`.
> 
> Image

This is due to the current grouping, which uses start + end in the grouping. So for our data, since start and end are never matching across subjects, it doesn't aggregate them. This isn't a bug but rather a conscious choice, so I think if we change it it should be done in a separate issue.

Some options could be grouping by ATPTREF instead, or allowing the user to choose if they group by Dose or not (eg sometimes they want to calculate ratios across doses (not acc ratio))

Shaakon35 (approved):

Gero1999 (approved):
yep ok I will open another issue to make sure the ratios are good for everything and make it priority as well. On the rest of the aspects this one LGTM

js3110:
Ran:
```
devtools::lint()
devtools::test()
spelling::spell_check_package()
devtools::document()
```

all found no errors

Gero1999:
related issue here: https://github.com/pharmaverse/aNCA/issues/1286

PR #1295, opened 2026-05-05
fix: preserve multi-dose context in param_selection_data

### LINKED ISSUE #1294: Bug: Parameter selection shows single dose instead of multiple dose when TRTRINT is NA
### Description

When a dataset has multiple doses but `TRTRINT` is `NA`, selecting a single profile (ATPTREF) in NCA Settings causes the parameter selection to incorrectly classify the study as single dose instead of multiple dose.

The root cause is in `param_selection_data` (`inst/shiny/modules/tab_nca/nca_setup.R`). It filters `conc$data` by `ATPTREF`, which removes rows from other dosing occasions. `detect_study_types()` then checks `unique(DOSNOA) == 1` on the filtered data — if the selected profile corresponds to dose 1, the only `DOSNOA` left is `1`, so it concludes single dose.

When `TRTRINT` is present and non-NA (as in the example dataset), `detect_study_types` bypasses the `DOSNOA` check entirely (`missing_tau = FALSE` → `is_one_dose = FALSE`), which is why the bug doesn't reproduce with the built-in dummy data.

### To reproduce

1. Load a multiple-dose dataset where `TRTRINT` is `NA`
2. In NCA Settings, select a single profile (e.g., the first dosing occasion)
3. Observe the parameter selection matrix — it shows "Single Extravascular Dose" (or similar) instead of "Multiple Extravascular Doses"

### Expected behaviour

Parameter selection should show "Multiple ... Doses" regardless of which profile is selected, since the study design is multi-dose.

### Proposed fix

In `param_selection_data`, replace each row's `DOSNOA` with the max `DOSNOA` per dose group before filtering by `ATPTREF`. This preserves the multi-dose signal while still filtering data to the selected profile:

```r
param_selection_data <- reactive({
  req(adnca_data(), settings_output$analyte(), settings_output$pcspec())
  log_trace("Updating parameter selection data.")

  pknca_data <- adnca_data()

  # Preserve max DOSNOA per dose group before filtering by profile,
  # so detect_study_types still identifies multi-dose studies correctly
  # even when TRTRINT is NA.
  dose_groups <- group_vars(pknca_data$dose)
  pknca_data$conc$data <- pknca_data$conc$data %>%
    dplyr::group_by(dplyr::across(dplyr::all_of(dose_groups))) %>%
    dplyr::mutate(DOSNOA = max(DOSNOA, na.rm = TRUE)) %>%
    dplyr::ungroup() %>%
    dplyr::filter(
      PARAM %in% settings_output$analyte(),
      PCSPEC %in% settings_output$pcspec(),
      ATPTREF %in% settings_output$profile()
    )
  pknca_data
})
```

This is safe because `parameter_selection_server` never uses per-row `DOSNOA` — it only passes `conc$data` to `detect_study_types`, which checks `unique(DOSNOA) == 1`.

js3110:
## Issue

Closes #1294

## Description

When `TRTRINT` is `NA`, `detect_study_types` falls back to checking `unique(DOSNOA) == 1` to determine single vs multiple dose. The `param_selection_data` reactive filters `conc$data` by `ATPTREF` (profile), which strips out rows from other dosing occasions. If the selected profile corresponds to dose 1, the only `DOSNOA` left is `1`, so the study is misclassified as single dose.

This fix computes `max(DOSNOA)` per dose group before the `ATPTREF` filter, so the multi-dose signal is preserved in the filtered data.

The bug doesn't reproduce with the built-in example dataset because it has `TRTRINT = 24` (non-NA), which bypasses the `DOSNOA` check entirely.

## Definition of Done

- Parameter selection correctly shows "Multiple ... Doses" when a single profile is selected in a multi-dose study with `TRTRINT = NA`

## How to test

1. Load a multiple-dose dataset where `TRTRINT` is `NA`
2. In NCA Settings, select a single profile (e.g., the first dosing occasion)
3. Verify the parameter selection matrix shows "Multiple ... Doses" instead of "Single ... Dose"

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

The change is scoped to `param_selection_data` in `nca_setup.R`. `parameter_selection_server` never uses per-row `DOSNOA` — it only passes `conc$data` to `detect_study_types`, which checks `unique(DOSNOA) == 1`. Overwriting `DOSNOA` with the group max is safe in this context.

NEWS update and version bump still needed.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.1.0.9166
+Version: 0.1.0.9167
 Authors@R: c(
     person("Ercan", "Suekuer", email = "ercan.suekuer@roche.com", role = "aut",
            comment = c(ORCID = "0009-0001-1626-1526")),
--- a/inst/shiny/modules/tab_nca/nca_setup.R
+++ b/inst/shiny/modules/tab_nca/nca_setup.R
@@ -71,14 +71,22 @@ nca_setup_server <- function(id, data, adnca_data, extra_group_vars, settings_ov
     )
 
     # Lightweight filtered data for parameter selection — only depends on
-    # analyte/pcspec, not on method, slopes, flags, or exclusions.
+    # analyte/pcspec/profile, not on method, slopes, flags, or exclusions.
     # Filters conc$data directly instead of building a full PKNCA data object.
     param_selection_data <- reactive({
       req(adnca_data(), settings_output$analyte(), settings_output$pcspec())
       log_trace("Updating parameter selection data.")
 
       pknca_data <- adnca_data()
+
+      # Preserve max DOSNOA per dose group before filtering by profile,
+      # so detect_study_types still identifies multi-dose studies correctly
+      # even when TRTRINT is NA.
+      dose_groups <- group_vars(pknca_data$dose)
       pknca_data$conc$data <- pknca_data$conc$data %>%
+        dplyr::group_by(dplyr::across(dplyr::all_of(dose_groups))) %>%
+        dplyr::mutate(DOSNOA = max(DOSNOA, na.rm = TRUE)) %>%
+        dplyr::ungroup() %>%
         dplyr::filter(
           PARAM %in% settings_output$analyte(),
           PCSPEC %in% settings_output$pcspec(),


daviesmeg (approved):
Looks good to me, the NCA was calculated with this fix instead of giving an error.

Gero1999 (approved):
LGTM! I reviewed changing the default dataset, and also checked that the other test-datasets work. Thanks!

PR #1308, opened 2026-05-12
fix: translate PPTESTCDs to PKNCA names in excretion interval flags

### LINKED ISSUE #1307: Bug: Excretion analysis crashes when submitting — PPTESTCD/PKNCA name mismatch in interval flags
### Description

The excretion analysis tab crashes with `object 'type_interval' not found` when selecting excretion parameters (e.g., RCAMINT, FREXINT) and pressing Submit.

### Root cause

`excretion_analysis.R` uses `input$param_select` directly to set interval flags:

```r
mutate(across(any_of(input$param_select), ~ TRUE, .names = "{.col}"))
```

`input$param_select` contains **PPTESTCDs** (`"RCAMINT"`, `"FREXINT"`), but the interval columns use **PKNCA names** (`"ae"`, `"fe"`). `any_of()` silently finds no matches, so no parameter flags are set to TRUE. PKNCA runs with no requested parameters, returns an empty result, and the downstream `add_label_attribute()` crashes because `type_interval` is missing from the empty result.

### Steps to reproduce

1. Load a dataset with excretion/urine data (must have WTBL column)
2. Go to Additional Analysis → Excretion tab
3. Check "Adjust for Body Weight"
4. Select RCAMINT and FREXINT as parameters
5. Select samples and profiles
6. Press Submit

### Error

```
Error in mutate: ℹ In argument: `PPTESTCD_unit = case_when(...)`.
Caused by error in `case_when()`:
! Failed to evaluate the left-hand side of formula 1.
Caused by error:
! object 'type_interval' not found
```

### Fix

Translate `input$param_select` from PPTESTCDs to PKNCA names before using them as interval column selectors:

```r
pknca_params <- translate_terms(input$param_select, "PPTESTCD", "PKNCA")
```

Then use `pknca_params` instead of `input$param_select` in both `across(any_of(...))` calls (profile intervals at line 160 and sample intervals at line 189).

This affects both the profile and sample interval construction in `inst/shiny/modules/tab_nca/excretion_analysis.R`.

Gero1999:
## Issue

Closes #1307

## Description

The excretion analysis crashes when submitting because `input$param_select` contains PPTESTCDs (e.g., `RCAMINT`, `FREXINT`) but the interval columns use PKNCA names (`ae`, `fe`). The `across(any_of(input$param_select))` silently matches nothing, so no parameter flags are set to TRUE. PKNCA returns an empty result, and `add_label_attribute()` crashes on the missing `type_interval` column.

The fix translates `input$param_select` from PPTESTCDs to PKNCA names via `translate_terms()` before using them as interval column selectors. **Both** the profile intervals (line 160) and sample intervals (line 188) need this translation.

## How to test

1. Load a dataset with excretion/urine data (with WTBL column)
2. Go to Additional Analysis → Excretion tab
3. Check \"Adjust for Body Weight\"
4. Select RCAMINT and FREXINT as parameters
5. Select both samples and profiles as interval types
6. Press Submit
7. Verify results table renders without error and shows computed values for the selected parameters

## Contributor checklist
- [x] New logic is documented
- ~App or package changes are reflected in NEWS~ `The bug appeared during development of v0.2`

## Notes to reviewer

~The current PR only fixes the sample intervals path. The profile intervals path (line 160) has the same bug and also needs `translate_terms()` applied — see issue #1307 for details.">~ Applied to both


Belubbit (approved):
Works as specified

daviesmeg (approved):
Look good to me!

PR #1311, opened 2026-05-15
feat: automate data-raw rebuild with helper script and CI staleness check

### LINKED ISSUE #1268: Automate data-raw rebuild: helper script + CI staleness check

Gero1999:
## Issue

Closes #1268

## Description

Adds a single-command rebuild script and a CI staleness check for `data-raw/` outputs.

**`data-raw/rebuild.R`** — Sources all `data-raw/*.R` scripts (except `rebuild.R` and `test_suggests_hidden.R`) in isolated environments. Before sourcing, it loads all committed `.rda` datasets into memory. After sourcing (which overwrites the `.rda` files), it reloads them and compares with `identical()`. Any mismatch prints `MISMATCH: <dataset>` to stderr.

**`.github/workflows/data-raw-check.yml`** — CI workflow with two checks:
1. Runs `rebuild.R` and fails if any `MISMATCH:` lines appear in the output (detects stale `.rda` data via in-memory comparison, avoiding the `.rda` binary non-determinism problem).
2. Checks non-`.rda` outputs (CSS, test CSVs) via `git diff`, since these are deterministic text files.

Added to `main.yml` alongside the other CI jobs. Can be skipped with `[skip data-raw]` in the commit message.

### Why not just `git diff` everything?

R's `save()` / `usethis::use_data()` serialization is not deterministic across R versions and platforms — identical data can produce different binary `.rda` output. This causes false-positive CI failures. The in-memory comparison with `identical()` compares the actual R objects, not the binary representation.

## Definition of Done

- `Rscript data-raw/rebuild.R` runs all relevant scripts without error
- CI fails when committed `.rda` data is stale (rebuild.R prints MISMATCH)
- CI fails when non-`.rda` outputs are stale (e.g., SCSS edited but `main.css` not recompiled)
- CI passes when all outputs are up to date
- CI ignores `.rda` binary differences (non-deterministic serialization)

## How to test

1. Run `Rscript data-raw/rebuild.R` locally — verify all scripts execute and `>> All datasets are up to date.` is printed
2. Intentionally edit a source file (e.g., change a value in `data-raw/metadata_nca_parameters.csv`) without regenerating the `.rda` — run `Rscript data-raw/rebuild.R` and verify it prints `MISMATCH: metadata_nca_parameters`

## Contributor checklist
- [x] New logic is documented
- ~Package version is incremented~
- ~Settings upload works with the new implementation (if applicable)~

## Notes to reviewer

- This is a derivative branch from PR #1269 to test the in-memory comparison approach before merging into the main PR branch.
- The CI workflow installs `usethis` and `sass` as extra packages since they're needed by the `data-raw/` scripts but aren't in `DESCRIPTION` Imports. If future `data-raw/` scripts need additional packages, they should be added to the `extra-packages` list in `data-raw-check.yml`."
- `multianalyte-ADNCA.csv` is changed because in `main` at some point it was missed to be updated


.github/workflows/data-raw-check.yml
```diff
@@ -0,0 +1,91 @@
...
+          r-version: ${{ inputs.r-version }}
+          use-public-rspm: true
+
+      - name: Install R Dependencies
+        uses: r-lib/actions/setup-r-dependencies@v2
+        with:
+          extra-packages: |
+            local::.
+            any::usethis
+            any::sass
```
vedhav: I realized that `usethis` is not added as suggests in the package dependency but used in the package.
I also don't think we need sass for this workflow so `any::sass` can be removed.

.github/workflows/data-raw-check.yml
```diff
@@ -0,0 +1,91 @@
...
+          # and fail if any mismatches are found.
+          Rscript data-raw/rebuild.R 2>&1 | tee rebuild_output.log
+          if grep -q "^MISMATCH:" rebuild_output.log; then
+            echo ""
+            echo "::error::data-raw outputs are stale. Run 'Rscript data-raw/rebuild.R' and commit the results."
+            grep "^MISMATCH:" rebuild_output.log
+            rm -f rebuild_output.log
+            exit 1
+          fi
+          rm -f rebuild_output.log
```
vedhav: This logic works. My only worry is that we can shoot ourselves in the future when someone attempts to improve the rebuild log and removes the `MISMATCH` word from it.
So, I like when we keep all related things together. In this case I'd recommend erroring out right from R itself with a simple if statement that checks if we are in a GitHub action, `GITHUB_ACTIONS` env variable is `true` when GitHub CI runs the rebuild script, and when we spot mismatches, we can simply `stop()` with a message.
[Other packages use this strategy too](https://github.com/tidyverse/googledrive/blob/8de11bfd4349ae6e7d98b51f7a6ebc148fe988fa/R/utils-io.R#L47).

vedhav (commented):
Thanks for quickly raising an alternative PR! I added a couple of comments.

.github/workflows/data-raw-check.yml
Gero1999: this is indeed a much more appropiate way :)

Gero1999:
thansk for the suggestions @vedhav, let me know if the changes satisfy them, I will for now close #1269

vedhav (approved):
Works great! Good job!

KOBANAK (commented):
Tested locally the script correctly detects mismatches when source files are modified. It works as expected.

KOBANAK (approved):

PR #1316, opened 2026-05-22
Remove tern dependency, replace with inline ggplot2 (TLGs)

### LINKED ISSUE #1319: Enhancement: Remove tern dependency from TLG plots
### Description

`tern` is in `Imports` but is only used in 3 call sites in `R/g_pkcg.R`:

- `pkcg01`: `tern::g_ipp(plotting_choices = \"separate_by_obs\")` — produces a template ggplot that is immediately stripped of its color aesthetic. The splitting logic is unused since aNCA does its own grouping via `id_plot`.
- `pkcg02`: `tern::g_ipp(plotting_choices = \"all_in_one\")` — same pattern, only the base ggplot structure is kept.
- `pkcg03`: `tern::g_lineplot()` + `tern::control_lineplot_vars()` — used for mean concentration-time profiles with summary statistics (mean±SD, mean±SE, mean±CI, median±CI).

In all three cases, aNCA overrides most of what tern sets up (theme, scales, aesthetics, labels). Replacing these with direct ggplot2 code removes a heavy transitive dependency tree (`tern` → `nestcolor`, `formatters`, `rtables`, etc.) and gives full control over the plot construction.

### Definition of Done

- [ ] `tern::g_ipp()` replaced with inline `ggplot() + geom_line() + geom_point()` in `pkcg01` and `pkcg02`
- [ ] `tern::g_lineplot()` replaced with inline summary stat computation + ggplot2 in `pkcg03`
- [ ] `tern` removed from DESCRIPTION Imports and NAMESPACE
- [ ] vdiffr snapshots regenerated
- [ ] All existing tests pass
- [ ] Plots render correctly in the app (LIN/LOG/SBS scales, all summary methods)

Gero1999:
## Issue

Closes #1319

## Description

Removes the `tern` package from Imports by replacing all three call sites with equivalent inline ggplot2 code:

- **`pkcg01`** (individual plots): `tern::g_ipp(plotting_choices = \"separate_by_obs\")` was called only to produce a template ggplot from the first group, then `[[1]]` was taken and the color aesthetic was immediately stripped. Replaced with a direct `ggplot() + geom_line() + geom_point()`.

- **`pkcg02`** (combined plots): `tern::g_ipp(plotting_choices = \"all_in_one\")` replaced with equivalent ggplot2 code keeping `group = USUBJID` and `colour = USUBJID` for the overlay.

- **`pkcg03`** (mean plots): `tern::g_lineplot()` + `tern::control_lineplot_vars()` replaced with inline summary stat computation (mean, sd, se, median, CI via `qt()`, median CI via `wilcox.test()`) and a ggplot with `geom_point` + `geom_line` + `geom_errorbar`. N-count labels (`\"Group (N = X)\"`) are appended to match tern's format.

Also fixed a pre-existing roxygen bug: `pkcg03` had `@importFrom tern g_ipp` but actually used `tern::g_lineplot`.

## Definition of Done

- [x] All `tern::` calls removed from `R/g_pkcg.R`
- [x] `tern` removed from DESCRIPTION Imports
- [x] `importFrom(tern, g_ipp)` removed from NAMESPACE
- [x] New `@importFrom` entries added for `stats::qt`, `ggplot2::geom_line`, `ggplot2::geom_errorbar`
- [x] vdiffr snapshots regenerated
- [x] Tests pass

## How to test

1. Run `devtools::test(filter = \"g_pkcg\")` — all existing tests should pass after snapshot regeneration.
2. In the app: submit TLG order with pkcg01 (individual), pkcg02 (combined), pkcg03 (mean) graphs in LIN/LOG/SBS scales. Verify plots render correctly and the customizing widgets work perfectly
3. For pkcg03: test all summary methods (Mean±SD, Mean±SE, Mean±CI, Median±CI) and whisker options (Both/Upper/Lower).

## Contributor checklist
- [x] New logic is documented
- [x] Settings upload works with the new implementation — N/A, no settings changes
- [x] If any `.scss` change was done, run `data-raw/compile_css.R` — N/A, no SCSS changes
- [x] If a package dependency was added/changed, run `data-raw/test_suggests_hidden.R` — dependency was removed (tern from Imports, nestcolor from Suggests), which only reduces surface area

## Notes to reviewer

- The changes in the code are not that big. Most changes come from the fact that SVG test images have been renewed. All 28 vdiffr snapshot SVGs were deleted. The first test run will create new baselines. Plots will look slightly different since tern applied `nestcolor::theme_nest()` by default — `pkcg03` now uses `theme_bw()` explicitly, and `pkcg01`/`pkcg02` use no base theme (matching the rest of the customization that was already applied on top).

- The median CI computation uses `wilcox.test(conf.int = TRUE)` wrapped in `tryCatch`, returning `NA` for groups with < 2 observations — matching tern's `s_summary` behavior.


R/g_pkcg.R
```diff
@@ -849,28 +853,42 @@ pkcg03 <- function(
...
+        shape = .data[[strata_n]],
+        group = .data[[strata_n]]
+      )
+    ) +
+      geom_point(size = 2, na.rm = TRUE) +
+      geom_line(na.rm = TRUE) +
+      geom_errorbar(
+        aes(
+          ymin = .data[[whiskers_value[1]]],
+          ymax = .data[[whiskers_value[max(1, length(whiskers_value))]]]
```
Shaakon35, line 879: **Bug — one-sided whiskers collapse to zero height**

When `whiskers_lwr_upr` is `"Upper"` or `"Lower"`, `whiskers_value` is trimmed to a single element upstream (lines 767-769). Here, `max(1, length(whiskers_value))` evaluates to `1`, so both `ymin` and `ymax` map to the same column — the error bar becomes invisible.

In tern's `g_lineplot`, the missing whisker defaults to the mid value, producing a visible one-sided bar (e.g., from mean up to upper CI bound).

Suggested fix:

```r
geom_errorbar(
  aes(
    ymin = if (length(whiskers_value) >= 2) {
      .data[[whiskers_value[1]]]
    } else if (grepl("upr$", whiskers_value[1])) {
      .data[[mid_value]]
    } else {
      .data[[whiskers_value[1]]]
    },
    ymax = if (length(whiskers_value) >= 2) {
      .data[[whiskers_value[2]]]
    } else if (grepl("lwr$", whiskers_value[1])) {
      .data[[mid_value]]
    } else {
      .data[[whiskers_value[1]]]
    }
  ),
  width = 0.45
)
```

Or simpler — normalize `whiskers_value` to always be length 2 before this block:

```r
if (length(whiskers_value) == 1) {
  if (grepl("lwr$", whiskers_value)) {
    whiskers_value <- c(whiskers_value, mid_value)
  } else {
    whiskers_value <- c(mid_value, whiskers_value)
  }
}
```

R/g_pkcg.R
```diff
@@ -1056,6 +1074,60 @@ generate_subtitle_mean <- function(plot_data, subtitle, plotgroup_vars, plotgrou
...
 
+# Compute summary statistics per group and timepoint for pkcg03
+compute_summary_stats <- function(plot_data, xvar, yvar, mean_group_var,
+                                  interval_value, mid_value) {
+  df_stats <- plot_data %>%
+    group_by(.data[[mean_group_var]], .data[[xvar]]) %>%
+    summarise(
+      n = dplyr::n(),
+      mean = mean(.data[[yvar]], na.rm = TRUE),
+      sd = sd(.data[[yvar]], na.rm = TRUE),
```
Shaakon35, line 1112: **Minor — redundant computation in `summarise()`**

`mean()`, `sd()`, and `sqrt(n())` are each computed multiple times. `dplyr::summarise` allows referencing columns defined earlier in the same call:

```r
summarise(
  n = n(),
  mean = mean(.data[[yvar]], na.rm = TRUE),
  sd = sd(.data[[yvar]], na.rm = TRUE),
  se = sd / sqrt(n),
  median = median(.data[[yvar]], na.rm = TRUE),
  mean_ci_lwr = mean - qt(0.975, n - 1) * se,
  mean_ci_upr = mean + qt(0.975, n - 1) * se,
  mean_sdi_lwr = mean - sd,
  mean_sdi_upr = mean + sd,
  mean_sei_lwr = mean - se,
  mean_sei_upr = mean + se,
  .groups = "drop"
)
```

Cleaner and avoids recomputing the same values 6 times per group.

R/g_pkcg.R
```diff
@@ -1056,6 +1074,60 @@ generate_subtitle_mean <- function(plot_data, subtitle, plotgroup_vars, plotgrou
...
+      mean_sei_upr = mean(.data[[yvar]], na.rm = TRUE) +
+        sd(.data[[yvar]], na.rm = TRUE) / sqrt(dplyr::n()),
+      .groups = "drop"
+    )
+
+  if (interval_value == "median_ci") {
+    median_ci <- plot_data %>%
+      group_by(.data[[mean_group_var]], .data[[xvar]]) %>%
+      summarise(
+        median_ci_lwr = tryCatch(
```
Shaakon35, line 1128: **Minor — `wilcox.test()` called twice per group**

`wilcox.test(.data[[yvar]], conf.int = TRUE)` runs the full test twice — once to extract `$conf.int[1]` and again for `$conf.int[2]`. Since this involves permutation-based computation, it's worth calling once.

One approach — compute outside `summarise` and join:

```r
median_ci <- plot_data %>%
  group_by(.data[[mean_group_var]], .data[[xvar]]) %>%
  summarise(
    ci = list(tryCatch(
      stats::wilcox.test(.data[[yvar]], conf.int = TRUE)$conf.int,
      error = function(e) c(NA_real_, NA_real_)
    )),
    .groups = "drop"
  ) %>%
  mutate(
    median_ci_lwr = vapply(ci, `[`, numeric(1), 1),
    median_ci_upr = vapply(ci, `[`, numeric(1), 2)
  ) %>%
  select(-ci)
```

Shaakon35:
Hello @Gero1999 looks good to me, thank you!

Two points to take into account in the future of the TLGs:
- pre-process TODO: Keep only time points for which at least 50% of the values are not BLQ at a time point for each group
- As we don't use nestcolor, we will need to introduce nice color theme and try to keep the same color for the same patient across plots.

_________________________________________________________________________________
code 50% BLQ:
```
    # Keep only time points for which at least 50% of the values are not BLQ at a time point for each group
  	keep_timepoint <- adpc %>% group_by(
        group, .data[[timevar]]) %>%
        summarise(
          # Count number of samples for each timepoint by group
          n_samples =  n_distinct(USUBJID),
          # Count number of BLQ samples for each timepoint by group
                  n_blq = (sum(grepl("BLQ|LTR|<[1-9]|<PCLLOQ", AVALC))),
          # Compute BLQ ratio for each timepoint by group
                  n_blq_ratio = n_blq/n_samples,
                  .groups = "drop"
                  ) %>%
    filter((n_blq_ratio <= 0.5 & n_samples > 1) | .data[[timevar]]==0) %>%
    select(group, .data[[timevar]])
	
	# Filter adpc to keep only timepoints of interest (where group and timevar match entries in keep_timepoint)
  	adpc <- adpc %>% inner_join(keep_timepoint, by = c("group", timevar)) 
```

Shaakon35 (approved):

Shaakon35:
Should we remove the "print(tlg_order())" in tab_tlg.R?

R/g_pkcg.R
Gero1999: good point, not sure if in main this was working but I will fix

Gero1999:
> Hello @Gero1999 looks good to me, thank you!
> 
> Two points to take into account in the future of the TLGs:
> 
> * pre-process TODO: Keep only time points for which at least 50% of the values are not BLQ at a time point for each group

This better to do it apart in a different branch, perhaps adding a checkbox or widget when submitting TLGs


> * As we don't use nestcolor, we will need to introduce nice color theme and try to keep the same color for the same patient across plots.

We can do as in `Exploration`, where we allow the user to choose among a theme of colours that do not need external packages

R/g_pkcg.R
```diff
@@ -849,28 +848,42 @@ pkcg03 <- function(
...
+      )
+    ) +
+      geom_point(size = 2, na.rm = TRUE) +
+      geom_line(na.rm = TRUE) +
+      geom_errorbar(
+        aes(
+          ymin = .data[[whiskers_value[1]]],
+          ymax = .data[[whiskers_value[max(1, length(whiskers_value))]]]
+        ),
+        width = 0.45
```
KOBANAK: geom_point and geom_line both pass na.rm = TRUE, but geom_errorbar does not. To avoid warnings in case we have NA values.
```suggestion
        width = 0.45,
        na.rm = TRUE
```

KOBANAK (approved):
Hi, I think the ggplot2 code is much cleaner, and the unit tests passed on my end.

I don't know if this should be taken care of in another issue but while doing manual  testing in the app, I noticed :

- When a custom title is entered in the UI for combined plots, it is not being displayed on the plot. (see screenshot below)

- Across the other plots, using the "!" + column name syntax no longer retrieves the label value as it did previously. Is this behavior expected?

Image

R/g_pkcg.R
Gero1999: good point, agreed! :)

Gero1999:
> * When a custom title is entered in the UI for combined plots, it is not being displayed on the plot. (see screenshot below)
> * Across the other plots, using the "!" + column name syntax no longer retrieves the label value as it did previously. Is this behavior expected?

The "!" is indeed not working. And `pkcg02 - Combined Linear` is true that does not make the change unless you change back and forth to another figure, which is strange. For the others for me is working fast as long as you click off the input widget once you introduce the text

However both things also happen in `main`, so I will move forward and open new issues for these

AngelaManginelli (approved):
Hi, 
I have collected a few comments when I was playing around with the app. I hope those are useful, even if not for this issue. 
Best regards
Angela.

[TLG_Listing_Graph.docx](https://github.com/user-attachments/files/28346342/TLG_Listing_Graph.docx)

Shaakon35:
> Hi, I have collected a few comments when I was playing around with the app. I hope those are useful, even if not for this issue. Best regards Angela.
> 
> [TLG_Listing_Graph.docx](https://github.com/user-attachments/files/28346342/TLG_Listing_Graph.docx)

Hello @AngelaManginelli I cannot access the docx. Can you share it via another route please?

PR #1333, opened 2026-05-27
Enhancement/UI nitpicks

### LINKED ISSUE #1253: Enhancement: UI nitpicks
### Description
The purpose of this issue is to make a list with all the very small things within the App (hovering messages, button labels, dropdowns) that could be changed to improve understanding and readability. 

### Definition of Done
- [ ] I have seen the button "Save" and in fact I wondered if this would also save the setting, but if you go with the mouse on the button, the help says "Export all selected results" and you only see the option if you click on it. Suggest to add "and setting"
- [ ] Rename in `Units table` the title names as mentioned here: https://github.com/pharmaverse/aNCA/pull/1239#pullrequestreview-4162450271

h5hoang:
## Issue

Closes #1253

## Description

A collection of small UI consistency and readability improvements across the app:

- Renamed "Download locally the NCA Data" to "Download NCA Data" for simpler phrasing
- Standardized "colour" to "color" in PK/Dose QC Plot sidebar label
- Standardized all "Choose the..." dropdown labels to "Select the..." for consistency
- Standardized help button (?) placement to right-aligned across all NCA Setup sections (General Exclusions, Parameter Exclusions, Slope Selection)
- Renamed "Save" button to "Export as ZIP" with a descriptive hover tooltip
- Renamed PCSPEC column to "Specimen" and added "Short Parameter" column in the Units table

## Definition of Done

- [x] Rename "Download locally the NCA Data" to "Download NCA Data"
- [x] Standardize "colour" vs "color" mismatch in Exploration tab sidebars
- [x] Standardize "Choose the..." vs "Select the..." dropdown labels
- [x] Fix Help button (?) in Slope Selection taking its own row
- [x] Standardize Help button (?) placement across NCA Setup sections
- [x] Rename Save button and update tooltip for clarity
- [x] Rename PCSPEC to "Specimen" and add "Short Parameter" in Units table

## How to test

1. **NCA > NCA Results**: Verify the download button says "Download NCA Data"
2. **Exploration > PK/Dose QC Plot**: Verify sidebar says "color" not "colour"
3. **All dropdowns**: Verify labels use "Select the..." consistently (check NCA Setup settings, Exploration sidebars, Parameter Plots)
4. **NCA > Setup sections** (General Exclusions, Parameter Exclusions, Slope Selection): Verify help button (?) is right-aligned
5. **Save/Export button**: Verify it says "Export as ZIP" and hover tooltip says "Choose items to include in the export"
6. **NCA > Setup > Parameter Selection > Parameter Units modal**: Verify PCSPEC column shows as "Specimen" and a new "Short Parameter" column is present

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

All changes are UI-only (labels, tooltips, layout). No logic or behavior changes


inst/shiny/modules/tab_nca/units_table.R
```diff
@@ -104,6 +106,8 @@ units_table_server <- function(id, mydata) {
       editable = c("PPSTRESU", "conversion_factor"),
       columns = list(
         PPTESTCD = colDef(name = "Parameter"),
+        PPTESTCD_short = colDef(name = "Short Parameter"),
+        PCSPEC = colDef(name = "Specimen"),
```
Gero1999: issue: this one I think is potentially risky. The thing is that PCSPEC only shows off in the units table if there are two criteria satisfied: 1) It was mapped by the user, 2) It is relevant to stratify the units

Perhaps we can instead use instead `get_label` on any variable that is remaining without a custom colDef or something similar, and that way standard names will be populated for whatever extra column is here (PCSPEC or PARAM or whatever)

inst/shiny/modules/tab_nca/zip.R
```diff
@@ -27,7 +27,7 @@ zip_ui <- function(id) {
         "font-weight: 500;",
         "font-size: 1rem;"
       ),
-      title = "Export all selected results as a ZIP archive",
+      title = "Choose items to include in the export",
```
Gero1999: suggestion/nitpick: I am thinking that perhaps the user thinks that before pressing the button they need to click/select the items they want

```suggestion
      title = "Press to select the items to include in the export",
```

Gero1999 (changes requested):
A small issue identified in the code, but the UI looks much more consistent, thanks!

On the other hand, this is more a nitpick I identified, in case you can find an easy-solution:

https://github.com/user-attachments/assets/8a2c52a6-6d93-4904-ad45-645c9ec69c5d

inst/shiny/modules/tab_nca/units_table.R
h5hoang: Ah i see, i was under the assumption that PCSPEC would always be present as a grouping column in the units table, so the easiest way seemed to be just hardcoding the label. thanks for the feedback, I'll go ahead and try out your suggestion! 🫡

h5hoang:
@Gero1999 

Thanks for the feedback again!!

I've made some changes that reflect the comments you had starting with:

1. Updating the tooltip hover message: 
image

2. Realigning the (?) button to be in the same position across the NCA tabs 

https://github.com/user-attachments/assets/b74951e7-6992-4ece-bde2-f5be29c57c77

3. using `get_label` rather than hardcoding PCSPEC, so for any grouping column that appears in the unit table, it will get its label automatically from the metadata 
image

inst/shiny/modules/tab_nca/setup/manual_slopes_table.R
```diff
@@ -5,20 +5,21 @@
 #'
 #' @param id Shiny module id
 #' @return Shiny UI element (fluidRow)
```
Gero1999: nitpick: I am in general not a big fan of using an arg for fixed/constant value, I would rather put this internally. However, I understand why you did it here, thinking that the `help_widget` logically speaking belongs to the `slope_selector.R` rather than `manual_slopes_table.R`.

I think I can pass it, just if possible add the corresponding docs in the module header. This is not exported but provides good context


```suggestion
#' @param id Shiny module id
#' @param help_widget Dropdown help button content
#' @return Shiny UI element (fluidRow)
```

Gero1999 (approved):
LGTM! I just added a non-blocking comment, good job 💪🏽

inst/shiny/modules/tab_nca/units_table.R
```diff
@@ -104,12 +106,16 @@ units_table_server <- function(id, mydata) {
...
+        PPTESTCD_short = colDef(name = "Short Parameter"),
         PPORRESU = colDef(name = "Default Unit"),
         PPSTRESU = colDef(name = "Custom Unit"),
         conversion_factor = colDef(name = "Conversion Factor"),
         is_hidden = colDef(show = FALSE),
         default = colDef(show = FALSE)
       ),
+      defaultColDef = colDef(
+        header = function(value) get_label(value)
+      ),
```
Shaakon35: **Issue 1: `defaultColDef` header function overrides explicit `name` values**

In reactable, when both `name` and `header` are set, `header` takes precedence for rendering. Since `defaultColDef` sets `header = function(value) get_label(value)`, this function will be called for **all** columns — including those with explicit `name` properties like `PPTESTCD = colDef(name = "Parameter")`.

This means the explicit `name` values on lines 108–112 are dead code. The actual display depends entirely on what `get_label()` returns for each column name. If `get_label("PPTESTCD")` doesn't return `"Parameter"`, the column header will be wrong.

**Proposed fix:** Remove `defaultColDef` and add `header` only to columns that need dynamic resolution (like PCSPEC). Keep the explicit `name` values for columns where you already know the label:

```r
columns = list(
  PPTESTCD = colDef(name = "Parameter"),
  PPTESTCD_short = colDef(name = "Short Parameter"),
  PCSPEC = colDef(header = function(value) get_label(value)),
  PPORRESU = colDef(name = "Default Unit"),
  PPSTRESU = colDef(name = "Custom Unit"),
  conversion_factor = colDef(name = "Conversion Factor"),
  is_hidden = colDef(show = FALSE),
  default = colDef(show = FALSE)
),
# no defaultColDef needed
```

This way each column explicitly controls its own header, and there's no risk of `get_label()` returning unexpected values.

inst/shiny/modules/tab_nca/units_table.R
```diff
@@ -104,12 +106,16 @@ units_table_server <- function(id, mydata) {
       editable = c("PPSTRESU", "conversion_factor"),
       columns = list(
         PPTESTCD = colDef(name = "Parameter"),
+        PPTESTCD_short = colDef(name = "Short Parameter"),
```
Shaakon35: **Issue 2: `get_label("PPTESTCD_short")` likely won't resolve to "Short Parameter"**

Due to Issue 1, the `defaultColDef`'s `header` function will call `get_label("PPTESTCD_short")` for this column. `PPTESTCD_short` is a synthetic column name created in this PR — it won't exist in `metadata_nca_variables` or any other label lookup table. `get_label()` will likely fall back to returning the raw string `"PPTESTCD_short"`, which is not the intended display name.

**Proposed fix:** If you keep `defaultColDef`, override `header` explicitly for this column:

```r
PPTESTCD_short = colDef(
  name = "Short Parameter",
  header = function(value) "Short Parameter"
)
```

Or better, apply the fix from Issue 1 (remove `defaultColDef`) and the existing `name = "Short Parameter"` will work as intended.

inst/shiny/modules/tab_nca/setup/slope_selector.R
```diff
@@ -29,99 +29,97 @@ slope_selector_ui <- function(id) {
...
-                img(src = "images/slope_plot_check.gif", alt = "Check")
-              ),
-              div(
-                class = "gif-container",
-                tags$h1("Zoom"),
-                tags$h6("Click and drag to select and zoom in a specific area.",
-                  " Double click to zoom out."
-                ),
-                img(src = "images/slope_plot_zoom.gif", alt = "Zoom")
+  tagList(
```
Shaakon35: **Issue 3: `manual_slopes_table_ui` and help dropdown are now outside `.slope-selector-module`**

The old code wrapped everything in `div(class = "slope-selector-module", ...)`. The new code uses `tagList(...)` at the top level, and `.slope-selector-module` only wraps the plot widgets section (line 84).

This breaks several CSS rules in `_slope_selector.scss` / `main.css` that target children of `.slope-selector-module`:

- `.slope-selector-module .sw-dropdown` — help dropdown float:right (no longer applies to the help widget since it's outside the wrapper)
- `.slope-selector-module .sw-dropdown-content` — dropdown width
- `.slope-selector-module .reactable` — table padding/margin (the manual slopes reactable is now outside)

**Proposed fix:** Wrap the entire `tagList` content in the `.slope-selector-module` div to preserve CSS scope:

```r
div(
  class = "slope-selector-module",
  manual_slopes_table_ui(
    ns("manual_slopes"),
    help_widget = dropdown(...)
  ),
  div(
    # ... plot widgets (currently at line 84+)
  )
)
```

This keeps the help button right-alignment fix while preserving the existing CSS rules for the dropdown content and reactable table.

inst/shiny/modules/tab_nca/setup/manual_slopes_table.R
```diff
@@ -5,20 +5,21 @@
...
-  fluidRow(
-    # Selection and exclusion controls #
-    div(
-      class = "plot-widget-group",
-      actionButton(ns("add_rule"), "+ Exclusion/Selection", class = "btn-success")
-    ),
-    div(
-      class = "plot-widget-group",
-      actionButton(ns("remove_rule"), "- Remove selected rows", class = "btn-warning")
+  tagList(
```
Shaakon35, line 12: **Issue 4: Return type changed from `fluidRow` to `tagList`**

This function previously returned a single `fluidRow(...)`. It now returns `tagList(fluidRow(...), fluidRow(...))` — two sibling `fluidRow` elements.

This changes the DOM structure for callers. The old code produced one `<div class="row">` containing both the buttons and the table. The new code produces two sibling `<div class="row">` elements. If any caller or CSS rule expected a single container element, the layout may shift.

Currently only `slope_selector.R` calls this function, so the blast radius is small. But the `@return` docstring on line 7 still says `"Shiny UI element (fluidRow)"` — it should be updated to `tagList`.

**Proposed fix:** Update the docstring and verify the layout visually:

```r
#' @return Shiny UI element (tagList of fluidRows)
```

Also consider whether the two `fluidRow` elements should share a common wrapper div for consistent spacing, especially since the `.slope-selector-module .reactable` CSS rule (padding/margin) may no longer apply (see Issue 3).

Shaakon35:
Apart from that, all good, I am ready to approve once you make the changes

h5hoang:
thank you both for the feedback, I appreciate it a lot! i've pushed fixes for all the comments which are:
- removed `defaultColDef` and switched to a `column` function that keeps explicit names for known columns and only applies `get_label` to dynamic grouping columns
- wrapped everything back inside the `.slope-selector-module` div
- updated the docstrng for `manual_slopes_table_ui` with `@param help_widget` and fixed `@return`
- adjusted the padding in SCSS/CSS so the `(?)` button lines up consistently across all 3 tabs in the NCA set up (this is more so there's visual alignment of the button, same placement across the tabs)

let me know if this works and if any other changes need to be made or if any other nitpicks arises 🙏

Shaakon35 (approved):
All good!

Shaakon35:
### Blocker: `reactable()` does not accept a function for `columns`

In `inst/shiny/modules/tab_nca/units_table.R`, the `columns` parameter was changed from a named list to a function:

```r
columns = function(data) {
  fixed_cols <- list(...)
  dynamic_cols <- setdiff(names(data), names(fixed_cols))
  # ...
  c(fixed_cols, group_cols)
},
```

`reactable::reactable()` requires `columns` to be a named list — passing a function causes a runtime error:

> `columns` must be a named list of column definitions

Verified locally with `reactable 0.4.4`.

**Proposed fix:** compute the column definitions before the `reactable()` call using a reactive or inline expression:

```r
output$units_table <- renderReactable({
  data <- ...
  
  fixed_cols <- list(
    PPTESTCD = colDef(name = "Parameter"),
    PPTESTCD_short = colDef(name = "Short Parameter"),
    PPORRESU = colDef(name = "Default Unit"),
    PPSTRESU = colDef(name = "Custom Unit"),
    conversion_factor = colDef(name = "Conversion Factor"),
    is_hidden = colDef(show = FALSE),
    default = colDef(show = FALSE)
  )
  dynamic_cols <- setdiff(names(data), names(fixed_cols))
  group_cols <- lapply(dynamic_cols, function(col) {
    colDef(header = function(value) get_label(value))
  })
  names(group_cols) <- dynamic_cols
  all_cols <- c(fixed_cols, group_cols)
  
  reactable(
    data,
    columns = all_cols,
    ...
  )
})
```

Shaakon35 (changes requested):
see above issue

h5hoang:
@Shaakon35 thanks for the feedback! you're right that `reactable::reactable()` doesn't accept a function for `columns`, but the code here is going through the `reactable_server` wrapper (`inst/shiny/modules/common/reactable.R`) rather than calling `reactable()` directly

from my understanding, the wrapper was built to handle functions and reactives for any argument, where before it calls `reactable()`, it evaluates everything:

```r
opts <- lapply(reactable_opts, function(x) {
  if (is.reactive(x))  { x() }
  else if (is.function(x)) { x(labeled_data()) }
  else { x }
})
```

so by the time `reactable()` is actually called, the `columns` should already be  a resolved named list of `colDef` objects which is similar to how the default `columns = define_cols` works everywhere else in the app.

also the proposed fix would break things since `reactable_server` returns the `unit_edits` reactive that the edit observer relies on and swapping to a bare `renderReactable` would lose the editable columns functionality entirely

Shaakon35 (approved):

Gero1999:
another PR merged @h5hoang 💪🏽 

I will just do quick merge this time to incorporate the changes

PR #1341, opened 2026-06-02
fix: slope rules table clearing and data corruption on row removal (#1302)

### LINKED ISSUE #1302: To test: slope rules table not clearing when all rules are removed
## Context

PR #1262 refactors `manual_slopes` to stay `NULL` until the user adds a rule. This introduces a potential UI bug: when the user removes all slope rules, the reactable table may not clear visually.

## What to test

1. Upload a dataset
2. Go to **NCA Setup → Slope Selector**
3. Add a rule (double-click a point twice, or use the **"+ Exclusion/Selection"** button)
4. Verify the row appears in the table
5. Select the row and click **"- Remove selected rows"**
6. **Check:** does the table clear, or does the deleted row remain visible?

## Root cause (if confirmed)

Two guards prevent the table from updating when `manual_slopes` becomes `NULL`:

**`slope_selector.R`** — the `updateReactable` observer:
```r
observeEvent(manual_slopes(), {
  req(manual_slopes())  # ← silently aborts when NULL
  reactable::updateReactable(outputId = "manual_slopes", data = manual_slopes())
})
```

**`manual_slopes_table.R`** — the `renderReactable` output:
```r
output$manual_slopes <- renderReactable({
  req(manual_slopes())  # ← silently aborts when NULL
  ...
})
```

In both cases, `req()` blocks execution when `manual_slopes()` is `NULL`, so the table UI keeps showing the last row even though the data was cleared.

## Suggested fix

Handle the `NULL` → clear transition explicitly, e.g. with `ignoreNULL = FALSE` on the observer and clearing/hiding the table output when `manual_slopes()` is `NULL`.

## Related

- PR #1262

wangzhengdna-lang:
## Issue

Closes #1302

## Description

When all slope rules were removed via the **"- Remove selected rows"** button, the reactable table failed to clear visually. Successive row removals further corrupted the displayed data: TYPE changed from "Exclusion" to "Selection", RANGE was cleared, and a ghost column (row.names leaked as data) appeared on the left.

**Reported by @Gero1999** — identified that `req()` guards silently abort on NULL, preventing table clearing after PR #1262 refactored `manual_slopes` to NULL initialization.

## Root Cause (4 layers)

| Layer | Cause | Fix |
|-------|-------|-----|
| **1. Table not clearing** | `validate(need(FALSE))` retains previous output; `req()` silently aborts | `return(NULL)` in renderReactable to explicitly clear |
| **2. Data corruption (TYPE→Selection, RANGE cleared, ghost column)** | `reactable.extras` widgets (`dropdown_extra`, `text_extra`) send Shiny input events during post-render initialization, overwriting `manual_slopes()` with widget defaults | Generation counter + 500ms suppression window in edit observer |
| **3. Stray columns (ATPTREF from plot clicks)** | `rbind`/`bind_rows` merge data frames with different column sets | Column normalization before all combines (add handler, `check_slope_rule_overlap`, render) |
| **4. Broken `updateReactable` (silent no-op since #1262)** | `outputId` in slope_selector.R missing one `-manual_slopes` namespace suffix; fixing it caused double-trigger races with `renderReactable` | Reverted to no-op; `renderReactable` + `bindEvent(refresh_reactable())` already handles all updates |

## Definition of Done

- [x] Slope rules table clears when all rows are removed
- [x] Row removal does not corrupt remaining rows (TYPE, RANGE, REASON unchanged)
- [x] No extra/ghost columns appear after successive removals
- [x] Adding rows after clearing works correctly
- [x] Results > Manual Adjustments shows "No rules defined" when rules are empty
- [x] Column normalization prevents stray columns from plot-click paths
- [x] Edit events are suppressed during widget re-initialization

## How to test

**Requires manual testing (Shiny UI, cannot be covered by unit tests):**

1. **Table clearing**: Add 3 rules → remove all one by one → table should clear after last removal
2. **Data integrity**: Add rules with different TYPE/RANGE → remove middle row → remaining rows should keep original values
3. **Plot-click + button mix**: Click plot to add exclusion → use "+ Exclusion/Selection" button → remove rows → no extra columns appear
4. **Settings upload**: Export settings with slope rules → re-import → verify rules restore correctly
5. **Results tab**: Remove all rules → check Results > Manual Adjustments shows "No rules defined"

## Files changed

- `inst/shiny/modules/tab_nca/setup/manual_slopes_table.R` — Core fix: `return(NULL)`, column normalization, edit suppression guard
- `inst/shiny/modules/tab_nca/setup/slope_selector.R` — Reverted broken `updateReactable` to no-op
- `inst/shiny/functions/utils-slope_selector.R` — Column normalization in `check_slope_rule_overlap`
- `inst/shiny/modules/tab_nca.R` — Results tab clearing observer
- `tests/testthat/test-utils-slope_selector.R` — Updated NULL→0-row DF assertions + column normalization tests
- `inst/shiny/tests/testthat/test-utils-slope-selector.R` — Updated NULL→0-row DF assertions + column normalization tests

## Contributor checklist
- [x] Code passes lintr checks (only 2 pre-existing cyclocomp warnings, unrelated to this PR)
- [x] Code passes all unit tests (1693 pass, 0 fail, 0 warn, 3 skip)
- [x] New logic covered by unit tests — `check_slope_rule_overlap` column normalization has 4 new tests
- [x] New logic is documented — CLAUDE.md updated with patterns; inline comments explain suppression guard
- [x] Package version is incremented (0.1.0.9175 → 0.1.0.9176)
- [ ] R script works with the new implementation — N/A (no NCA logic changes)
- [ ] Settings upload works with the new implementation — requires manual verification (see How to test)

## Notes to reviewer

- The edit suppression guard (generation counter + 500ms delay) is a novel pattern in this codebase. It addresses a `reactable.extras` behavior where widgets send Shiny input events during post-render initialization, which would otherwise overwrite valid data with widget defaults.
- The `updateReactable` in `slope_selector.R` was intentionally reverted to a no-op rather than fixed. Fixing the namespace would cause double-trigger races with `renderReactable`. The `renderReactable` + `bindEvent(refresh_reactable())` pattern already handles all table updates.
- The `tab_nca.R` observer targets the Results tab's `reactable_server("manual_slopes")` (output ID `<ns>-manual_slopes-table`), NOT the Setup tab's `manual_slopes_table_server` (output ID `<ns>-manual_slopes-manual_slopes`). These are different Shiny modules.


Shaakon35:
**`updateReactable` in `tab_nca.R` targets a non-existent element ID**

The new code in `tab_nca.R` calls:

```r
reactable::updateReactable(ns("manual_slopes-table"), data = ...)
```

This constructs the ID `"nca-manual_slopes-table"`, but the child module's reactable output is `output$manual_slopes` (registered via `reactableOutput(ns("manual_slopes"))`), which produces the actual ID `"nca-manual_slopes-manual_slopes"`.

So `"manual_slopes-table"` ≠ `"manual_slopes-manual_slopes"` — the `updateReactable` call targets a non-existent element and silently does nothing.

Beyond the wrong ID, this pattern breaks module encapsulation. The parent is reaching into the child module's namespace by guessing its internal output name. If the child renames its output, this breaks silently.

**Suggested fix:** Remove the `updateReactable` call entirely. The child module already handles empty/NULL data in its `renderReactable`:

```r
data <- manual_slopes()
if (is.null(data) || nrow(data) == 0) {
  return(NULL)
}
```

When `slope_rules()` becomes NULL/empty, the existing reactive chain (`slope_rules → manual_slopes → renderReactable`) already clears the table. The parent just needs to ensure `slope_rules()` is set correctly, which it already does.

wangzhengdna-lang (commented):
@Shaakon35 Thanks for the careful review. However, the namespace analysis here is mixing up two different Shiny modules:

### The two modules are different

| | Results tab reactable | Setup tab reactable |
|---|---|---|
| **Module** | `reactable_server("manual_slopes", ...)` | `manual_slopes_table_server("manual_slopes", ...)` |
| **Output ID pattern** | `output\$table` → `<ns>-manual_slopes-table` | `output\$manual_slopes` → `<ns>-manual_slopes-manual_slopes` |
| **Location** | `tab_nca.R:118` | `slope_selector.R:245` |

The `updateReactable` in `tab_nca.R` targets `ns("manual_slopes-table")` which resolves to `tab_nca-manual_slopes-table` — this **matches** the `reactable_server` module's output ID (`<ns>-manual_slopes-table`). The namespace IS correct.

### Why the observer is still needed

The `reactable_server` module uses `req(data())` in its `renderReactable` (line 73-78 of `reactable.R`). When `slope_rules()` becomes NULL, `req()` silently aborts — **retaining the previous table content** instead of clearing it. This is the exact same pattern we fixed in `manual_slopes_table.R` (changed `validate(need(FALSE))` → `return(NULL)` there).

Since we can't change `reactable_server` (it's shared by all tables in the app), the observer is the targeted fix: when slope rules are cleared, it explicitly updates this one table to show "No rules defined".

inst/shiny/modules/tab_nca/setup/manual_slopes_table.R
```diff
@@ -82,6 +82,23 @@ manual_slopes_table_server <- function(
...
 
+    # Flag to suppress edit events during widget (re-)initialization after table re-render.
+    # When TRUE, edit observer ignores incoming events to prevent reactable.extras widgets
+    # from writing default values (e.g. first dropdown choice, empty text) into the data.
+    # Uses a generation counter so rapid successive operations don't prematurely unblock.
+    suppress_edit_events <- reactiveVal(0)  # 0 = not suppressing, >0 = suppressing
+
+    # Set suppression on each re-render; clear after delay only if no newer re-render occurred.
+    observeEvent(refresh_reactable(), {
+      gen <- suppress_edit_events(0) + 1  # increment generation (0→1, 1→2, etc.)
```
Shaakon35: **[Warning] Generation counter never increments — self-reset bug.**

```r
gen <- suppress_edit_events(0) + 1  # increment generation (0→1, 1→2, etc.)
```

`suppress_edit_events(0)` is a **setter** call — it writes `0` into the reactiveVal and returns `0`. So `gen` is **always `1`**, never `2, 3, …` as the comment claims. The generation scheme is defeated:

- Two re-renders within 500ms both compute `gen == 1`.
- The first `delay()` to fire sees `identical(suppress_edit_events(), 1)` → resets suppression to `0`, **prematurely unblocking edits while the second widget init is still in flight** — exactly the corruption this guard is meant to prevent.

**Proposed resolution** — read the current value, don't set it:

```r
observeEvent(refresh_reactable(), {
  gen <- suppress_edit_events() + 1   # read current, then increment
  suppress_edit_events(gen)
  shinyjs::delay(EDIT_SUPPRESS_MS, {
    # only clear if no newer re-render bumped the generation
    if (identical(suppress_edit_events(), gen)) {
      suppress_edit_events(0)
    }
  })
}, ignoreInit = TRUE)
```

Worth a `testServer` assertion that two rapid `refresh_reactable()` bumps leave suppression active until the *last* delay fires (ties into the missing-tests comment).

inst/shiny/modules/tab_nca/setup/manual_slopes_table.R
```diff
@@ -82,6 +82,23 @@ manual_slopes_table_server <- function(
...
+      suppress_edit_events(gen)
+      shinyjs::delay(500, {
```
Shaakon35: **[Warning] Document the 500ms assumption and extract it as a named constant.**

The 500ms window works well in practice — no change to the value needed. But as written it's a bare magic number, and the reason it's safe (widget re-init completes within this window) isn't captured at the call site.

**Proposed resolution** — name the constant and document the assumption:

```r
# Delay before re-enabling edit events after a table re-render.
# reactable.extras widgets fire spurious edit events (default dropdown/empty
# text) while they re-initialize; 500ms reliably covers that init window so
# those events are suppressed rather than written back into the data.
EDIT_SUPPRESS_MS <- 500

# ...
shinyjs::delay(EDIT_SUPPRESS_MS, { ... })
```

Keeps the behavior identical while making the timing intent self-explanatory for future readers.

inst/shiny/modules/tab_nca/setup/manual_slopes_table.R
```diff
@@ -82,6 +82,23 @@ manual_slopes_table_server <- function(
     # create a reactive to update the reactable UI when the table changes
     refresh_reactable <- reactiveVal(0)
 
+    # Flag to suppress edit events during widget (re-)initialization after table re-render.
+    # When TRUE, edit observer ignores incoming events to prevent reactable.extras widgets
+    # from writing default values (e.g. first dropdown choice, empty text) into the data.
+    # Uses a generation counter so rapid successive operations don't prematurely unblock.
+    suppress_edit_events <- reactiveVal(0)  # 0 = not suppressing, >0 = suppressing
```
Shaakon35, line 97: **[Warning] Core fix has no automated tests — repeat-regression area.**

The four behaviors this PR fixes — table clears on empty, no data corruption on row removal, no ghost columns, Results-tab clearing — have no automated coverage; only `check_slope_rule_overlap` is tested. This code path already regressed once (#1262 → #1302), so a guard is warranted. Per `shiny.instructions.md` the UI can't be auto-tested by agents, but `testServer` covers the reactive logic.

**Proposed resolution** — add `testServer` tests for `manual_slopes_table_server`:

```r
describe("manual_slopes_table_server", {
  it("clears the table when the last rule is removed", {
    testServer(manual_slopes_table_server, args = list(/* … */), {
      # start with one rule, remove it
      session$setInputs(remove_rule = 1)
      expect_true(is.null(manual_slopes()) || nrow(manual_slopes()) == 0)
    })
  })

  it("preserves remaining rows' TYPE/RANGE/REASON on removal", {
    testServer(manual_slopes_table_server, args = list(/* … */), {
      before <- manual_slopes()
      session$setInputs(remove_rule = 1)
      expect_equal(manual_slopes()[, c("TYPE", "RANGE", "REASON")],
                   before[-1, c("TYPE", "RANGE", "REASON")])
    })
  })

  it("ignores edit events while suppression is active", {
    testServer(manual_slopes_table_server, args = list(/* … */), {
      refresh_reactable(refresh_reactable() + 1)  # triggers suppression
      expect_gt(suppress_edit_events(), 0)
      # simulate a widget-init edit -> data must be unchanged
      before <- manual_slopes()
      session$setInputs(`edit_TYPE` = list(row = 1, value = "spurious"))
      expect_equal(manual_slopes(), before)
    })
  })
})
```

The third test also guards the generation-counter fix from the separate comment.

wangzhengdna-lang:
@Shaakon35 Thanks for the detailed review. The branch has been updated with your feedback.

What was done:

1. **Fixed the generation counter self-reset bug** (`22a9320`).
   - `gen <- suppress_edit_events(0) + 1` now reads the current value first: `gen <- suppress_edit_events() + 1`.
   - The generation now properly increments (0→1→2...) so rapid successive re-renders don't prematurely unblock edit events.

2. **Extracted the 500ms delay into a named constant** (`22a9320`).
   - `EDIT_SUPPRESS_MS <- 500` with a comment documenting the assumption that `reactable.extras` widget re-initialization completes within this window.

3. **Added testServer tests for `manual_slopes_table_server`** (`a15ce2e`).
   - `inst/shiny/tests/testthat/test-manual-slopes-table.R` covers:
     - Table clears when the last rule is removed.
     - Remaining rows' `TYPE`/`RANGE`/`REASON` are preserved on removal.
     - Edit events are ignored while suppression is active (guards the generation-counter fix).
   - Merged `origin/main` and bumped version to `0.1.0.9185` (`9f5dbf0`).
   - Fixed pre-existing indentation lint and added `nolint` for pre-existing cyclomatic complexity.

Validation:
- `inst/shiny/tests/testthat/test-manual-slopes-table.R`: **3 PASS / 0 FAIL / 0 WARN**
- `devtools::test()`: **1983 PASS / 0 FAIL / 0 WARN / 3 SKIP**

wangzhengdna-lang:
Update: branch merged with `origin/main` (version bumped to 0.1.0.9189) and both CI failures are fixed.

**Lint.** The indentation "fix" bundled into `a15ce2e` went in the wrong direction — it changed function-signature args in `R/PKNCA.R`, `R/exploration_plots.R` and `R/ratio_calculations.R` from 2-space to 4-space, while lintr requires 2 spaces (same mistake as flagged on #1338). Since this branch makes no functional changes to those files, I restored them to the `origin/main` versions (`3595cae`). `lintr::lint_package()`: 0 lints.

**Spellcheck.** The #1302 NEWS entry uses "namespaced" and "reactable", which were missing from `inst/WORDLIST`. Added both.

**Verification.**
- Full test suite: 2009 PASS / 0 FAIL
- `inst/shiny/tests/testthat/test-manual-slopes-table.R` (this PR's testServer tests): 3 PASS / 0 FAIL
- The column-normalization tests added in `43e9d04` pass as well

**One heads-up (pre-existing, not from this PR):** `test-utils-slope-selector.R` has 8 failures on `origin/main` itself — the range-merging expectations from `638801a` (e.g. "should remove overlapping points" expecting merged ranges) no longer match `check_slope_rule_overlap()`'s actual behavior, and a few cases reference `FIXTURE_PKNCA_DATA` which isn't available in that test context. These tests are not executed by CI, so the rot went unnoticed. I left them untouched to keep this PR focused — happy to open a separate issue for them if you agree.

PR #1355, opened 2026-06-13
Forward-compatibility with the upcoming PKNCA release

billdenney:
## Summary
The development version of PKNCA (0.12.1.9000) breaks aNCA in three places. This makes
aNCA forward-compatible while preserving current behavior on released PKNCA (0.12.1).

## 1. Half-life inclusion no longer errors
PKNCA dev rejects an interval that has both `include_half.life` and `exclude_half.life`
"in use", and treats a column that is not entirely `NA` (e.g. all-`FALSE`) as in use.
aNCA initialized `exclude_half.life` to `FALSE`, so selecting points to *include* aborted
`pk.nca()` with "Cannot both include and exclude half-life points for the same interval".

Fix: initialize `exclude_half.life` to `NA` (matching `include_half.life`, already left
`NA` until a point is selected). Reads are guarded with `%in% TRUE`, so `NA` and `FALSE`
are treated identically ("not excluded") — no behavior change under released PKNCA.

## 2. New PKNCA parameters not yet mapped to CDISC (tests skipped on CRAN)
PKNCA dev emits parameters aNCA does not yet map: `lambda.z.corrxy` (new
`pk.calc.half.life()` output) appears unlabeled in `pivot_wider_pknca_results()`, and new
`vz.*` params such as `vz.last` create extra Vz rows in `export_cdisc()`. The two affected
tests are marked `skip_on_cran()` so an upcoming PKNCA release cannot block aNCA on CRAN,
while the mismatch still surfaces off-CRAN until the mapping is extended.
Follow-up: add the new params to `data-raw/metadata_nca_parameters.csv` and `export_cdisc()`.

## Verification
- `R CMD check` against PKNCA dev: OK (0 errors / 0 warnings / 0 notes)
- Full test suite passes (the two parameter tests skip on CRAN); lintr clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -73,6 +73,7 @@
 ## Bug fixes
 
 ### NCA Calculations
+* `exclude_half.life` is now initialized to `NA` instead of `FALSE` (matching `include_half.life`), so manually selecting half-life points to include no longer errors with "Cannot both include and exclude half-life points for the same interval" under the development version of PKNCA
 * Renal clearance (RENALCL) removed from direct PK calculations (inaccurate in PKNCA) — use ratio table instead (#781)
 * Multidose parameters (MRTMDO, MRTMDP, VSSMDO, VSSMDP, TAT) removed from direct calculations (#869)
 * Last dose interval end time extends to last observed sample instead of being cut off at tau (#1235)
--- a/R/PKNCA.R
+++ b/R/PKNCA.R
@@ -169,7 +169,11 @@ PKNCA_create_data_object <- function( # nolint: object_name_linter
   df_conc$is.excluded.hl <- FALSE
   df_conc$is.included.hl <- FALSE
   df_conc$REASON <- ""
-  df_conc$exclude_half.life <- FALSE
+  # NA (not FALSE) marks "no half-life exclusion yet". This mirrors how
+  # include_half.life is left NA until a point is selected and prevents PKNCA's
+  # "cannot both include and exclude half-life points" check from firing when
+  # only inclusions are set (an all-FALSE column counts as "in use").
+  df_conc$exclude_half.life <- NA
 
   # Create PKNCA conc object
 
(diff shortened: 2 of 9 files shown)


billdenney:
I would like to release a new version of PKNCA soon (within the next ~2 weeks) and I know that an ANCA CRAN failure will hold up the PKNCA release, so can you please address these issues? I am separately working through https://github.com/humanpred/pknca/pull/536 which would move the parameter assignment within PKNCA and then prevent one of these errors in the future.

Shaakon35:
@Gero1999 it looks good for me but I would need your review to merge this

Gero1999:
@Shaakon35 for testing properly please make sure to install the GitHub PKNCA version.  There are some problems with the HL plots rendering correctly due to the new errors introduced by PKNCA. They don't allow simultaneous selection+exclusion. 

@billdenney I gave you write access so in the future you can create repo-branches. I will create a new one and rebase yours so we can keep working in there, and afterwards open a new PR

Shaakon35:
Two items from review:

**1. `skip_on_cran()` is too broad (medium)**

The two `skip_on_cran()` additions skip tests unconditionally on CRAN regardless of which PKNCA version is installed. If a future released PKNCA ships the breaking changes before aNCA adds the parameter mappings, these tests will silently pass on CRAN. A version-conditional skip would be more precise, e.g.:

```r
skip_if(packageVersion("PKNCA") >= "0.12.2", "New PKNCA params not yet mapped")
```

**2. Test fixture not updated (medium)**

`tests/testthat/setup.R` line 148 still initializes `exclude_half.life = FALSE` in `FIXTURE_PKNCA_DATA`. Production code now initializes to `NA`. Tests pass because `%in% TRUE` treats both identically, but the fixture doesn't reflect the actual production state. Future tests that assume the fixture matches production could have subtle bugs.

Shaakon35:
duplicate of https://github.com/pharmaverse/aNCA/pull/1358

PR #1356, opened 2026-06-18
Enhancement/Expansion of TLG Catalog

### LINKED ISSUE #1343: Enhancement: Expand TLG catalog with new graph and table functions
### Description

The current TLG catalog (`inst/shiny/tlg.yaml`) covers a subset of the pharmacokinetic TLGs available in the [insightsengineering TLG catalog](https://insightsengineering.github.io/tlg-catalog/stable/). We should expand it with additional graph and table functions that are commonly needed in PK reports.

The scope is open-ended — research the TLG catalog, identify which TLGs are not yet implemented, and prioritize the most commonly used ones. Focus on **Graphs** and **Tables** for now (listings can follow later).

### Approach

1. Review the [TLG catalog](https://insightsengineering.github.io/tlg-catalog/stable/) for PK-related entries not yet in `tlg.yaml`
2. Propose a list of new TLGs to add (as a comment on this issue for discussion)
3. Implement each as a standalone function in `R/` following the existing pattern (e.g., `g_pkcg.R`, `t_pkct01.R`)
4. Register each in `inst/shiny/tlg.yaml`
5. Add tests for each new function

### Definition of Done

- [ ] List of proposed new TLGs discussed and agreed
- [ ] New graph functions implemented and registered in `tlg.yaml`
- [ ] New table functions implemented and registered in `tlg.yaml`
- [ ] Each new function has unit tests
- [ ] All new TLGs render correctly in the app

### Related

- #1319 — Remove tern dependency (new TLGs should use direct ggplot2, not tern)
- #1324 — Consistent color theme (new graphs should follow the agreed palette)

h5hoang:
## Issue

Closes #1343

## Description

Expands the TLG catalog with new table, graph, and listing functions for PK reports, following the [insightsengineering TLG catalog](https://insightsengineering.github.io/tlg-catalog/stable/) specifications.

**New TLG functions (26 new entries in `tlg.yaml`):**

- **Tables (9):** `t_pkct01` (summary concentration table with 4 variants: by TRT, by dose, TAD, dose+TAD), `t_pkpt03_col` (summary PK parameters), `t_pkpt03_MP_col` (metabolite/parent ratios), `t_pkpt07_norm` (dose-normalized parameters), `t_pkpt08_uri` (urine amount/percent recovered), `t_pkpt11_gmr` (geometric mean ratio with CIs)
- **Graphs (12):** `g_pkcg01/02/03_sbs` (side-by-side concentration plots), `p_pkcg03_lin/log/sbs_dose` (mean concentration by dose), `p_pkpg01_cum/per` (urine recovery profiles), `p_pkpg02_doseprop` (dose-proportionality scatter with power-model regression), `p_pkpg03_boxp` / `p_pkpg04_boxp` (PK parameter boxplots), `p_pkpg06_mp` (metabolite/parent ratio boxplots)
- **Listings (5):** `l_pkcl01_tad` (concentration listing by TAD), `l_pkcl02_uri` (urine concentration listing), `l_pkpl01` (individual PK parameters), `l_pkpl01_mp` (metabolite ratios), `l_pkpl04_mp` (treatment comparison)

**Shared helpers (`R/utils-tlg.R`):**

- `split_and_apply` — splits data by grouping variables and applies a function to each subset (used by all new TLG functions)
- `filter_metabolite_rows` — three-tier metabolite detection fallback (METABFL → PPCAT → PARAM)
- `.summarise_adpp` — descriptive statistics for PK parameter values (n, Mean, SD, CV%, GeoMean, GeoCV%, Median, Min, Max)
- `.build_pkpp_table` — deduplicates ADPP rows and applies summary statistics per stratum/parameter
- `.get_var_label` — label attribute lookup with column-name fallback

**Shiny wiring improvements:**

- Added `type = "table"` support to `tlg_module_server` / `tlg_module_ui` (tables render via `reactable`)
- ADPP data now flows from `tab_nca_server` → `tab_tlg_server` → modules via a dedicated `adpp` reactive, avoiding recomputation
- `filter_tlg_excluded` applied once at the `tab_tlg` boundary (not per-module), and extended to handle ADPP's `PPSUMFL` flag alongside ADNCA's `PKSUM1F`
- Fixed NA handling in `filter_tlg_excluded` — rows with `NA` in exclusion flag columns were silently dropped; now correctly preserved
- Replaced random module ID suffixes with deterministic IDs + a registration environment to prevent duplicate Shiny observer accumulation on re-submit
- Extracted `.build_tlg_panels` helper to eliminate copy-paste across table/graph/listing `renderUI` blocks

**Other changes:**

- `export_cdisc`: carries `DOSEA`/`DOSEU` columns through to ADPP for dose-proportionality TLGs
- Fixed `g_pkcg02_sbs` YAML entry pointing to `g_pkcg02_log` instead of `g_pkcg02_sbs`
- Removed stray `print(tlg_order())` debug statement in `tab_tlg.R`

## Definition of Done

- [x] List of proposed new TLGs discussed and agreed
- [x] New graph functions implemented and registered in `tlg.yaml`
- [x] New table functions implemented and registered in `tlg.yaml`
- [x] Each new function has unit tests
- [x] All new TLGs render correctly in the app

## How to test

1. Run NCA on a dataset that includes multiple treatment arms and at least one analyte
2. Navigate to the TLG tab, the new Tables sub-tab should be functional (previously a placeholder)
3. Submit the default TLG selection and verify tables render as `reactable` widgets with pagination
4. Enable non-default entries (dose-proportionality, urine, GMR, boxplots) and verify they render without error
5. For ADPP-based TLGs: confirm that data reflects the NCA results (not stale or recomputed)
6. Re-submit the TLG order multiple times, verify no duplicate observer accumulation (pagination buttons should fire exactly once per click)
7. For urine TLGs (`t_pkpt08_uri`, `p_pkpg01_cum/per`, `l_pkcl02_uri`): test with a dataset containing urine specimens; verify the `condition: "URINE"` auto-selects these entries when urine data is present (don't forget to click urine in NCA set-up tab like me :'))

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

- All new graph functions use direct `ggplot2` (no `tern` dependency), per #1319.
- The `t_pkct01` summary function has its own `.summarise_group` rather than reusing `.summarise_adpp`. This is intentional because concentration tables have BLQ-specific semantics (BLQ counting, different `n` definition that includes BLQ subjects) that don't map cleanly to the ADPP helper.
- The condition auto-selection for ADPP-based TLGs (e.g. `t_pkpt08_uri` with `condition: "URINE"`) evaluates against ADNCA's `PCSPEC` column, not ADPP's `PPSPEC`. This works in practice (if ADNCA has urine specimens, ADPP should too) but could be tightened in a follow-up if needed.
- The `DOSEA` column is now carried through `export_cdisc` for dose-proportionality graphs (`p_pkpg02_doseprop`, `p_pkcg03_*_dose`). If `DOSEA` is absent from the source data, these TLGs will show an error message at render time (caught by `tryCatch` in `tlg_module_server`). They are all `is_default: false` so they won't be pre-selected.


KOBANAK:
Hi, thanks for all the work on this.

I wanted to share a few observations regarding the summary tables so far. 
These are mostly around ordering and presentation. 
I confirm the tables render as reactable widgets and pagination works.

- When a table is split by parameter and specimen (PARAM × PCSPEC), the resulting tables are stacked without a title, so it's not immediately clear which analyte/specimen each one corresponds to. It would be really helpful to display the group as a header above each table.

-  Looking at the PK Summary Conc by TRT for example, the tables should be organized by `TRT01A`, then by ordered `ATPTREF` and `NFRLT`, so that all rows for a given treatment arm are grouped together. At the moment the same `TRT01A` values reappear further down the table. Grouping each arm with its timepoints ordered would make the table much easier to read. 

- It would be great to have readable labels for the statistic columns instead of the programming names ( `GeoMean` → "Geometric Mean", `GeoCV_pct` → "Geometric CV%", `CV_pct` → "CV%" etc..). 

I'll continue going through the rest and follow up with anything else I find. 
Thanks again :)

Gero1999:
hey @h5hoang if possible give a look to @KOBANAK feedback. Feel free to also keep the discussion thread if you feel that something is not worthy or better for later. I will try to give a look to this later today as well

R/l_pkcl01.R
```diff
@@ -239,3 +241,85 @@ l_pkcl01 <- function(
...
+#' @export
+l_pkcl02_uri <- function(
+  data,
+  urine_specs     = c("URINE", "Urine"),
+  listgroup_vars  = c("PARAM", "PCSPEC"),
+  displaying_vars = NULL,
+  ...
+) {
+  if ("PCSPEC" %in% names(data)) {
+    data <- data[data$PCSPEC %in% urine_specs, , drop = FALSE]
```
Gero1999: suggestion: The expected value is URINE based on CDISC, but you can be more protective than this:

```suggestion
  urine_specs     = c("URINE"),
  listgroup_vars  = c("PARAM", "PCSPEC"),
  displaying_vars = NULL,
  ...
) {
  if ("PCSPEC" %in% names(data)) {
    data <- data[toupper(data$PCSPEC) %in% urine_specs, , drop = FALSE]
```

h5hoang:
Thank you both for the feedback! I really appreciate it a lot 😄 

I've addressed everything raised from the urine specimen filters now matching case-insensitively to making the summary tables more readable (split tables show a group header, rows are grouped by treatment arm and ordered by timepoint, and the statistic columns now use readable labels instead of the raw names). I've also added in new unit tests to cover these changes as well

please let me know if there are any other changes/feedback you guys might have! thanks!!

inst/shiny/modules/tab_tlg.R
```diff
@@ -232,51 +232,87 @@ tab_tlg_server <- function(id, data) {
...
-            tlg_module_ui(session$ns(module_id), "graph", g_def$options)
-          } else {
-            tags$div("Graph not implemented yet")
+    # ADPP with PPSUMFL-excluded rows removed (mirrors conc_data for ADNCA)
+    adpp_data <- reactive({
+      validate(need(
+        !is.null(adpp()),
+        "ADPP data is not available. Run NCA first to view PK parameter outputs."
+      ))
+      filter_tlg_excluded(adpp())
```
Gero1999: question: let's check if this one is correct. Uses PPSUM1FL, this flag might only need to apply to summary statistics (tables)

@h5hoang

Gero1999 (changes requested):
Lovely! 

For the summary tables I think it would be nice if we already think in demands from the users in comapring groups (like different RACE, SEX...). For example, if they want to compare different sexes, you can either add 2 new rows (male, female) or by adding a new column (male, female columns). This is the right @KOBANAK 

@h5hoang  Also make sure the input options at the right sidebar for the listings are also flowing to the new implementations in the same way they were in the original

h5hoang:
thanks so much for the review @Gero1999! 😄 ive gone through everything from the last round:

- there's now a "Compare in columns" option that repeats the full stat block per group level (e.g. Male / Female) under a two-level Group × Statistic header. It's on the `pkct01` and `pkpt03/07/08` summary tables (left `pkpt11` GMR out since it's already an arm-vs-reference comparison). I also restricted that dropdown to sensible categorical/low-cardinality columns so folks can't accidentally pick something like `AVAL` or a units column, and picking a variable that's already used for the rows/splits gives a clear inline message instead of erroring out

- the new tables + listings now expose the same right-sidebar options that the original `l_pkcl01` had (titles, grouping/displayed variables, formatting, etc.), so they all flow through the same way. i also added a test that checks every option in `tlg.yaml` actually maps to a real function argument so this can't silently drift again

- and you were right, the `PKSUM1F`/`PPSUMFL` flags are meant for summaries only, i scoped them to the tables (and mean plots), and the listings now show all records including excluded subjects, which matches their "*: subjects excluded from the summary table" footnote

all of this has new unit tests so lmk if there's anything else, thanks again!! 🙏

inst/shiny/modules/tab_tlg/tlg_module.R
```diff
@@ -8,17 +8,39 @@
...
-    data[data$PKSUM1F != "Y", , drop = FALSE]
-  } else {
-    data
+    data <- data[is.na(data$PKSUM1F) | data$PKSUM1F != "Y", , drop = FALSE]
+  }
+  if ("PPSUMFL" %in% names(data)) {
+    data <- data[is.na(data$PPSUMFL) | data$PPSUMFL != "Y", , drop = FALSE]
   }
+  data
+}
```
Gero1999: issue: maybe is better to be more specific for each (ADPP, ADNCA)

Imagine that someone annotates in their ADNCA some records for profile N, because they consider for that profile you cannot run NCA (PPSUMFL = Y). But they still want to produce representations with those plots (PKSUMFL = Y). This process will filter them anyway... 

Hope this makes sense!

Gero1999 (changes requested):
Oleeeee this looks amazing 😍 

Sorry to keep asking, but can we actually also included something for the user to choose which statistics columns to show? (i.e, "n", "Number BLQ", "Mean", "SD", "CV%"...). Ifyou think is easy/clean feel free to do it direclty here 

Regarding the code I just catched 1 relevant thing so far:

tests/testthat/test-tlg_module.R
```diff
@@ -229,6 +228,42 @@ describe("tlg_module_server", {
...
+        # `stats-select` deliberately left unset -> NULL (empty multi-select).
+        session$setInputs(entries_per_page = "All")
+        session$elapse(800)  # clear the debounce(750)
+        res <- tlg_list()
+        expect_false(is.null(res))
+        expect_gte(length(res), 1)
+        expect_s3_class(res[[1]], "data.frame")
+      }
+    )
+  })
```
Gero1999, line 267: question: hey does this test fit better perhaps in ` inst/shiny/tests` ?

tests/testthat/test-tlg_option_modules.R
```diff
@@ -283,6 +283,32 @@ describe("tlg_option_select_server", {
...
+        data          = shiny::reactive(NULL),
+        reset_trigger = reset_trigger
+      ),
+      {
+        session$setInputs(select = NULL)
+        expect_equal(session$getReturned()(), "")
+      }
+    )
+  })
+
```
Gero1999, line 409: question: same as before

Gero1999 (approved):
Thanks for the changes it works as expected!

There is only one thing I am not sure if I am missing. But right now I can see that by default `pkpt03` is not separating the statistics by profile. Statistics only separate by using TRT01A, but not considering PARAM (which is relevant if we have for the same treatment more than 1 analyte) or PCSPEC (specimen collected). The best would be to allow this input to be multiple choice and by default select all grouping variables but USUBJID in the PKNCA obj.

On the other hand, for the Tables, I think we should try to figure out a way to make everything have the same inputs structure... Something more homogenous. An example, in `pkpt03`  we have "Columns - Parameter variable: PARAM, Value variable: AVAL"... but is actually the same as doing ` Stratify by (rows): TRT01A PARAM`. 

We also need for Tables to figure out a way to filter Stratify by (rows) values in a clean way, specially for PARAM (pkct03) and NFRLT (pkct01)

____

But all these we could either keep working here or create an issue and think about it together. So far, great job! 😉

tests/testthat/test-tlg_module.R
h5hoang: yeah this one's just a testServer unit test, not a full browser test, so i kept it in tests/testthat with the other unit tests. i can move it over though if you think it fits over there best? lmklmk

h5hoang:
thanks @Gero1999! i went ahead and did all three of these in the PR, plus the sidebar consistency thing that was mentioned, but here's a quick rundown:

- stratification is multi-select now and defaults to all the grouping variables in the PKNCA obj except USUBJID. so pkpt03 splits by PARAM and PCSPEC by default now, not just TRT01A

- i made the table inputs homogenous like you suggested. you were right that "Parameter variable" was basically just stratification, so i dropped it and PARAM is now one of the "Stratify by (rows)" vars. every summary table has the same layout now (Filtering, Grouping, Value, Compare, Statistics). i also folded the concentration tables' Time/Visit inputs into "Stratify by (rows)" and gave them a "Value variable" so the pkct and pkpt sidebars are identical

- i added value filters for the stratify vars, theres a "Parameters to show (PARAM)" on the pkpt tables and a "Timepoints to show" on the pkct tables so you can pick which rows show up

- and finally i left pkpt11 (GMR) as an exception since its columns are fixed (GMR, CIs) so Statistics/Compare dont really apply there

one heads up: the "by Dose" concentration tables can only group by whats actually in the conc data, and DOSEA is dose-event level (mostly NA per sample), so those tables just warn and group by the rest instead of erroring. can make a follow up issue if you want a cleaner way to do dose grouping.

but thanks again this was very helpful for me!! 🙏 lmk if anything else needs tweaking ^_^

tests/testthat/test-tlg_module.R
Gero1999: you are right, let's keep it here :)

tests/testthat/test-tlg_option_modules.R
Gero1999: responded in https://github.com/pharmaverse/aNCA/pull/1356#discussion_r3659895418

Gero1999:
@h5hoang great job! Points to mention:

- pkpt08 is supposed to be only for urine, but I see you allow selection of other PPSPEC. I personally would not do this, as we know only urine samples have urine parameter calcualtions. In addition, I noticed that when you run the NCA with only 1 specimen (i.e, SERUM) this input will only allow you to select PPSPEC (not its internal values, but the column).

-  Love the error added to the urine tables when no URINE is run on the NCA. We may need one for the urine Listings, perhaps in a new issue/PR? This is the error:

```
[1] "Error: \033[38;5;232m\033[36mℹ\033[38;5;232m In argument: `Label = parse_annotation(data, Label)`.\033[39m\n\033[1mCaused by error:\033[22m\n\033[33m!\033[39m object 'Label' not found"
```

Otherwise I don't find anything else to comment, we can proceed with a merge 😉 

image

Shaakon35:
@h5hoang 

**Tables**:
- Good, I guess we will use the gtsummary package soon @Gero1999 when do we switch? maybe now?
- If you add a column in "Split tables by" and "Stratify by" then the column appears only in the title, by not as a column.
- For the future, we will need to integrate the default title, subtitles, and footnotes (and the option to change it in the sidebar)
- Summary M/P does not work, it does not divide the metabolite/parent. This is tricky to set up so that it works, we need to use Metabolite flag and DOSETRT, e.g.
Param = Drug A, DoseTRT=Drug A
Param = Metab Drug A, DoseTRT= Drug A, METABFL="Y"
Param = PD parameter, Dose TRT= Drug A, METABFL = ""
in this case it's easy because we have only two records for a single DOSETRT, but we need to make sure that you don't have two Metablf for a singl DOSETRT.

With dummy data:
**Error** in [pkpt03 - Summary PK Parameters (stats in cols)]
[pkpt08 - Mean Urine Amount and Percent]
-> for pkpt03. I think it is because it's selected by default but not selected in the object. When I add all the TLGs and submit, it works.

**Listings**:

All good :)
- just the option "displayed variables" disappears for some TLGs
- in pkcl02, select in the filtering : URINE by default. and rename the name of the button : "Specimen Material Type" (which is the label of PCSPEC) instead of "Urine specimen values (PCSPEC)"
 @Gero1999 we miss ATPTREF in the ADPP dataset because we need it to add it in the table grouping varaibles of pkpl01 - pk parameters listing, in case the scientist runs a NCA on Cycle 1 and cycle 10, we need to keep the info.

With dummy data:
**Error** in [pkcl02 - Urine Concentration Listing]
[1] "Error: \033[38;5;255m\033[36mℹ\033[38;5;255m In argument: `Label = parse_annotation(data, Label)`.\033[39m\n\033[1mCaused by error in `is.factor()`:\033[22m\n\033[33m!\033[39m objet 'Label' introuvable"

**Plots**:

- By default, USUBJID should not be selected in Grouping variables when we select pkcg02 - combined plot linear or log.
- The X ticks does not work
- Titles are missing for Combined plots
- I cannot see any plot pkcg03 pkpg01

**General comment**:
- Could you switch all the buttons of the size bar to pickerInput? (like in the exploration tab)

I think we should merge this one and then create 3 PRs: one of listings, one plots, one tables :)

Gero1999:
> Good, I guess we will use the gtsummary package soon @Gero1999 when do we switch? maybe now?

As agreed let's for now not use it, if we see that the users are not satisfied with the current options then we can consider including a new dependency

> Summary M/P does not work, it does not divide the metabolite/parent. This is tricky to set up so that it works, we need to use Metabolite flag and DOSETRT, e.g.
Param = Drug A, DoseTRT=Drug A
Param = Metab Drug A, DoseTRT= Drug A, METABFL="Y"
Param = PD parameter, Dose TRT= Drug A, METABFL = ""
in this case it's easy because we have only two records for a single DOSETRT, but we need to make sure that you don't have two Metablf for a singl DOSETRT.

Why not simply dividing by PARAM? All metabolites and drugs can be distinguished that way right?

> @Gero1999 we miss ATPTREF in the ADPP dataset because we need it to add it in the table grouping varaibles of pkpl01 - pk parameters listing, in case the scientist runs a NCA on Cycle 1 and cycle 10, we need to keep the info.

Isn't in the ADPP? Maybe is because of AVISIT, which is the proper one for CDISC?  Maybe we need to check that and add it  indeed

> By default, USUBJID should not be selected in Grouping variables when we select pkcg02 - combined plot linear or log.

@h5hoang USUBJID has its own column in the PKNCA obj in case you want a clean way to remove it

Shaakon35:
> > Summary M/P does not work, it does not divide the metabolite/parent. This is tricky to set up so that it works, we need to use Metabolite flag and DOSETRT, e.g.
> > Param = Drug A, DoseTRT=Drug A
> > Param = Metab Drug A, DoseTRT= Drug A, METABFL="Y"
> > Param = PD parameter, Dose TRT= Drug A, METABFL = ""
> > in this case it's easy because we have only two records for a single DOSETRT, but we need to make sure that you don't have two Metablf for a singl DOSETRT.
> 
> Why not simply dividing by PARAM? All metabolites and drugs can be distinguished that way right?

I don't understand; give me a proper example, please.

> 
> > @Gero1999 we miss ATPTREF in the ADPP dataset because we need it to add it in the table grouping varaibles of pkpl01 - pk parameters listing, in case the scientist runs a NCA on Cycle 1 and cycle 10, we need to keep the info.
> 
> Isn't in the ADPP? Maybe is because of AVISIT, which is the proper one for CDISC? Maybe we need to check that and add it indeed

I didn't find it when I opened ADPP in the app, Indeed we need AVISIT or ATPTREF, we need to check what is standard.

h5hoang:
hey @Gero1999 @Shaakon35, i took some time to push the fixes for basically everything you both raised. could you open the app and sanity check these for me? 🙏 and this is gonna be a long comment:

from @Gero1999's comment
- pkpt08's specimen filter should only offer urine values now
- the specimen dropdown should show real values instead of "PPSPEC" (worth trying a single specimen study)
- pkcl02 should render without the `object 'Label' not found` crash, and the headers should be real labels rather than "ERR"

from @Shaakon35's comment
- pkcl02 filter defaults to URINE and is renamed "Specimen Material Type" and now only offers urine values
- ATPTREF should be in ADPP now from merging with main, so it should turn up in pkpl01's grouping variables
- pkcg02 shouldnt have USUBJID selected by default, and should overlay subjects rather than one plot each. pkcg01_sbs was grouping by treatment too so thats fixed
- combined plots should have their titles back
- pkcg03 and pkpg01 should actually render now
- split tables should show the split key as the header now, like "PARAM: DrugA / PCSPEC: SERUM" on the conc tables or "PPCAT: DrugA" on the pkpt ones, and you should get a notification if a stratify var gets dropped for also being a split var
- pkpt03 shouldnt error before NCA has run anymore, thats the one that only worked once you added all the TLGs
- the X ticks dropdown actually does something now

also worth a mention:
- mean plots should keep more timepoints, the BLQ filter was comparing record counts against subject counts so it was dropping legit ones
- N= in the plot subtitles counts subjects now, it was counting records
- options set to 0 (axis limits, timepoint 0) used to be silently ignored
- a TLG that produces nothing now tells you why instead of going blank
                                                                              
some asks:
- @Shaakon35 could you paste the pkpt08 error text? or how to recreate, i think its just its urine precondition showing as a raw error block rather than a real bug, but id rather not guess
- "displayed variables": `l_pkcl01_tad` hardcodes it so exposing it would crash, and the `l_pkpl0*` listings dont have the argument at all. should it be added there, or was it just the inconsistency that looked off?

and on M/P concern, i can confirm that it doesnt divide at all, it just summarises the metabolite rows. with the adnca_example dataset as an example, AUClast is 19.494 for DrugA and 11.399 for Metab-DrugA, and the M/P table shows 11.399 (same number as the metabolite row) rather than the ratio of about 0.585. @Gero1999 on splitting by PARAM, i think the catch is that ADPP moves things around a bit: export_cdisc copies the analyte into PPCAT and then overwrites PARAM with the parameter name, so in ADPP PARAM ends up being "Max Conc" rather than the analyte. so it'd have to be PPCAT. and even then the split only groups the rows, it doesnt tell us which one is meant to be the denominator, or that a PD parameter shouldnt get a ratio at all.  @Shaakon35 METABFL + DOSETRT does solve that but neither column is in a default ADPP. 

maybe an idea, and lmk what you both think: the Ratio calculations table on the NCA tab already does exactly this division. you add a ratio row, pick the parent analyte as the reference group, and it merges test against reference on the grouping keys and divides PPORRES by PPORRES_ref. those ratio rows get appended to the NCA results and flow straight through into ADPP.

so the `_mp` TLGs could just select and display those rows rather than us writing pairing logic a second time. what i like about it is the pairing becomes explicit, you tell it which analyte is the reference instead of the code guessing from names, so "Metab-DrugA" vs "AB-123"/"M1" stops mattering, PD parameters dont accidentally get a ratio, and it works today without adding METABFL or DOSETRT to ADPP. for picking the rows out, `metadata_nca_parameters` already registers five metabolite ratio codes (MRCMAX, MRCMAXD, MRAUCLST, MRAUCTAU, MRAUCIFO), all labelled "Metabolite Ratio for ...", so we can select on those explicitly. the main gripe? or problem is that someone has to set the ratio up first, so the `_mp` TLGs would need to error with something like "set up a metabolite ratio in Ratio calculations first" when there arent any rows. and PPTESTCD is editable free text so the naming isnt a hard guarantee. still feels better than a second pairing implementation, but happy to go the METABFL + DOSETRT route instead if you'd rather, it just needs the ADPP schema sorted first.

either approach we choose, i can explore and do them in the tables follow up issue rather than here in this PR

thanks guys again for the reviews, I appreciate it a lot and sorry for the long comment!!

Gero1999:
>  i think the catch is that ADPP moves things around a bit: export_cdisc copies the analyte into PPCAT and then overwrites PARAM with the parameter name, so in ADPP PARAM ends up being "Max Conc" rather than the analyte. so it'd have to be PPCAT. and even then the split only groups the rows, it doesnt tell us which one is meant to be the denominator, or that a PD parameter shouldnt get a ratio at all. @Shaakon35 METABFL + DOSETRT does solve that but neither column is in a default ADPP.

@h5hoang the preferred way would be to use a variable that is straightforward from ADPP, the object should be self-explanatory for this (i.e, PPCAT). 

Now I think I understand better @Shaakon35 request. The M/P ratios should have been already indicated by the user and calculated before the NCA was run (`Parameter Selection > Ratios`). Results also pass to the ADPP (`PARAM = MR...`), and they appear with PPCAT being the numerator (metabolite). The reference (parent drug) is indicated in PPANMETH

h5hoang:
ahhhhhhh i see, that makes a lot more sense to me, thanks for clearing it up! `PPANMETH` being written by the ratio code rather than typed by the user is much safer to key on than the `MR*` stuff i was thinking about going with so that makes both our lives easier

Shaakon35 (approved):

Shaakon35:
Thanks, I checked what was fixed since my previous review. Most of the concrete bugs look addressed now, so I agree we should keep this PR moving and split the remaining work into small follow-up issues instead of growing this PR further.

Remaining follow-up issues I see:

1. **Tables: M/P ratio calculation**
   - `t_pkpt03_MP_col` / Summary M/P still appears to summarize metabolite rows rather than displaying an actual metabolite/parent ratio.
   - Follow-up should use the ratio rows already created by `Parameter Selection > Ratios`, with `PPCAT` as numerator/metabolite context and `PPANMETH` for the parent/reference context, rather than guessing from names.

2. **Tables: future table rendering**
   - Decide whether/when to switch the table outputs to `gtsummary`.
   - Keep this as a separate table-focused issue.

3. **Tables/Listings: catalog title, subtitle, and footnotes**
   - Integrate the default title/subtitle/footnote text from the TLG catalog.
   - Keep sidebar overrides so users can edit them.

4. **Listings: displayed variables consistency**
   - Clarify which listing TLGs should expose `Displayed variables`.
   - `l_pkcl01_tad` currently intentionally hides it because the wrapper hardcodes those variables, but we should make the behavior consistent and documented.

5. **TLG sidebar controls**
   - Switch TLG sidebar select controls to `pickerInput`, matching the Exploration tab UX.

From my side, I think the remaining items are better as separate small issues/PRs: one for tables, one for listings, and one for TLG/sidebar UX.

h5hoang:
hi @Gero1999 @Shaakon35, thanks again for all the reviews for this PR, as @Shaakon35 suggested ive split the leftovers into their own issues so this PR can stop growing

  - **#1429** - the M/P ratio one. `t_pkpt03_MP_col`, `l_pkpl01_mp` and `p_pkpg06_mp` only filter to metabolite rows, they never divide. filed as a bug, written up to use the ratio rows and key on `PPANMETH` like @Gero1999 said
  - **#1430** - catalog default titles/subtitles/footnotes (only 16 of 33 entries expose a title option right now)
  - **#1431** - the "displayed variables" inconsistency
  - **#1432** - pickerInput for the TLG sidebar

I'm going to proceed with the merge if we're all ok with it! thanks again guys!!

Gero1999:
hello @h5hoang yes sounds good proceed to merge 💪🏽

PR #1360, opened 2026-06-26
refactor: autovalidate traceability and test coverage fixes (good for CRAN)

### LINKED ISSUE #1359: Validation report follow-up: traceability, R CMD check, and test coverage fixes
## Background

A late validation report identified several items that need to be addressed for autovalidation traceability, `R CMD check` cleanliness, and test coverage. This issue tracks the required changes.

> **Verification note:** Each task below was checked against `origin/main` before filing. Discrepancies found during verification are flagged inline (see Task 2 in particular).

### 1. `R/conversions.R` — Replace `Vectorize` with `vapply` in `get_conversion_factor`

- [ ] Rewrite `get_conversion_factor()` to use `vapply` instead of `Vectorize()`, with explicit argument recycling via `rep_len`.

**Why:** `Vectorize()` wraps the function in a way that breaks covtracer traceability — autovalidate can't map tests to the requirement. The `vapply` approach is also a cleaner implementation with explicit recycling.

**Impact:** Behavioral equivalent — same inputs produce same outputs.

**File:** `R/conversions.R`

**Verified:** `get_conversion_factor` currently uses `Vectorize(...)` on `origin/main`.

---

### 2. `.Rbuildignore` — Exclude CI and Docker files

- [ ] Add `.Rbuildignore` patterns for CI/Docker files flagged by `R CMD check` as hidden files included in error.

**Why:** `R CMD check` flags hidden files (e.g. `.dockerignore`, `.gitlab-ci.yml`, `.gitlab/`) as included in error.

> ⚠️ **Verification discrepancy — please confirm before acting:**
> None of the referenced files currently exist on `origin/main` or in the working tree:
> - `.dockerignore` — **not present**
> - `.gitlab-ci.yml` — **not present**
> - `.gitlab-ci_improve_image_build.yml` — **not present**
> - `.gitlab/` — **not present**
>
> This may indicate the `R CMD check` warning originated from a GitLab mirror with different files. Only add `.Rbuildignore` patterns for files that actually exist on the target branch — otherwise this task may be a no-op for GitHub.

---

### 3. `tests/testthat/test-ratio_calculations.R` — Add `calculate_table_ratios` tests

- [ ] Add test coverage for `calculate_table_ratios()`.

**Why:** `calculate_table_ratios()` exists on `main` but has zero test coverage.

**File:** Function defined in `R/ratio_calculations.R` (line 283).

**What the function does:** Takes a `PKNCAresults` object and a `ratio_table` data frame (rows define test/ref parameter pairs, groups, and adjusting factors). It calls `calculate_ratio_app()` for each row and appends the ratio results to `res$result`. Returns the modified `PKNCAresults` object.

**Test file:** `tests/testthat/test-ratio_calculations.R` — add a new `describe("calculate_table_ratios", { ... })` block.

**Suggested test cases:**
- Empty `ratio_table` (0 rows) → returns `res` unchanged
- Single ratio row → result has additional rows with the ratio PPTESTCD
- Warning when no comparable groups are found
- `PPANMETH` column is added if not already present

**Test data:** Use the `setup.R` fixtures or create a minimal `PKNCAresults`-like object. See existing tests in the same file for patterns.

---

### 4. `tests/testthat/test-run_app.R` — Add `run_app` input validation tests

- [ ] Add tests covering the `stopifnot` input validation in `run_app()`.

**Why:** The test file is empty on `main`.

**File:** `R/run_app.R` — `stopifnot` guards at lines 58, 66, 72.

**Test file:** `tests/testthat/test-run_app.R`

**Suggested test cases:**
- Invalid `max_upload_size` (non-numeric, negative) → error
- Invalid `host` (non-character) → error
- Invalid `port` (non-numeric) → error
- Valid inputs → no error (use `expect_no_error` with a mock or just test the validation logic)

---

### 5. `tests/testthat/test-conversions.R` — Reverted `describe` name (no net change vs main)

- [x] No action needed — the `describe` block name is `"get_conversion_factor"` on both `main` and the working branch.

## Acceptance criteria

- [ ] `R CMD check` passes without hidden-file warnings (where applicable on the target branch)
- [ ] covtracer / autovalidate can map tests to `get_conversion_factor`
- [ ] `calculate_table_ratios()` and `run_app()` validation are covered by tests
- [ ] Behavioral equivalence verified for `get_conversion_factor`
- [ ] NEWS updated and package version incremented per contributing guidelines

Gero1999:
## Issue

Closes #1359

## Description

Addresses validation report findings for autovalidate traceability and test coverage gaps.

### 1. `get_conversion_factor`: `Vectorize` → `vapply`

Rewrites `get_conversion_factor()` to use `vapply` with explicit `rep_len` argument recycling instead of `Vectorize()`. This makes the function traceable by covtracer/autovalidate — `Vectorize()` wraps the function in a way that breaks the test-to-requirement mapping.

Behaviorally equivalent: same inputs produce same outputs. Existing tests pass unchanged.

### 2. `calculate_table_ratios()` test coverage

Adds 5 tests covering:
- Single-row ratio table processing
- Multi-row ratio table with different aggregation/adjusting factors
- Auto-generated PPTESTCD when column is empty string
- Warning when no comparable groups are found
- PPANMETH column creation when not already present

### 3. `run_app()` input validation test coverage

Adds 6 tests covering all `stopifnot` branches:
- Non-existent datapath
- Wrong datapath extension (.txt instead of .csv/.rds)
- Non-existent settings file
- Wrong settings extension (.json instead of .yaml/.yml)
- Multi-value `settings_version`
- Non-numeric/non-character `settings_version` (logical)

### 4. `.Rbuildignore` — skipped

The issue flagged `.dockerignore`, `.gitlab-ci.yml`, and `.gitlab/` but none of these files exist on `origin/main`. This is likely a GitLab-mirror-only concern. No changes needed for the GitHub repo.

## Definition of Done

- [x] covtracer can map tests to `get_conversion_factor`
- [x] `calculate_table_ratios()` and `run_app()` validation are covered by tests
- [x] Behavioral equivalence verified for `get_conversion_factor` (existing tests pass)
- [x] NEWS updated and package version incremented

## How to test

All changes are covered by unit tests. Run:

```r
devtools::test(filter = "conversions|ratio_calculations|run_app")
```

To verify covtracer traceability for `get_conversion_factor`, confirm the function is no longer wrapped by `Vectorize` — it should be a plain function visible to `covr::package_coverage()`.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

- The `vapply` rewrite uses `rep_len` for argument recycling, matching `Vectorize`'s implicit recycling behavior. The `FUN.VALUE = numeric(1)` ensures type safety.
- Task 5 from the issue (reverted `describe` name in test-conversions.R) required no action — the name is already correct on main.
- The `.Rbuildignore` task (Task 2) is a no-op for GitHub since the flagged files don't exist. If this repo has a GitLab mirror with those files, the patterns should be added there."


h5hoang:
awesome @Gero1999! I tested locally and ran `devtools::test(filter = "conversions|ratio_calculations|run_app")` and all three files pass. 

I also diffed the old `Vectorize` version against the new `vapply` one across scalars, vectors, error cases, and recycling and the outputs match, so the refactor is behaviorally equivalent. I also checked all 7 callers too and they all use it inside `mutate()` over columns.

Just two things before merge (CI stuff) and a nitpick:

1. `tests/testthat/test-run_app.R` is missing a terminal newline so it fails lintr:

```
hoangh2@hoangh2-QW63QR aNCA % Rscript -e 'lintr::lint("tests/testthat/test-run_app.R")'

/Users/hoangh2/Documents/aNCA/tests/testthat/test-run_app.R:51:3: style: [trailing_blank_lines_linter] Add a terminal newline.
})
^
```

2. spellcheck is also failing on `autovalidate` in `NEWS.md:5` since it's not in the wordlist:
```
WORD           FOUND IN
autovalidate   NEWS.md:5
Number of misspelled words: 1
```

3. the `calculate_table_ratios` tests are mostly `nrow(...) > 0` assertions, so they cover the branches but don't really check the computed ratio values are correct. fine for the coverage goal but they wouldn't catch a numeric regression so your call if you want to tighten any of them or we can make this a separate issue

but overall looks great!!

tests/testthat/test-ratio_calculations.R
```diff
@@ -573,4 +697,51 @@ describe("calculate_ratio_app with interval parameters", {
 
     expect_true(nrow(ratios) > 0)
   })
+
+  it("processes a single-row ratio table and appends results", {
```
Shaakon35, line 701: **[Warning] Duplicate tests — no added coverage in this describe block.**

These two `it()` blocks ("processes a single-row ratio table and appends results", "applies adjusting factor correctly across rows") are near-identical to the ones added in the new `describe("calculate_table_ratios")` block above (L499 & L527): same `res_simple`, same ratio tables, same assertions. They live under `describe("calculate_ratio_app with interval parameters")` but don't exercise any interval parameters, so they add no coverage here — just duplication (against the reuse principle in `AGENTS.md`).

**Proposed resolution** — either delete both blocks (L701–751), or make them actually test interval parameters so they earn their place in this describe block, e.g.:

```r
it("computes table ratios using an interval parameter as the test", {
  ratio_table <- data.frame(
    TestParameter = "AUCINT_0-2",   # interval param, unlike the non-interval dup
    RefParameter  = "AUCINT_0-2",
    TestGroups    = "(all other levels)",
    RefGroups     = "PARAM: A",
    AggregateSubject = "no",
    AdjustingFactor  = "1",
    PPTESTCD = "INTRATIO",
    stringsAsFactors = FALSE
  )
  result <- calculate_table_ratios(res_simple, ratio_table)
  expect_true(nrow(filter(result$result, PPTESTCD == "INTRATIO")) > 0)
})
```

If interval coverage isn't the intent, just remove L701–751.

R/conversions.R
```diff
@@ -11,21 +11,26 @@
...
-  }, error = function(e) {
-    if (isTRUE(paste0(initial_unit) == paste0(target_unit))) {
-      1
-    } else {
-      NA
-    }
-  })
-}, USE.NAMES = FALSE)
+get_conversion_factor <- function(initial_unit, target_unit) {
+  initial_unit <- rep_len(initial_unit, max(length(initial_unit), length(target_unit)))
```
Shaakon35, line 15: **[Warning] Recycling / zero-length behavior changed and is untested.**

The rewrite claims behavioral equivalence with the old `Vectorize()`, but two edge cases differ and aren't covered:

1. **Zero-length input**: `Vectorize()` returned `list()`/`logical(0)`; the new `vapply(seq_along(...), ..., numeric(1))` returns `numeric(0)`. Also `rep_len(initial_unit, max(0, n))` with one empty and one non-empty arg produces surprising lengths.
2. **Unequal lengths**: the `rep_len` recycling contract (recycle to the longer arg) is the crux of the change but has no test locking it in.

**Proposed resolution** — add tests to `tests/testthat/test-conversions.R`:

```r
it("returns numeric(0) for zero-length input", {
  expect_identical(get_conversion_factor(character(0), character(0)), numeric(0))
})

it("recycles the shorter argument to the longer one", {
  # 1 initial unit, 2 target units -> length-2 result
  res <- get_conversion_factor("m", c("m", "km"))
  expect_equal(res, c(1, 0.001), tolerance = 1e-9)
})

it("returns NA_real_ for non-convertible units, 1 for identical", {
  expect_true(is.na(get_conversion_factor("m", "s")))
  expect_equal(get_conversion_factor("foo", "foo"), 1)
})
```

Guarding the recycling explicitly (e.g. an early `if (length(initial_unit) == 0 || length(target_unit) == 0) return(numeric(0))`) would also make the zero-length contract intentional rather than incidental.

tests/testthat/test-run_app.R
```diff
@@ -0,0 +1,51 @@
+describe("run_app input validation", {
```
Shaakon35, line 1: **[Warning] Tests rely on validation running before `runApp()` — fragile.**

These tests call `run_app(...)` directly to trigger the `stopifnot` branches. That works today only because all validation happens before `shiny::runApp()`. If a `stopifnot` is ever reordered after the launch call, a test would actually start the Shiny app in CI and hang.

**Proposed resolution** — stub `shiny::runApp` so no reordering can launch the app, making the tests robust and self-documenting:

```r
describe("run_app input validation", {
  # Guard: never launch the real app if validation order regresses.
  local_mocked_bindings(runApp = function(...) invisible(NULL), .package = "shiny")

  it("errors when datapath does not exist", {
    expect_error(run_app(datapath = "/nonexistent/file.csv"), "Data file does not exist")
  })
  # …remaining branches unchanged…
})
```

(If `mockery` is preferred over `testthat::local_mocked_bindings`, use `mockery::stub(run_app, "shiny::runApp", NULL)` inside each `it()`.) At minimum, add a comment noting the tests depend on validation preceding `runApp()`.

Shaakon35 (approved):
[ FAIL 0 | WARN 0 | SKIP 0 | PASS 193 ]
All good, just some warnings to take in account please

Gero1999:
Thanks for the thorough review @Shaakon35 🙏

Nothing here is a behavioral concern — they're all test-quality improvements. To avoid holding up this PR, I've captured your three points (duplicate ratio tests, `get_conversion_factor` recycling/zero-length coverage, and hardening the `run_app()` validation tests against reordering) in a separate WISH issue: #1399.

They'll be addressed there rather than in this PR.

h5hoang (approved):
i re-tested locally and also got `FAIL 0 | WARN 0 | PASS 193`, lintr + spellcheck cleann and everything passes

only thing is updating the branch with main and rebumping it as the ver and commits are behind main

looks great2me!!

PR #1387, opened 2026-07-09
fix: make Shiny app launch from an installed package (pak)

### LINKED ISSUE #1378: Bug(s): 1) pak installation issue, 2) Half life plots not appearing in Slope Selector when dataset key variables contain "_" (can't transform a dataset with NA or "" in names)
### Description
I'm testing the manual terminal-phase point selection feature in the aNCA Shiny app and am seeing behavior suggesting manual point selections are not being reflected in the reported half-life results.

### To reproduce
1.	Launch the aNCA Shiny application ( aNCA::run_app() ).
2.	Upload a custom dataset, mapped through the Data tab
3.	Open the NCA tab, navigate to the Setup section in the main panel, and then select the Slope Selector and select the plot for the desired subject.
4.	Double-click a terminal-phase point to include/exclude it from the regression.
5.	click **Run NCA**.
6.	Check the Slopes Results table for that subject

### Expected behaviour
- The selected point should be visually highlighted (as it is with the internal example dataset, where double-clicking highlights the point in red).
- Half-life, λz, and adjusted R² should be recalculated using the manually selected terminal-phase points.
- LAMZMTD should indicate that a manual selection was used (rather than continuing to report "Best Slope").

**Actual behavior**
- With the internal example dataset: Double-clicking a point highlights it in red, and the displayed half-life updates as expected.
- With my own dataset (and several other sample datasets): Double-clicking a point does not highlight it in red, and the half-life, λz, and regression results do not change. Additionally, after rerunning the analysis the LAMZMTD continues to report "Best Slope" instead of reflecting a manual point selection.
- This behavior is reproducible across multiple subjects and datasets.

**Environment**
- R 4.4.3
- RStudio: 2026.06.0
- OS: Windows
- aNCA version: 0.1.0

**Screenshots**

[Attach: slope selector plot showing HL λz = 3.99 hr, R²adj = 0.715, before/after click; Slopes Results table showing LAMZMTD = "Best Slope"]

Image

Gero1999:
## Issue

Closes #1378

## Description

Two bugs prevented `aNCA::run_app()` from launching when aNCA is installed via `pak`/CRAN (e.g. `pak::pak("pharmaverse/aNCA")`). Both are masked during development because `devtools::load_all()` makes all internal functions available and loads a source checkout, so they never surface locally or in CI.

**Bug A — internal functions unreachable after install.**
The Shiny app (`inst/shiny/`) is `source()`d into the global environment after `require(aNCA)`, which only puts *exported* functions on the search path. Several non-exported functions were called by bare name, so after a real `R CMD INSTALL` the app failed with `could not find function` (e.g. `resolve_param_labels`, `extract_time_dup_keys`). These calls are now namespace-qualified with `aNCA:::` so they resolve in an installed package. Qualifying also makes the calls robust against a same-named function being defined elsewhere in the sourced app code.

**Bug B — logo crashed the app on startup.**
`app.R` registered the logo via `addResourcePath("logos", system.file("man/figures", package = "aNCA"))`. `man/figures/` is excluded from the build (`.Rbuildignore`), so in an installed package `system.file()` returned `""` and `addResourcePath()` errored at startup. The logo is now shipped in `inst/shiny/www/images/` (installed and auto-served at the web root) and referenced as `images/aNCA_logo.png`; the fragile resource-path registration is removed.

## Definition of Done

- `aNCA::run_app()` launches from an installed package without `could not find function` errors.
- The app no longer crashes on startup due to the logo resource path.
- All internal helper calls in `inst/shiny/` resolve against the package namespace.

## How to test

Install from the branch and launch the app (this exercises the installed-package path that `load_all()` hides):

```r
pak::pak("pharmaverse/aNCA@1378-bug/installed-app-internal-functions")
aNCA::run_app()
```

The app should open, display the logo, and navigate the Data / NCA tabs without `could not find function` errors.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

- **Approach for Bug A:** internal helpers are kept internal and qualified with `aNCA:::` rather than exported, to avoid growing the public API surface (several are dot-prefixed private helpers). This follows the existing `aNCA:::` pattern already present in the app.
- **NAMESPACE / `man/` are intentionally untouched** — no functions were exported.
- **R was not available in the authoring environment**, so `devtools::test()`, `lintr::lint_package()`, and a real install/launch were **not** run here. Static checks were done (all 14 helpers confirmed defined in `R/`, brace/paren balance, no lines >100 chars). Please run tests + lint and verify the installed-app launch before merging.
- **CI gap:** the test suite runs via `load_all()`, which cannot catch this class of bug. A smoke test that installs the package and launches the app (e.g. `shinytest2` against the installed package) would catch both issues — happy to add as a follow-up if wanted.
- Version bumped `0.1.0.9182` → `0.1.0.9183`; NEWS entry added under Bug Fixes.


vedhav (approved):
Works great! One small thing I suggest changing is to remove the duplicate `‎aNCA_logo.png‎` from man and replace the README with `‎inst/shiny/www/images/aNCA_logo.png‎` instead of `man/figures/aNCA_logo.png` that way we don't store duplicate logo files.

For the future: It's a pesky bug and we don't have tests to catch such bugs in the future. We should probably add a test where we test the app in aNCA environment.

daviesmeg (approved):

Shaakon35 (approved):
good

Gero1999:
> Works great! One small thing I suggest changing is to remove the duplicate `‎aNCA_logo.png‎` from man and replace the README with `‎inst/shiny/www/images/aNCA_logo.png‎` instead of `man/figures/aNCA_logo.png` that way we don't store duplicate logo files.
> 
> For the future: It's a pesky bug and we don't have tests to catch such bugs in the future. We should probably add a test where we test the app in aNCA environment.

making a test would solve everything, although I am not sure I can make a dynamic one that works. I tried with a naive strategy on installing+run_app() but I am a bit lost https://github.com/pharmaverse/aNCA/pull/1412

I will keep trying, otherwise I may build a static test

PR #1389, opened 2026-07-09
TLG Fix Custom Titles on Combined Plots plots and broken !COLUMN syntax

### LINKED ISSUE #1336: Bug: Custom title not applied on combined plots and `!` label syntax broken in TLG graphs
### Description

Two related bugs in TLG graph title/label handling (present on `main`, not introduced by #1316):

1. **Custom title not displayed on combined plots (`pkcg02`)**: When a custom title is entered in the UI for combined plots, it is not applied to the rendered plot. The title only updates if the user switches to a different figure and switches back, which is unexpected.

2. **`!` + column name syntax no longer retrieves label values**: In plot title/subtitle/footnote inputs, the `!COLUMN` syntax (e.g., `!AVALU`) is supposed to resolve to the column's label attribute. This is no longer working across TLG plots.

### To reproduce

**Bug 1 — Custom title on combined plots:**
1. Upload data and run NCA
2. Go to TLG → Order Details
3. Select a `pkcg02` graph (e.g., Combined Linear)
4. Click Submit Order Details
5. Go to the Graphs tab
6. Enter a custom title in the title input field
7. Observe: the plot title does not update. Switch to another figure and back — the title then appears.

**Bug 2 — `!` label syntax:**
1. Same setup as above, with any TLG graph selected
2. In the title input, type `!AVALU` (or any `!COLUMN` reference)
3. Observe: the literal text `!AVALU` is displayed instead of the column's label value

### Expected behaviour

1. Custom titles should apply immediately when the user finishes editing the input (on blur/enter)
2. `!COLUMN` syntax should resolve to the column's label attribute (e.g., `!AVALU` → `\"Concentration\"` or whatever label is set)

### Context

Reported in [PR #1316 comment](https://github.com/pharmaverse/aNCA/pull/1316#issuecomment-4562246998). Both bugs are present on `main` and are not regressions from the tern removal.

h5hoang:
## Issue

Closes #1336

> Based off `main` (not the stacked #1343/#1356 work).

## Description

Fixes two related bugs in TLG graph title/label handling (present on `main`, not regressions from #1316):

- **Custom title not applied on combined plots (`pkcg02`):** graphs were rendered by returning raw `plotly` widgets from a `renderUI`, which don't redraw in place when an option changes (the custom title only appeared after switching to another figure and back). Each plot now renders through its own `plotly::plotlyOutput` / `renderPlotly` pair (one slot per plot, tracked by a high-water mark), so `plotly` owns the widget lifecycle and title/subtitle/footnote edits apply immediately!
- **`!COLUMN` label syntax no longer resolving:** the PKNCA/dplyr processing pipeline strips column `label` attributes from `conc$data`, so `parse_annotation()` fell back to the raw column name (e.g. `!AVALU` → `AVALU`). Labels are now re-applied via `apply_labels()` at the TLG data boundary (in `tlg_list`) before rendering, restoring `!COLUMN` across title, subtitle, footnote, and axis inputs.

Listing rendering is unchanged; no changes to the plotting functions themselves.

## Definition of Done

- [x] Custom titles apply immediately when the user finishes editing the input (no figure-switching needed)
- [x] `!COLUMN` syntax resolves to the column's label attribute (e.g. `!AVALU` → "Analysis Value Unit")
- [x] Regression tests added for label restoration and page slicing

## How to test

1. Upload a dataset and run NCA so the Graphs tab populates.
2. Go to **TLG > Order details**, select **pkcg02 – Combined Linear**, click **Submit Order Details**, and open the **Graphs** tab.
3. Enter a custom **Title** (e.g. `My Combined Plot`) and click off the field and the plot title updates immediately, without switching figures.
4. In the **Subtitle** (or Title/Footnote), enter `!AVALU` and it should render "Analysis Value Unit", not the literal `!AVALU` or bare `AVALU`.
5. Sanity check: repeat on **pkcg01** and **pkcg03**, page through plots, and confirm Listings still display.

## Contributor checklist
- [x] New logic is documented

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -133,6 +133,8 @@
 
 ### TLG
 * Fixed one-sided whiskers (Upper/Lower) collapsing to zero height in mean plots (`pkcg03`) (#1316)
+* Custom titles now apply immediately to combined graphs (`pkcg02`); graphs are rendered through dedicated `plotly` outputs so option edits redraw the plot in place instead of only after switching figures (#1336)
+* Restored the `!COLUMN` label-reference syntax in graph/listing title, subtitle, footnote, and axis inputs — column label attributes are re-applied before rendering, since the PKNCA processing pipeline strips them (#1336)
 
 ## Dependency changes
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -133,8 +133,7 @@
 
 ### TLG
 * Fixed one-sided whiskers (Upper/Lower) collapsing to zero height in mean plots (`pkcg03`) (#1316)
-* Custom titles now apply immediately to combined graphs (`pkcg02`); graphs are rendered through dedicated `plotly` outputs so option edits redraw the plot in place instead of only after switching figures (#1336)
-* Restored the `!COLUMN` label-reference syntax in graph/listing title, subtitle, footnote, and axis inputs — column label attributes are re-applied before rendering, since the PKNCA processing pipeline strips them (#1336)
+* Fixed TLG graph title/label handling: custom titles now apply immediately to combined graphs (`pkcg02`) via dedicated `plotly` outputs, and the `!COLUMN` label-reference syntax works again across title, subtitle, footnote, and axis inputs (column labels are re-applied before rendering, since the PKNCA pipeline strips them) (#1336)
 
 ## Dependency changes
 
--- a/inst/shiny/modules/tab_tlg/tlg_module.R
+++ b/inst/shiny/modules/tab_tlg/tlg_module.R
@@ -62,9 +62,8 @@ render_graph_outputs <- function(output, session, current_page_items) {
           req(my_i <= length(items))
           item <- items[[my_i]]
           req(!is.character(item))
-          # g_pkcg() returns plotly widgets by default (plotly = TRUE); the
-          # ggplot branch is the fallback for plotly = FALSE callers.
-          if (inherits(item, c("gg", "ggplot"))) plotly::ggplotly(item) else item
+          # implemented graph functions (g_pkcg*) always return plotly widgets
+          item
         })
       })
       n_registered <<- n
--- a/inst/shiny/modules/tab_tlg/tlg_module.R
+++ b/inst/shiny/modules/tab_tlg/tlg_module.R
@@ -21,6 +21,57 @@ filter_tlg_excluded <- function(data) {
   }
 }
 
+#' Wire up per-plot plotly outputs for a graph TLG module.
+#'
+#' Renders each graph through its own `plotlyOutput`/`renderPlotly` pair rather
+#' than returning raw plotly widgets from a single `renderUI`. Raw widgets did
+#' not redraw in place when an option changed (e.g. a custom title on combined
+#' pkcg02 plots) until the figure was hidden and shown again (issue #1336);
+#' letting plotly own each output binding makes edits apply immediately.
+#'
+#' @param output             Module `output` object.
+#' @param session            Module `session` object.
+#' @param current_page_items Reactive returning the items shown on the current
+#'                           page (plotly widgets, ggplots, or character errors).
+#' @noRd
+render_graph_outputs <- function(output, session, current_page_items) {
+  output$tlg_output <- renderUI({
+    items <- current_page_items()
+    tagList(purrr::imap(items, function(item, i) {
+      if (is.character(item)) {
+        tags$pre(item)
+      } else {
+        # preserve the height baked into the plotly object (set via ggplotly)
+        height <- if (!is.null(item$height)) paste0(item$height, "px") else "500px"
+        plotly::plotlyOutput(session$ns(paste0("plot_", i)), height = height)
+      }
+    }))
+  })
+
+  # Register a renderPlotly binding for each plot slot exactly once (tracked by
+  # high-water mark). The bodies read current_page_items() reactively, so option
+  # edits re-run them and plotly redraws the existing widget in place.
+  n_registered <- 0L
+  observe({
+    n <- length(current_page_items())
+    if (n > n_registered) {
+      for (i in seq.int(n_registered + 1L, n)) local({
+        my_i <- i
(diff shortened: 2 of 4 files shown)


inst/shiny/modules/tab_tlg/tlg_module.R
```diff
@@ -202,6 +201,50 @@ tlg_module_server <- function(id, data, type, render_list, options = NULL) {
...
+              items <- current_page_items()
+              req(my_i <= length(items))
+              item <- items[[my_i]]
+              req(!is.character(item))
+              if (inherits(item, c("gg", "ggplot"))) plotly::ggplotly(item) else item
+            })
+          })
+          n_registered <<- n
+        }
+      })
```
Shaakon35: **[Warning] Missing test for the primary fix.**

The core behavior this PR fixes — per-slot `plotly` registration so title edits redraw in place — has no test. Added tests only cover label restoration and page slicing.

**Resolution:** Add a `testServer` assertion that `output$plot_<i>` bindings are registered for `type == "graph"` (one binding per plot slot after the `observe` runs), so the registration loop is guarded directly.

inst/shiny/modules/tab_tlg/tlg_module.R
```diff
@@ -202,6 +201,50 @@ tlg_module_server <- function(id, data, type, render_list, options = NULL) {
...
+        n <- length(current_page_items())
+        if (n > n_registered) {
+          for (i in seq.int(n_registered + 1L, n)) local({
+            my_i <- i
+            output[[paste0("plot_", my_i)]] <- plotly::renderPlotly({
+              items <- current_page_items()
+              req(my_i <= length(items))
+              item <- items[[my_i]]
+              req(!is.character(item))
+              if (inherits(item, c("gg", "ggplot"))) plotly::ggplotly(item) else item
```
Shaakon35: **[Warning] Dead / unreachable branch.**

`g_pkcg*` functions already return plotly widgets, so the `gg`/`ggplot` path is never reached and is untested.

**Resolution:** Remove the conditional and return `item` directly (matches actual behavior, reduces complexity). If a raw ggplot can genuinely arrive here, keep the branch but add a comment explaining when, plus a test exercising the conversion.

inst/shiny/modules/tab_tlg/tlg_module.R
```diff
@@ -116,14 +116,8 @@ tlg_module_ui <- function(id, type, options) {
 #' @param render_list function that renders the list of entries, actual implementation of the TLG
 #' @param options     list of options to customize input parameters
 #'
-tlg_module_server <- function(id, data, type, render_list, options = NULL) {
+tlg_module_server <- function(id, data, type, render_list, options = NULL) { # nolint: cyclocomp_linter
```
Shaakon35: **[Warning] Complexity silenced rather than reduced.**

The graph-rendering block (`renderUI` + the high-water-mark `observe`) was added inline and `cyclocomp_linter` suppressed.

**Resolution:** Extract the graph output setup into a helper (e.g. `render_graph_outputs(output, session, current_page_items)`) and call it in the `type == "graph"` branch. This removes the need for the `nolint` and keeps the module readable.

Shaakon35 (approved):
Appart than that, it works well so I approve :)
please change the default number of plots **from 1 to all** please, that would help a lot for the waiting time

NEWS.md
```diff
@@ -133,6 +133,8 @@
 
 ### TLG
 * Fixed one-sided whiskers (Upper/Lower) collapsing to zero height in mean plots (`pkcg03`) (#1316)
+* Custom titles now apply immediately to combined graphs (`pkcg02`); graphs are rendered through dedicated `plotly` outputs so option edits redraw the plot in place instead of only after switching figures (#1336)
+* Restored the `!COLUMN` label-reference syntax in graph/listing title, subtitle, footnote, and axis inputs — column label attributes are re-applied before rendering, since the PKNCA processing pipeline strips them (#1336)
```
Gero1999: I would put only one message better

Gero1999 (approved):
I think this may do the trick  💪🏽 

It is true that I am not sure how fragile can it get, but I feel that once we change the default display to 1 it might get to be better. Otherwise the waiting time is too long

inst/shiny/modules/tab_tlg/tlg_module.R
Gero1999: For shiny tests we don't need full coverage, only for files under `/R`. Feel free to ignore. No need to overcomplicate

inst/shiny/modules/tab_tlg/tlg_module.R
Gero1999: @h5hoang it is fine to add the nolint for cyclomatic complexity if you feel "simplifying" or "abstracting" the function will actually make it less understandable.

It should not be a standard practice because in most cases is not true, but dw about this one if you don't want, I think is ok :)

h5hoang:
thanks for the reviews guys @Shaakon35 @Gero1999! 😄 before I make the final merge, I just wanted to double check with the current rounds of changes I made for the nitpicks

- I changed the default from "All" to 1 per page so the graphs tab doesn't hang on that initial render anymore, users can still bump it up to All/2/4/etc. whenever they want but default

- I pulled the graph output setup (the `renderUI` + high-water-mark `observe`) out into its own `render_graph_outputs()` helper and dropped the `cyclocomp_linter` nolint, lint passes clean now without needing to silence it

- for the gg/ggplot branch, I traced it through and the implemented graphs (`g_pkcg*`) always return plotly widgets (`plotly = TRUE` is never overridden, and `tlg.yaml` doesn't expose that option), so that branch was never actually reached. I just removed it and returned the item directly now

- for news bullet points, I merged them into one 

on top of that I merged latest main in and bumped the version to 0.1.0.9185 to clear the conflict

lmk if theres anything else or any other nitpicks, otherwise I'll go ahead and merge, thanks again!! 🙏

Gero1999:
LATM @h5hoang, ready to go! :)

PR #1396, opened 2026-07-15
Validate manual half-life calculation using sample dataset

PavanLomati:
Hi @Gero1999, 

Here is the sample dataset for manual half-life calculation


Gero1999:
I already checked the dataset, we can close the PR. Fix was done in https://github.com/pharmaverse/aNCA/issues/1378

PR #1401, opened 2026-07-17
fix: Slope selector (HL-plots) not appearing with error `Error in mutate: Can't transform a data frame with NA or "" names` due to underscore ("_" ) in STUDYID

### LINKED ISSUE #1378: Bug(s): 1) pak installation issue, 2) Half life plots not appearing in Slope Selector when dataset key variables contain "_" (can't transform a dataset with NA or "" in names)

Gero1999:
## Issue

Closes #1378

## Description

The Slope Selector crashed with `Error in mutate: Can't transform a data frame with NA or "" names` whenever a grouping value contained an underscore (e.g. `STUDYID = "same_or_similar"`, `PARAM = "Drug A"`).

Root cause: half-life plots are keyed by names of the form `key1=val1_key2=val2_...` (built in `get_halflife_plots()`). `parse_plot_names_to_df()` reversed this by splitting on **every** underscore, so underscores inside a value were treated as segment separators. This produced fragments with no `key=`, which became `NA`/`""` column names and misaligned rows under `rbind()`, aborting `arrange_plots_by_groups()` (and therefore the plot rendering) in the NCA > Slope Selector tab.

Fix: split names only at real `key=` boundaries (`_(?=[A-Za-z.][A-Za-z0-9.]*=)`, PCRE) and combine rows with `bind_rows()` so columns align by name. Values may now safely contain underscores or spaces. Malformed segments are dropped rather than turned into empty column names, so the crash is structurally impossible.

## Definition of Done

- Slope Selector renders plots for datasets whose `STUDYID`/`PARAM`/other grouping values contain underscores.
- `parse_plot_names_to_df()` never returns `NA`/`""` column names.
- No regression for names without underscores in values.

## How to test

1. `devtools::load_all()` then `aNCA::run_app()`.
2. Upload a dataset with an underscore in a grouping value (e.g. `STUDYID = "same_or_similar"`), map it, and open **NCA -> Setup -> Slope Selector**.

For the dataset you can use this one:

```
# Generate data/Sample.csv exactly as shown

sample_df <- data.frame(
  STUDYID = c(
    rep("same_or_similar", 12),
    rep("divergent", 12)
  ),
  USUBJID = c(
    rep(S1, 12),
    rep(S2, 12)
  ),
  ATPTREF = 1,
  ROUTE = rep("intravascular bolus", 24),
  TMDD = rep("without TMDD", 24),
  Variability = rep("0.01 prop variance, 0 add variance", 24),
  AFRLT = rep(c(0, 0.5, 1, 2, 3, 4, 8, 12, 16, 24, 36, 48), 2),
  ARRLT = rep(c(0, 0.5, 1, 2, 3, 4, 8, 12, 16, 24, 36, 48), 2),
  AVAL = c(
    0.922, 0.715, 0.61, 0.405, 0.3, 0.329, 0.177, 0.091, 0.04, 0.014, 0, 0,
    1.185, 1.021, 0.71, 0.647, 0.516, 0.32, 0.109, 0.058, 0.023, 0.002, 0, 0
  ),
  DOSEA = rep(30, 24),
  DOSETRT = rep("mg", 24),
  PARAM = c(
    rep("Drug A", 12),
    rep("Drug B", 12)
  ),
  stringsAsFactors = FALSE
)

# Write CSV without row names/quotes (matches your sample style)
write.csv(sample_df, file = "../../Downloads/dataset_with_underscore_in_STUDYID.csv", row.names = FALSE)

```

3. Plots render (previously: `Can't transform a data frame with NA or "" names`).

Covered by unit tests in `tests/testthat/test-utils-slope_selector.R`:

```r
devtools::test(filter = "utils-slope_selector")
```

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

- **NEWS entry not yet included in this branch.** Suggested entry under a new `### Slope Selector` bug-fix subsection: *"Slope Selector no longer crashes with `Can't transform a data frame with NA or '' names` when a grouping value (e.g. `STUDYID`, `PARAM`) contains an underscore. Plot names are now parsed at `key=` boundaries instead of splitting on every underscore (#1378)."*
- **R was not available in the authoring environment**, so `devtools::test()` and `lintr::lint_package()` were **not** run here. The split regex was validated against the failing inputs using the same PCRE engine (both `same_or_similar` and `divergent` parse to identical key sets, so `bind_rows` aligns cleanly). All changed lines are <=100 chars. Please run tests + lint before merging.
- Version bumped `0.1.0.9184` -> `0.1.0.9185`.
- Click handling is unaffected -- it already uses structured plotly `customdata`, not the parsed name string.


Shaakon35 (approved):
It works, even if the data is too small to allow the mean plots:

maybe we woudl need a better error in the future:

Image

KOBANAK (approved):
I tested with another dataset with "_" in STUDYID and it works as expected. 
We no longer see the previous error in the slope selector section.

PR #1410, opened 2026-07-22
fixed: issue #1365

### LINKED ISSUE #1365: UX: show the max upload size in the file upload UI
🛠️🎉 **Hackathon issue**

## Description

The app sets a 30 MB upload limit (`R/run_app.R:56`, `options(shiny.maxRequestSize = 30 * 1024^2)`), but this limit is not communicated anywhere in the UI. Users only discover it by hitting a generic \"Maximum upload size exceeded\" error (see #657).

## Suggested approach

Add helper text near the `fileInput` in `inst/shiny/modules/tab_data/data_upload.R` (around line 19) that states the accepted file types **and** the maximum size, e.g. *\"Accepted: csv, rds, xpt · Max 30 MB\"*.

Bonus: source the limit from a single constant so the UI text and `run_app.R` stay in sync.

## Definition of Done

- [ ] Max upload size is visible in the upload UI
- [ ] Value matches the actual `shiny.maxRequestSize` setting

## Context

Good first issue for the [Hackathon Hub](https://github.com/pharmaverse/aNCA/issues/1357). Related to #657.

poonam-16:
## Issue

Closes #1365 

## Description

Added a helper text that states the maximum upload limit

Additionally, sourced the limit from a single constant so the UI text and run_app.R stay in sync

## Definition of Done

- [x] Max upload size is visible in the upload UI
- [x] Value matches the actual shiny.maxRequestSize setting

## Contributor checklist
- [ ] New logic is documented

### DIFF
--- a/R/run_app.R
+++ b/R/run_app.R
@@ -50,10 +50,12 @@
 #'   run_app(settings = "/path/to/settings.yaml", settings_version = "NCA draft")
 #' }
 #' @export
+
+max_upload_size_mb <- 30
 run_app <- function(datapath = NULL, settings = NULL,
                     settings_version = 1L, ...) {
   # Increase max upload size to 30 MB
-  options(shiny.maxRequestSize = 30 * 1024^2)
+  options(shiny.maxRequestSize = max_upload_size_mb * 1024^2)
   if (!is.null(datapath)) {
     stopifnot(
       "Data file does not exist" = file.exists(datapath),
--- a/inst/shiny/modules/tab_data/data_upload.R
+++ b/inst/shiny/modules/tab_data/data_upload.R
@@ -15,7 +15,10 @@ data_upload_ui <- function(id) {
     div(
       class = "upload-container",
       id = ns("upload_container"),
-      p("Upload your PK dataset and Settings file (optional)."),
+      p("Upload your PK dataset and Settings file (optional).",
+        tags$br(),
+        tags$small(style = "color: #6c757d; display: block; margin-top: 4px;",
+                   sprintf("Maximum upload size: %s MB", max_upload_size_mb))),
       fileInput(
         ns("data_upload"),
         width = "50%",


R/run_app.R
```diff
@@ -50,10 +50,12 @@
 #'   run_app(settings = "/path/to/settings.yaml", settings_version = "NCA draft")
 #' }
 #' @export
+
+max_upload_size_mb <- 30
```
Gero1999, line 54: issue: as you see this is cutting the roxygen docx (all that is before `@export`) which may produce a fail in geenerating the documentation. Instead, move this line right before:

``` 
options(shiny.maxRequestSize = max_upload_size_mb * 1024^2)
```

inst/shiny/modules/tab_data/data_upload.R
```diff
@@ -15,7 +15,10 @@ data_upload_ui <- function(id) {
     div(
       class = "upload-container",
       id = ns("upload_container"),
-      p("Upload your PK dataset and Settings file (optional)."),
+      p("Upload your PK dataset and Settings file (optional).",
+        tags$br(),
+        tags$small(style = "color: #6c757d; display: block; margin-top: 4px;",
+                   sprintf("Maximum upload size: %s MB", max_upload_size_mb))),
```
Gero1999, line 21: Nice idea! Having a single source of truth is always good. However, for an installed pacakge the `max_upload_size_mb` variable won't be available in this submodule. 

I would recommend that we instead search for the value in the settings `getOption("shiny.maxRequestSize"`, but in case for any reason run_app.R and the setting was not set (i.e, someone decided to run the App manually without using our function, `shiny::runApp(...)`) then we tell teh code to assume we are using the default that shiny has (5 MB). In code this will translate to this:

```suggestion
                   sprintf("Maximum upload size: %s MB",
                           getOption("shiny.maxRequestSize", 5 * 1024^2) / 1024^2))),
```

Gero1999:
Thanks for the contribution @poonam-16! I like the idea of a single source of truth, very witty! 

Unfortunately, this does not seem t work on an installed package. I added some comments

Gero1999 (changes requested):

Gero1999:
Thanks again for the contribution. I moved this work to a maintained `pharmaverse/aNCA` branch so we can continue iterating on the review fixes directly while preserving the original contribution as a cherry-picked commit.

Replacement PR: #1498

I am closing this PR to avoid having two open PRs for the same issue.

PR #1411, opened 2026-07-24
feat: add option to drop end-interval concentration for regular parameters

### LINKED ISSUE #1414: Enhancement: Drop end-of-interval concentration belonging to the next dose
### Description

For multiple-dose data, a concentration measured exactly at the end of a dosing interval belongs to the **next** dose (e.g. an imputed or measured C0 of the next profile). Because aNCA derives each non-last interval's `end` as the next dose time and intervals are closed at that boundary, this next-dose point leaks into the **previous** dose profile and distorts its results (Cmax, Tmax, Clast, AUClast, λz).

We need an option to drop that end-boundary concentration for regular (non-partial) parameter calculations, so the previous dose profile is not affected by the next dose's starting point. Genuine troughs must be unaffected, and the last dose profile per subject/group must keep its real terminal sample.

### Definition of Done

- [ ] A **"Drop End Concentration"** switch in *NCA Setup → Data Imputation* (toggles the behaviour)
- [ ] When enabled, the concentration at `time == end` is dropped for **non-last main intervals** per concentration group
- [ ] Partial/interval parameters (`type_interval == "manual"`) are never affected
- [ ] The **last** dose profile per group keeps its genuine terminal sample (not dropped)
- [ ] Method is a no-op when no point sits exactly at the interval end (genuine troughs preserved)
- [ ] Setting persisted/restored via settings YAML and templates
- [ ] Covered by unit tests

Gero1999:
## Issue

Closes #1414 

## Description

Adds a global **"Drop End Concentration"** switch in *NCA Setup → Data Imputation*. When enabled, for regular (non-interval/partial) parameter calculations it drops the concentration measured exactly at the end of each **main** interval (`time == end`) before NCA runs.

Useful for multiple-dose profiles where the concentration at the interval end really belongs to the next dose (e.g. an imputed C0). Including that boundary point can distort Cmax, Tmax, and AUC. Genuine troughs are unaffected because the method is a no-op when no point sits exactly at the interval end.

Implemented via a self-contained `PKNCA_impute_method_end_conc_drop()` mirroring humanpred/pknca#572, with a TODO to remove it once that method ships in a released PKNCA.

## Definition of Done

- [x] Global "Drop End Concentration" checkbox (checked = drop)
- [x] Applied only to main intervals; partial/interval params untouched
- [x] Self-contained impute method with removal TODO
- [x] Setting persisted/restored via settings YAML and templates
- [x] Covered by unit tests
- [x] Documented in the manual and NEWS

## How to test
Added a tiny reproducer dataset for manual testing: `tests/testthat/data/test-end-conc-leak-ADNCA.csv` (1 subject, 2 IV-bolus dose profiles, 14 rows).

**How to use in the app:**
1. Upload the CSV (default mappings apply; `DOSNOP → ATPTREF`).
2. Dose 2's C0 is **1050 ng/mL at AFRLT = 24 h**, and profile 1's interval end is derived as the next dose time (24 h) — so that point sits exactly on profile 1's boundary.
3. Run NCA with **Drop End Concentration OFF**: profile 1's Clast/Tlast/AUClast/λz are distorted by the 1050 spike.
4. Toggle it **ON**: profile 1's boundary point is dropped and results normalise, while profile 2 (the last profile) keeps its genuine terminal sample.

## Notes to reviewer

- Self-contained impute method is a temporary copy of humanpred/pknca#572 (TODO to remove once released).
- R unavailable in authoring env: man/*.Rd and NAMESPACE edited by hand to match roxygen; please run devtools::document() (should be no diff). Tests rely on CI.
- Replace the #TODO in NEWS.md with the issue number once created.


inst/WORDLIST
```diff
@@ -7,9 +7,7 @@ AEFRLT
 AFRLT
 ARRLT
 ATPTREF
-AUCIFOD
```
Shaakon35, line 10: **Medium — unrelated WORDLIST churn (scope creep).** This PR removes 10 words (`AUCIFOD`, `AUCLSTD`, `CMAXD`, `aucint`, `multidose`, `nav`, `pcspec`, `pptestcd`, `rds`, `scrollable`) and adds only `humanpred`. Only `humanpred` relates to this feature.

I checked whether the removals break the spellcheck CI job: **they don't today** — every removed word still appears in `man/`/`vignettes/`, but all occurrences are inside inline code (backticks), `\code{}`, or fenced code blocks, which the `spelling` package skips.

The concern is maintenance risk: this narrows the dictionary for reasons unrelated to the feature. If anyone later references those terms in prose, spellcheck fails unexpectedly. **Suggest reverting the deletions and keeping only the `humanpred` addition** to keep this PR focused.

tests/testthat/data/test-end-conc-leak-ADNCA.csv
```diff
@@ -0,0 +1,15 @@
+STUDYID,USUBJID,PARAM,PCSPEC,AVAL,AVALU,DOSNOP,ROUTE,ADOSEDUR,DOSEDURU,NFRLT,AFRLT,ARRLT,AEFRLT,NRRLT,RRLTU,DOSEA,VOLUME,VOLUMEU,DOSEU,DOSETRT,GROUP,SUBJID,SPECIES,GENDER,STRAIN,ANIMAL_STATUS,ASSAY_LLQ,ASSAY_LLQ_UNIT,tr_route_of_admin,tr_form_of_route,MODALITY,PROJECT,THEME,FORMULATION,Concentration,ADDITIVE,stt_stud_study_type,METABFL,WTBL,WTBLU
```
Shaakon35, line 1: **Medium — headline scenario has no automated regression test.** This CSV is the flagship multi-dose "next-dose C0 leak" reproducer, but it's referenced only for manual testing — no automated test loads it. Current coverage is good at the unit level (the impute function + interval-column tagging in `update_main_intervals()` + a synthetic `pk.nca` tmax check), but the end-to-end path (`PKNCA_update_data_object` → NCA with `drop_end_conc = TRUE` distorting then normalizing Clast/Tlast/AUClast) isn't guarded against regression.

Consider wiring this CSV into an integration test that runs NCA with the switch OFF then ON and asserts the boundary point's effect on profile 1's terminal params. That would make at least one test fail if the feature regressed.

inst/WORDLIST
Gero1999: this one is not relevant

tests/testthat/data/test-end-conc-leak-ADNCA.csv
Gero1999: this one is not relevant

Gero1999:
@Shaakon35 both comments do not need to be addressed, should be fine

PR #1423, opened 2026-08-10
Fix Mapping Next button after duplicate modal cancel

### LINKED ISSUE #1420: Bug: Next button stays disabled after cancelling the duplicates modal (and mapping needs a loading spinner)
### Description

Follow-up to the class of problems fixed in #1403 (Next button stuck disabled after an incomplete-mapping error).

On the **Data** tab, when the mapped data contains blocking time-duplicate rows, a "Duplicate Rows Detected" modal is shown. If the user clicks **Cancel** on that modal, they are left with a **greyed-out (disabled) Next button** and no result — they cannot proceed and effectively have to restart the app. Users expect a result (or at least a re-enabled button) after cancelling.

Additionally, the mapping step can take a noticeable amount of time, but there is **no loading spinner** while mapping is being applied, so the app appears frozen/unresponsive.

### To reproduce

1. `devtools::load_all()` then `aNCA::run_app()`.
2. Upload a dataset that contains time duplicates (duplicate rows based on TIME, STUDYID, PCSPEC, DOSETRT, USUBJID, PARAM).
3. On the **Mapping** step, click **Next** — the "Duplicate Rows Detected" modal appears.
4. Click **Cancel** on the modal.
5. Observe: the **Next** button remains greyed out / disabled and no result is produced.

### Expected behaviour

- After cancelling the duplicates modal, the **Next** button is re-enabled so the user can correct the data / mapping and try again (mirroring the fix in #1403).
- A **loading spinner** is shown while the mapping is being applied, so the user gets clear feedback that work is in progress.

### Notes (relevant code)

- Next button enable/disable logic: `inst/shiny/modules/tab_data.R` (`.setup_step_navigation`). The re-enable observer currently depends on `data_step()` and `trigger_mapping_submit()`; cancelling the modal changes neither, so the button stays disabled.
- Duplicates modal / Cancel path: `inst/shiny/modules/tab_data/data_mapping.R` (the `modalButton("Cancel")` in the `df_duplicates()` modal has no handler to re-enable navigation).

Gero1999:
## Issue

Closes #1420

## Description

The Data tab disabled the **Next** button when a mapping submission started, but the duplicate-resolution modal's **Cancel** button was a plain `modalButton()`. Closing that modal did not notify the server, so the parent navigation state never learned that the mapping workflow had ended.

This PR centralizes the Mapping-step button state behind a `mapping_busy` reactive value and makes the duplicate modal Cancel action a server-side button. The mapping module now reports terminal mapping outcomes back to the parent module through callbacks, so the spinner/modal can be dismissed and **Next** can be re-enabled consistently after mapping errors, duplicate detection, successful processing, and duplicate-modal cancellation.

The duplicate-detection path now also defers publishing duplicate data until after the loading-modal cleanup has flushed. That keeps the loading modal dismissal and the duplicate modal display as separate client updates, avoiding the same Bootstrap show/hide timing race that affected the no-duplicates path.

## Definition of Done

- After cancelling the duplicate rows modal, the Data tab **Next** button is re-enabled.
- Mapping work shows a loading spinner while processing is in progress.
- The **Next** button disabled state is controlled by mapping busy state rather than a fixed timeout.
- Normal upload -> mapping -> filtering navigation is preserved.
- Duplicate-row detection reliably shows the duplicate modal after the loading modal is cleared.

## How to test

1. Run `devtools::load_all()` and then `aNCA::run_app()`.
2. Default/no-duplicates path: from the default data, click **Next** to Mapping and **Next** again. Confirm the loading spinner closes and the app advances to Filtering.
3. Duplicate path: upload a dataset containing time duplicates based on `TIME`, `STUDYID`, `PCSPEC`, `DOSETRT`, `USUBJID`, and `PARAM`.
4. On the Mapping step, click **Next**.
5. Confirm the mapping loading spinner appears while mapping is processed, then the "Duplicate Rows Detected" modal appears.
6. Click **Cancel**.
7. Confirm the modal closes and **Next** is enabled again so the user can correct the mapping/data and retry.
8. Repeat with **Keep Selected** and confirm resolved duplicates still allow the app to advance to Filtering.

## Contributor checklist

- [x] New logic is documented
- [x] R script works with the new implementation — N/A, Shiny navigation only
- [x] Settings upload works with the new implementation — N/A, no settings format changes
- [x] If any `.scss` change was done, run `data-raw/compile_css.R` — N/A, no SCSS changes
- [x] If a package dependency was added/changed, run `data-raw/test_suggests_hidden.R` — N/A, no dependency changes

## Notes to reviewer

This is Shiny navigation behavior, so the most important verification is interactive: reproduce the default no-duplicates path and the duplicate-modal cancel path and confirm the loading modal is not left behind and **Next** is no longer left disabled.


Gero1999:
wrong: "Processing data mapping..." just by using the default user path (next next) witht he default dummy data

Gero1999:
this would need a review @Shaakon35 @h5hoang to make sure it also worked for you and there is no way around I didn't not think of

Shaakon35:
Potential regression on the duplicate path: in the `time_duplicate_error` handler, `on_mapping_complete()` removes the loading modal and then `df_duplicates(e$duplicate_data)` invalidates the observer that calls `showModal()` for the duplicate dialog. Both modal messages can therefore be sent in the same Shiny flush. This recreates the Bootstrap show/hide timing race that the delayed submit fixes on the no-duplicates path, and can cause the duplicate dialog to be dropped or leave the modal state inconsistent.

Could we defer publication of the duplicate data until after the loading-modal removal has flushed? For example:

```r
on_mapping_complete()
session$onFlushed(
  function() df_duplicates(e$duplicate_data),
  once = TRUE
)
```

That makes the duplicate dialog a subsequent client update. Please also add an interactive regression test for mapping → duplicate detected → duplicate modal visible → Cancel re-enables Next; the current test exercises only the no-duplicates path.

Shaakon35 (approved):
Works perfect. Just check the AI review above.

PR #1425, opened 2026-08-10
Lock faceted Exploration y-axis ranges

### LINKED ISSUE #1424: Enhancement: lock y-axis range across faceted exploration plots
### Description

Users reported that faceted PK concentration-time plots in the Exploration tab are hard to compare because each facet can use a different y-axis range. This affects both linear and semi-log y-axis views.

Add a user-facing Exploration plot control, such as **Use same Y-axis range**, that lets users lock the y-axis range across facets when panels share the same y units. The setting should be available only when faceting is active and should be preserved by saved/exported/generated exploration plots.

Implementation guidance:
- Put the core behavior in the shared plotting helper, not only in the UI layer.
- Add a plotting parameter such as `lock_y_axis = FALSE`, then pass it through the exploration wrappers and Shiny module calls.
- For ggplot faceting, preserve the current free-scale behavior when disabled and use shared y scaling when enabled while keeping x-axis behavior as currently intended.
- Support both individual and mean exploration plots.
- For semi-log plots, avoid hard-coded y-axis breaks that can hide tick labels outside a fixed range.
- Keep API defaults backwards-compatible unless there is a clear app-level reason to change the Shiny default.

### Definition of Done

- [ ] Exploration plot controls include a user-facing **Use same Y-axis range** option when faceting is active.
- [ ] Faceted individual exploration plots can share one y-axis range when the option is enabled.
- [ ] Faceted mean exploration plots can share one y-axis range when the option is enabled.
- [ ] Existing free-scale behavior remains available when the option is disabled.
- [ ] Saved/exported/generated exploration plots preserve the selected setting.
- [ ] Linear and semi-log y-axis views are supported without hiding semi-log tick labels through hard-coded breaks.
- [ ] Focused tests cover locked and free y-scale faceted behavior.

Gero1999:
## Issue

Closes #1424

## Description

Faceted PK concentration-time plots in the Exploration tab could use independent y-axis ranges per facet, which made panels hard to compare. This affected both individual and mean plots, including semi-log views.

This PR adds a faceting-only **Use same Y-axis range** control to the Exploration plot sidebars. The app defaults the option to enabled when faceting is active, while the exported plotting APIs remain backwards-compatible with `lock_y_axis = FALSE` by default.

The core behavior is implemented in `g_lineplot()` with a new `lock_y_axis` parameter. Existing behavior still uses `facet_wrap(..., scales = "free")`; locked y-axis behavior uses `scales = "free_x"` so facets share y-axis scaling while x-axis behavior remains free. The setting is passed through individual and mean Exploration wrappers, Shiny calls, and generated plot code for saved/exported plots.

The semi-log y scale also no longer uses a hard-coded break list, allowing ggplot to choose visible tick marks for the actual plotted range.

## Edge case handling: y-axis units across facets

Locking a shared y-axis is only meaningful when the compared values share a unit, so units (`AVALU`) are harmonised in two stages before locking: **within each panel** (always, since a single panel cannot show two units on one axis) and **across panels** (only when `lock_y_axis = TRUE`). Compatible units are converted with `get_conversion_factor()` (the `units` package); the target unit is the most frequent `AVALU`, with an alphabetical tie-break. SD/CI family columns are scaled alongside the mean. Dimensionally-incompatible units (e.g. mass vs molar concentration) cannot be converted, so they are shown per facet and a notification suggests grouping variables to split them.

| # | Scenario | App behaviour | Expected behaviour (scientific) | Match |
|---|----------|---------------|---------------------------------|-------|
| 1 | Single facet / all facets already share one unit | Values unchanged; shared y-axis applied when locked | Correct — no conversion needed, axis directly comparable | ✅ |
| 2 | **Compatible** units mixed **within one panel** (e.g. ng/mL + µg/mL for the same analyte) | Converged to the panel's most-frequent `AVALU` (alpha tie-break); no warning | Must resolve to a single unit; conversion is exact and lossless | ✅ |
| 3 | **Incompatible** units **within one panel** (e.g. ng/mL + nmol/L, mass vs molar) | Values kept as-is, both units shown comma-separated in the strip label; warning suggests adding `PARAM`/`PCSPEC` to *facet by* | Should NOT be forced onto one axis; separating them per facet is correct | ✅ |
| 4 | **Convertible** units **across facets**, `lock_y_axis = TRUE` (facet A ng/mL, facet B µg/mL) | Converted to a shared target unit; single y-axis label; SD/CI scaled too | Correct — after conversion the shared axis is a valid comparison | ✅ |
| 5 | **Non-convertible** units **across facets**, `lock_y_axis = TRUE` (ng/mL vs nmol/L) | Convertible facets share the target; non-convertible facets keep their unit annotated in the strip; warning "units could not be aligned" | Correct not to force a shared numeric axis across different physical dimensions | ✅ |
| 6 | Different units across facets, `lock_y_axis = FALSE` | No cross-facet conversion; free scales per facet (within-panel convergence still applies) | Correct — each facet is self-consistent; cross-facet comparison is not implied | ✅ |
| 7 | SD / CI error bars present during a unit conversion | `SD_min`, `SD_max`, `CI_lower`, `CI_upper` scaled by the same factor as the mean | Required — variability bands must track the mean's unit | ✅ |
| 8 | Dose-normalised "both" mode (`linetype_by` set, raw + normalised on one plot) | Unit resolution skipped entirely; units left untouched | Units differ by design (e.g. ng/mL vs ng/mL/mg); leaving them avoids corrupting values | ✅ |
| 9 | Semi-log (`ylog_scale = TRUE`) with a locked, magnitude-spanning shared axis | Hard-coded breaks removed; ggplot picks ticks for the actual range | Correct for tick visibility; facets with small signals may appear compressed on a shared log axis (expected trade-off of locking) | ✅ |

## Definition of Done

- [x] Exploration plot controls include a user-facing **Use same Y-axis range** option when faceting is active.
- [x] Faceted individual exploration plots can share one y-axis range when the option is enabled.
- [x] Faceted mean exploration plots can share one y-axis range when the option is enabled.
- [x] Existing free-scale behavior remains available when the option is disabled.
- [x] Saved/exported/generated exploration plots preserve the selected setting.
- [x] Linear and semi-log y-axis views are supported without hiding semi-log tick labels through hard-coded breaks.
- [x] Compatible y-axis units are harmonised within and across facets; incompatible units are surfaced per facet with a notification.
- [x] Focused tests cover locked and free y-scale faceted behavior.

## How to test

1. Run `devtools::load_all()` and then `aNCA::run_app()`.
2. Upload or restore data and open the Exploration tab.
3. In **Individual Plots**, select a faceting variable.
4. Confirm **Use same Y-axis range** appears and is checked by default.
5. Confirm facets share the same y-axis range while the option is checked.
6. Uncheck **Use same Y-axis range** and confirm the previous free-scale behavior is restored.
7. Repeat the same check in **Mean Plots**.
8. Enable **Semi-log scale** and confirm y-axis tick labels remain appropriate for the plotted range.
9. Use **Show Code** or **Add to Exports** and confirm generated plot code includes the selected `lock_y_axis` value.
10. Facet across groups with differing but **compatible** units (e.g. ng/mL and µg/mL) with the option checked and confirm values are rescaled to a single labelled unit.
11. Facet across groups with **incompatible** units (e.g. ng/mL and nmol/L) and confirm a warning notification appears and the affected facet strips show their unit.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

R is not installed in the authoring container, so `devtools::document()`, `devtools::test()`, and `lintr::lint_package()` were not run here. `git diff --check` passed.

The Shiny app defaults the new faceted control to enabled, but the package-level plotting APIs default `lock_y_axis = FALSE` to preserve existing script behavior.


Belubbit (approved):
Looks good, thanks!

PR #1428, opened 2026-08-11
Enhancement: Allow export/download of all generated TLGs

### LINKED ISSUE #1344: Enhancement: Allow export/download of all generated TLGs
### Description

After the user submits the TLG order and the graphs/listings/tables are rendered, there is no way to download all generated TLGs at once. Users should be able to export all rendered TLGs (or a selection) as files (e.g., PNG/PDF for graphs, CSV/XLSX for tables and listings).

This is different from #54 (PDF export with Citril/Chevron pagination) — this is about a simpler bulk download of the rendered outputs in their current form.

### Definition of Done

- [ ] Download button available in the TLG section after TLGs are rendered
- [ ] Graphs exported as PNG or PDF files
- [ ] Tables exported as CSV or XLSX files
- [ ] Listings exported as CSV or XLSX files
- [ ] All exported files bundled in a ZIP archive with a clear folder structure (e.g., `Graphs/`, `Tables/`, `Listings/`)
- [ ] File names follow a consistent naming convention (e.g., `pkcg01_LIN_ANALYTE.png`)

### Related

- #54 — PDF export with Citril/Chevron (more advanced, with pagination and footnotes)
- #78 — PPT output improvements

h5hoang:
## Issue

Closes #1344

## Description

Rendered TLGs can now be exported. They appear as a branch in the existing **Export as ZIP** dialog, alongside the other results, once a TLG order has been submitted:

```
TLGs/
  Graphs/pdf/pkcg01_lin.pdf          one multi-page document per output
  Graphs/png/pkcg01_lin/S1.png       one image per plot, if PNG is chosen
  Tables/xlsx/pkct01.xlsx            one workbook, a sheet per split
  Tables/csv/pkct01/PARAM_DrugA.csv
  Listings/xlsx/pkcl01.xlsx
  manifest.csv                       new file I introduced as a tracker
```

**Getting the outputs out of the modules:**

- `tlg_module_server` returns its `tlg_list` reactive instead of nothing, and `tab_tlg_server` keeps those in a per-session registry, so the export writes what is actually on screen — sidebar option edits included — rather than re-running the catalog with defaults
- the registry is intersected with the current order, so a TLG removed and re-submitted does not linger in the archive
- the panel outputs and each module's options widget opt out of `suspendWhenHidden`; without that, a tab the user never opened contributed nothing to the export

**Formats (`inst/shiny/functions/zip-utils.R`):**

- rendered TLGs get their own control in the dialog rather than sharing the CDISC dataset formats, 
- defaults are PDF for graphs and XLSX for tables and listings. 
- graphs: PDF as one multi-page document per output, HTML similar to PDF but with interactivity, or PNG one file per plot. Rendered from the ggplot behind each widget
- tables and listings: XLSX collapses a split output into one workbook with a sheet per split; CSV keeps one file per split. Listings are reduced to their displayed columns, and grouped ("Compare in columns") tables get their two-level header flattened to `F - Geometric Mean`; PDF output with fixed layout/pagination.
- a TLG that errors or is still gated is recorded in `manifest.csv` and skipped, 

**Everything else:**

- `inst/shiny/functions/tlg_export.R` handles file naming, the folder layout and the manifest
- `save_dispatch()` gained PDF, XLSX and plotly→PNG rather than growing a second writer alongside it
- `.make_zip_filename()` takes a suffix so both exports share the project/study naming
- a TLG that errors or is still gated is recorded in `manifest.csv` and skipped, so one broken output does not cost the user the other thirty

## Definition of Done

- [x] Download button available in the TLG section after TLGs are rendered
- [x] Graphs exported as PNG
- [x] Tables exported as CSV/XLSX
- [x] Listings exported as CSV/XLSX
- [x] All exported files bundled in a ZIP with `Graphs/`, `Tables/`, `Listings/`
- [x] File names follow a consistent convention (`pkcg01_lin_DrugA.png`)

## How to test

1. Upload data, map, run NCA, go to TLG -> Submit Order Details
2. Open Export as ZIP (sidebar header) → tick TLGs in the tree → choose formats → Export 
3. Unzip: check the three folders and `manifest.csv`, that PNGs are not blank, and that the XLSX files open
4. Change a sidebar option (a title, or stratification) and re-download (the export should reflect it)
5. Remove a TLG from the order, re-submit, re-download (it should be gone from the zip)
6. Enable a TLG that errors on your data (the download should still complete, with the reason recorded in the manifest)

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

- there was an issue with graphs not downloading when you click export, and that you needed to open the graphs tab and click on every graph tab for it to be exported, but it should be fixed. but could you guys double check and make sure that graphs are part of the export?
- manifest.csv is an artifact I introduced as part of this PR, but it lists out every rendered output with its file path and a status (ok, skipped, empty, error) just so that if a TLG failed to be produced/exported, it would tell me rather than I have to dig through the folders and make sure all are present
- some graphs have overlapping x-axis labels if PK profile is dense, created an issue since that is out of scope for this PR (https://github.com/pharmaverse/aNCA/issues/1441)


inst/shiny/functions/tlg_export.R
```diff
@@ -0,0 +1,600 @@
...
+  }
+  if (length(drawable) == 0) {
+    return(.tlg_manifest_row(g_id, entry, NA_character_, "skipped",
+                             "none of the plots could be rendered to PDF"))
+  }
+
+  tryCatch({
+    grDevices::pdf(file.path(target_dir, rel), width = 10, height = 6, onefile = TRUE)
+    on.exit(grDevices::dev.off(), add = TRUE)
+    for (p in drawable) print(p)
```
Shaakon35, line 319: I am not sure it's this print, but it shows the name of the plot here:

Image

Shaakon35 (commented):
Small other thing, the export as zip for the TLGs is not active until we run the NCA.
We would need to make it active before (like it's the case of the exploration plot: as soon as exploration plot is open, we can download the plots).

Shaakon35 (approved):
I will approve because it's a good first draft solution.

Then every company could say that we need this or this format, but this we can see later.

For example, at Roche we would need one Pdf by Parameter. While currently we produce one pdf (l_pkconc for example) for all parameters

Shaakon35:
I found two export edge cases worth addressing:

1. **TLGs are unavailable in the ZIP dialog before NCA has run.** `.available_tree_items()` only adds the `TLGs` branch inside `if (nca_available)`. Concentration TLGs can render from mapped ADNCA before NCA/ADPP exists, but in that workflow the export dialog shows no TLG nodes, so the rendered outputs cannot be selected. Could the TLG-branch construction move outside the NCA-only condition, while leaving NCA results/CDISC gated as they are?

2. **Long, similarly prefixed split labels can make the default XLSX export fail.** `.tlg_sheet_names()` truncates names to 31 characters, calls `make.unique()`, and truncates again. If two split labels only differ after character 31, `make.unique()` adds `_1` but the final truncation removes it, recreating duplicate Excel sheet names; `writexl::write_xlsx()` then rejects the workbook. Please preserve room for the uniqueness suffix (or deduplicate after the final length constraint) and add a regression test with two labels sharing their first 31 characters.

h5hoang:
thanks @Shaakon35 , I've just addressed these. 

basically TLGs can now be exported before running NCA, and the xlsx sheet names stay unique even when the labels share the same first 31 characters.

the extra plot name was actually coming from the app, not the print() in the PDF export. i've removed that heading for pkcg01/02 but kept the names for the exported files

I'll implement the one PDF per parameter in a separate issue down the line, unless you think it should be included in this PR. also could you check that the extra plot name from your screenshot is gone now and the title/subtitle still look right?

thanks again for the feedback! very much appreciated 🙏

inst/shiny/functions/tlg_export.R
```diff
@@ -0,0 +1,631 @@
...
+#'
+#' The suffix is reserved *inside* the cap instead: the stem is shortened by exactly as much
+#' as the suffix needs. Comparison is case-insensitive because Excel and common Windows/macOS
+#' file systems treat `DrugA` and `druga` as the same name.
+#'
+#' @param nm        Character vector of candidate names, already capped at `max_chars`.
+#' @param max_chars Hard limit the result must respect.
+#' @returns `nm`, made unique, with every element still at most `max_chars` long.
+#' @noRd
+.tlg_unique_within <- function(nm, max_chars) {
```
Gero1999, line 580: question: just checking, but is this really needed logistically? What actual example/case applies to it

Gero1999 (approved):
I reviewed and I confirm:
1. It saves TLGs even before NCA running
2. PDFs, html, PNGs look good

~My only point would be to later on expand Listings and Tables so they can also export eithere HTML or PDF. Something nicer than the Excel (which is the current only option) so users can later on include them in their reports.~

3. Lisstings and Tables look perfect as well as PDF :)) 

For the rest, LGTM, great job! 💪🏽

PR #1433, opened 2026-08-21
fix: capture automatic volume unit simplification in exports (#1190)

### LINKED ISSUE #1190: Bug: R script template missing volume unit simplification step

Gero1999:
## Issue

Closes #1190

## Description

Automatic volume unit simplification (e.g. `mg*L/mL` -> `mg`, applied in `tab_data.R` after `PKNCA_create_data_object()`) never reached `session$userData$units_table()`. As a result the exported settings YAML, ZIP export, and generated R script lost the simplified units unless the user manually opened the Units modal, so re-running the exported R script produced different units than the app.

This is a fresh reimplementation of the approach from the stale #1228 branch (438 commits behind `main`), rebuilt on top of current `main`. Several pieces the original branch introduced (import-side partial merge via `apply_unit_defaults`, export filtering) have since landed independently, so this PR is scoped to the three remaining gaps:

- **Seed the session units table on NCA run.** `res_nca` now populates `units_table()` from the already-simplified, data-derived units the first time NCA runs (using `isolate()` to avoid a read/write reactive loop), so automatic changes are captured even when the modal is never opened.
- **Value-based change detection.** A new exported helper `changed_units()` keeps only rows where `PPSTRESU` differs from `PPORRESU`, replacing the previous `default` flag that only tracked edits made through the modal (and thus missed automatic simplification). The `default` column is removed from the units table flow.
- **Decouple units from the debounced `settings()` reactive.** Units were part of `settings()`, which created a dependency chain that could stall session auto-replay. Export paths (settings download + ZIP) now read `session$userData$units_table()` directly and filter with `changed_units()`.

## Definition of Done

- Volume unit simplification applied in the app is reflected in the exported settings YAML, ZIP export, and generated R script.
- Units table change detection is value-based, not flag-based.
- Units no longer participate in the debounced `settings()` reactive.

## How to test

1. Upload data whose default volume units simplify (e.g. a parameter with `mg*L/mL`).
2. Run NCA **without** opening the Units modal.
3. Download settings (and/or the ZIP export) and confirm the simplified unit (e.g. `mg`) appears in the units section.
4. Run the generated R script and confirm it reproduces the same units as the app.
5. Import the exported settings back and confirm the units table is populated correctly.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

- **R was not available in the authoring environment**, so `devtools::document()`, `devtools::test()`, and `lintr::lint_package()` were not run. `man/changed_units.Rd` and the `NAMESPACE` export were added by hand; please run `devtools::document()` to confirm no drift, and run the test suite/lintr locally.
- New unit tests cover `changed_units()` (differs-only, volume-simplification capture, no-change, NA exclusion, NULL/empty). The reactive seed in `res_nca` is not unit-tested; worth a manual in-app sanity check of the read/write isolation.
- This supersedes #1228 - consider closing that PR in favor of this one.


Gero1999:
I can see that now for VSSO it works indeed, and the R script will include this custom unit

Issue #1436, opened 2026-08-21
Enhancement: Align ADNCA and ADPP summary exclusions and clarify the UI


Gero1999:
### Description

Unify and clarify how summary exclusions are represented and communicated across ADNCA and ADPP.

Both exclusion flags have the same high-level purpose: a value of `"Y"` identifies a record that must be excluded from summary calculations and applicable plots/TLG outputs. The flags must remain dataset-specific and use the following names:

- `PKSUMXF` for ADNCA concentration records
- `PPSUMXF` for ADPP parameter records

These flags must remain independent. Excluding an ADNCA record must not implicitly exclude the corresponding ADPP parameter record, and excluding an ADPP parameter record must not remove the corresponding ADNCA record.

Rename the setup tabs so their dataset scope is immediately clear:

- **General Exclusions** -> **ADNCA Exclusions**
- **Parameter Exclusions** -> **ADPP Exclusions**

Replace ambiguous user-facing references to "TLG exclusion" with "summary exclusion" where appropriate. The current behavior excludes flagged records from summary statistics and applicable mean plots, but retains them in individual listings.

Add short, professional tooltips to every **Row Colors** legend option. Suggested wording:

- **NCA calculation exclusion:** "Excludes the concentration record from the complete NCA calculation."
- **Summary exclusion:** "Keeps the record in the analysis dataset but excludes it from summary statistics and applicable plots."
- **NCA + summary exclusion:** "Excludes the record from both the NCA calculation and summary outputs."
- **ADPP summary exclusion:** "Keeps the parameter record in ADPP but excludes it from summary statistics and applicable plots."

Tooltips should be available by mouse hover and keyboard focus. The visible label and color swatch must remain so that exclusion status is not communicated by color alone.

#### Open UX question: individual listings

Individual listings should continue to include excluded records. Before adding new listing behavior, confirm whether users need a professional visual indication that a listed record was excluded from summary calculations/plots - for example, an exclusion-status column, marker, or footnote.

If users require this behavior, create a separate TLG enhancement issue to define and implement it consistently across listings. Its implementation is outside the scope of this issue.

### Definition of Done

- [ ] ADNCA summary exclusions use `PKSUMXF` consistently in application logic, CDISC export, metadata, labels, filtering, tests, and documentation.
- [ ] ADPP summary exclusions use `PPSUMXF` consistently in application logic, CDISC export, metadata, labels, filtering, tests, and documentation.
[...]

Issue #1335, opened 2026-05-28
Enhancement: Simplify TLG Order Details table and internalize PCSPEC filtering


Gero1999:
## Problem

The TLG Order Details table (`tab_tlg.R`) shows too many columns, making it hard for users to focus on what matters. Several columns are either internal implementation details or editable fields that add confusion without clear value.

In particular, the **Condition** column (currently editable) exposes an internal filtering mechanism — it holds values like `"URINE"` that are matched against `PCSPEC` to auto-select urine-specific TLGs. This is an implementation detail that shouldn't be user-facing.

### 1. Remove the Condition column from the UI

The `Condition` field in `tlg.yaml` (currently only `"URINE"` or `null`) should remain in the YAML as internal metadata, but should **not** appear in the Order Details table. Users don't need to see or edit it.

### 2. Internalize PCSPEC-based filtering in TLG functions

Instead of relying on the Condition column to decide whether a urine TLG is applicable, each TLG function that requires urine data should handle this internally:

- **If `PCSPEC` column exists in the data:** filter to urine records automatically before producing the output
- **If `PCSPEC` column does not exist:** produce the plot/table for all data and emit a warning (which should surface as an app notification, e.g., via `showNotification()`)

This makes urine TLGs self-contained — they work correctly regardless of whether the user manually toggled them in the order table.

Currently affected TLG functions (from `tlg.yaml`):
- `t_pkpt08_uri` — Mean Urine Amount and Percent table
- `l_pkcl02_uri` — Urine Concentration Listing
- `p_pkpg01_cum` — Mean Cumulative Urine plot
- `p_pkpg01_per` — Mean Percent Urine plot

### 3. Review and reduce visible columns

Current columns in the Order Details table:

| Column | Keep? | Notes |
|--------|-------|-------|
| Type | ✅ | Useful for grouping (Table/Listing/Graph) |
| Dataset | ✅ | PK Concentrations vs PK Parameters |
| PKid | ❓ | Useful for technical users, but could be hidden or collapsed |
| Output | ✅ | Description with link — the main informative column |
| Label | ❓ | Overlaps with Output/Description |
| Description | ❓ | Overlaps with Output (which already shows description as a link) |
| Condition | ❌ | Remove — internalize in TLG functions |
| Footnote | ✅ | User-editable, keep |
| Stratification | ✅ | User-editable, keep |
| Comment | ✅ | User-editable, keep |

[...]

Issue #1018, opened 2026-02-25
Enhancement: Exclusion of ADNCA rows in TLGs


Shaakon35:
### Description
In the manual exclusions
We would need to add a checkbox to define if we want to exclude the row of ADNCA in the NCA and/or also in the TLGs (summary table and mean plots)

once the exclusion is done, the variable ANL01FL (row excluded from summary tables and mean plots) shoudl be created in ADNCA (therefore ADNCA is updated)

### Definition of Done
The requirements for the feature to be complete:
- [ ] two additional checkboxes

PR #1439, opened 2026-08-24
Fix summary-exclusion flags filtering individual plots (#1438)

### LINKED ISSUE #1438: Bug: Summary-exclusion flags (PKSUMXF/PPSUMXF) filter individual TLG plots, not only summary tables and mean plots Labels: bug, TLG
##Description

Summary-exclusion flags are applied more broadly than intended. The documented intent (#1018, #1436) is that a summary-excluded record (PKSUMXF="Y" ADNCA / PPSUMXF="Y" ADPP) is removed from summary tables and mean plots, but kept everywhere records are inspected individually. Currently the exclusion applies to every TLG graph, including individual concentration plots (pkcg01).

It may also be of interest to consider that excluded results are reported apart in different summary tables and/or in the listings with a symbol such as "*"

##Documented intent

#1018: "exclude rows from TLGs (summary tables and mean plots)".
#1436: "excludes flagged records from summary statistics and applicable mean plots, but retains them in individual listings."
Where the divergence comes from — tlg_data_key() in inst/shiny/modules/tab_tlg/tlg_module.R:

tlg_data_key <- function(type, dataset) {
  if (identical(type, "listing")) paste0(dataset, "_all") else dataset
}
A binary listing-vs-everything switch: all type: Graph/type: Table get filtered data, so individual pkcg01 plots are filtered too. Scope quietly widened from "mean plots" to "graphs" in commit ba5a13184 (#1356). No issue/PR ever proposed filtering individual plots — it's a side effect of the routing shortcut.

Affected outputs — over-filtered: g_pkcg01_* (individual, should NOT be), g_pkcg02_* (combined, debatable). Correctly filtered: g_pkcg03_*, p_pkcg03_*_dose, p_pkpg01/02/03/04/06, all t_* tables. Correctly unfiltered: all l_* listings. (Exploration tab unaffected — reads pknca_data() directly.)

##Proposed solution — invert the default: only summary/mean functions self-filter.

Shared helper auto-detecting the flag, no-op if absent: data <- filter_summary_excluded(data).
Call it at the top of summary/mean funcs only (t_pkct01*, t_pkpt03*, t_pkpt07_norm, t_pkpt08_uri, t_pkpt11_gmr, g_pkcg03*, p_pkcg03_*_dose, p_pkpg01_*, p_pkpg02_doseprop, p_pkpg03/04/06).
Individual/combined plots and listings get raw data.
Retire the boundary filtered/_all split so there's one source of truth. Correct in script/export paths too.
Decisions to confirm: (A) self-filter in functions [recommended] vs (B) declarative summary_filter: true in tlg.yaml; and whether combined pkcg02 should filter.

DoD: individual plots show excluded records; decision made for combined; summary/mean still filter; listings unchanged; per-dataset scoping preserved (no cross-filter); works in script path; misleading "mean plots" comments corrected; tests assert individual-retains vs summary-drops.

Related: #1436, #1437, #1356, #1018.

Gero1999:
## Issue

Closes #1438

## Description

Summary-exclusion flags (`PKSUMXF` for ADNCA, `PPSUMXF` for ADPP) were applied to every TLG graph, so individual concentration plots (`pkcg01`) and combined plots (`pkcg02`) hid summary-excluded records. The documented intent (#1018, #1436) is that these flags remove records from summary tables and mean plots only, while keeping them everywhere records are inspected individually.

The old routing (`tlg_data_key()`) was a binary listing-vs-everything switch: listings got unfiltered data, everything else (all tables and all graphs) got filtered data. That scope quietly widened from "mean plots" to "graphs" in #1356.

This inverts the default: only summary/mean functions self-filter, and the filtered/`_all` boundary split is retired so there is a single source of truth per dataset.

Changes:
- Add `filter_summary_excluded()` in `R/utils-tlg.R`. It auto-detects the relevant flag (`PKSUMXF`/`PPSUMXF`), is a no-op when absent, scopes each dataset to its own flag (no cross-filtering), and uses `dplyr::filter()` so column `label` attributes survive the row subset.
- Call it at the top of the summary/mean functions only: `t_pkct01`, `t_pkpt03_col`, `t_pkpt07_norm`, `t_pkpt08_uri`, `t_pkpt11_gmr`, `pkcg03` (mean conc), and `p_pkpg01_cum`/`p_pkpg02_doseprop`/`p_pkpg03_boxp`. Their delegators (`_dose`, `_tad`, `_MP_col`, `p_pkpg01_per`, `p_pkpg04/06`, `g_pkcg03_*`, `p_pkcg03_*_dose`) inherit automatically.
- Individual (`pkcg01`) and combined (`pkcg02`) plots and all listings now receive raw data.
- Retire `tlg_data_key()` and `filter_tlg_excluded()`; `tab_tlg.R` exposes one unfiltered source per dataset.

Because filtering now lives in the TLG functions, the exported R script path gets the correct behaviour too.

**Combined plots (`pkcg02`) decision:** kept unfiltered. `pkcg02` overlays individual subject traces, so records are still being inspected individually.

**Unrelated naming bug found while testing (#1448):** adding name-based assertions surfaced a pre-existing bug on `main` — `pkcg01`/`pkcg02` named their output list from `adnca[["id_plot"]]`, but `id_plot` only exists on the intermediate `adnca_grouped`, so the list came back unnamed. No prior test checked plot names, so it went unnoticed. Fixed here by naming from `adnca_grouped` (tracked separately in #1448).

## Definition of Done

- [x] Individual plots show summary-excluded records
- [x] Decision made for combined plots (`pkcg02`: unfiltered)
- [x] Summary tables and mean plots still filter
- [x] Listings unchanged (unfiltered)
- [x] Per-dataset scoping preserved (no cross-filter)
- [x] Works in the exported R script path (functions self-filter)
- [x] Misleading "mean plots" routing comments corrected
- [x] Tests assert individual-retains vs summary-drops

## How to test

Flag a record with `PKSUMXF == "Y"` (ADNCA) or `PPSUMXF == "Y"` (ADPP) and confirm in the TLG tab that:
- it still appears in individual (`pkcg01`) and combined (`pkcg02`) plots and in listings, and
- it is dropped from summary tables (`t_pkct01`, `t_pkpt*`) and mean plots (`pkcg03`, `p_pkpg*`).

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

R is not available in this environment, so tests/lintr were verified statically (line-length and brace/paren balance) rather than executed; please run `devtools::test()` and `lintr::lint_package()` locally.

`main` has been merged into this branch and conflicts in `inst/shiny/modules/tab_tlg.R` and `tests/testthat/test-tab_tlg.R` resolved (kept the single unfiltered ADPP source from this branch and folded in the ADPP "run NCA first" notification from #1335).

Two test failures from the first CI run are addressed: the `g_pkcg01` name assertion failed due to the #1448 naming bug (fixed here), and the `t_pkct01` exclusion test expected an `n == 0` row for a fully-excluded cell — the table drops empty groups, so the test now asserts the row is absent.


Shaakon35:
`filter_summary_excluded()` currently loops over every matching flag column, so a data frame containing both `PKSUMXF` and `PPSUMXF` is filtered by both. That conflicts with the stated per-dataset rule: ADNCA must use only `PKSUMXF`, while ADPP must use only `PPSUMXF`. A retained/supplemental flag from the other dataset can therefore silently remove a row from the wrong summary output and change its statistics.

Could we make the scope explicit instead of inferring it from column presence? For example, give the helper a `flag` (or `dataset`) argument and call it as `filter_summary_excluded(data, flag = "PKSUMXF")` from ADNCA functions (`pkcg03`, `t_pkct01`) and with `flag = "PPSUMXF"` from ADPP functions (`t_pkpt*`, `p_pkpg*`).

Please replace the current both-flags test with a regression test using a data frame that contains both columns: when validating the ADPP path, a row flagged only with `PKSUMXF == "Y"` must remain; symmetrically, `PPSUMXF == "Y"` must not affect ADNCA.

Gero1999:
> `filter_summary_excluded()` currently loops over every matching flag column, so a data frame containing both `PKSUMXF` and `PPSUMXF` is filtered by both. That conflicts with the stated per-dataset rule: ADNCA must use only `PKSUMXF`, while ADPP must use only `PPSUMXF`. A retained/supplemental flag from the other dataset can therefore silently remove a row from the wrong summary output and change its statistics.
> 
> Could we make the scope explicit instead of inferring it from column presence? For example, give the helper a `flag` (or `dataset`) argument and call it as `filter_summary_excluded(data, flag = "PKSUMXF")` from ADNCA functions (`pkcg03`, `t_pkct01`) and with `flag = "PPSUMXF"` from ADPP functions (`t_pkpt*`, `p_pkpg*`).

yes i actually agree with this idea, will need to make more changees then from what was done

Shaakon35 (changes requested):
A few things to add:
- if a point in ADNCA is excluded, then don't take it in account in the mean plot calculation (that's done I think). But in any other plot, add a different shape fr this excluded point.
- same in the listings, for a pk concentration excluded, just add a "*" next to the Actual contration. e.g. "0.5645 *", and if at least one "*" is present, then add an automated footnote to indicate: "*: Concentration excluded from the summary table and mean plots."

do the exact same for ADPP in the listing. (no adpp plot woiuld need a different shape of the point).

Gero1999:
> A few things to add:
> 
> * if a point in ADNCA is excluded, then don't take it in account in the mean plot calculation (that's done I think). But in any other plot, add a different shape fr this excluded point.
> * same in the listings, for a pk concentration excluded, just add a "_" next to the Actual contration. e.g. "0.5645 _", and if at least one "_" is present, then add an automated footnote to indicate: "_: Concentration excluded from the summary table and mean plots."
> 
> do the exact same for ADPP in the listing. (no adpp plot woiuld need a different shape of the point).

I thikn I opened a different issue for this and assigned @h5hoang  but I will double check

Gero1999:
thiss one needs quick review @h5hoang

PR #1440, opened 2026-08-24
Fix NCA errors and predose interval start when start imputation is off & BLQ imputation in script (#1121, #1266, #1445)

### LINKED ISSUE #1121: Bug: Setting 'Impute Start Concentration' to FALSE causes NCA run to throw error
### Description
Setting 'Impute Start Concentration' to FALSE (default = TRUE) causes NCA run to thrown the below error

```
ERROR [2026-03-25 20:19:09] Error calculating NCA results:
ℹ In index: 1.
! Please report a bug.
: Error with interval start=-0.084, end=24.117: The following imputation functions were not found: PKNCA_impute_method_FALSE
```
### To reproduce
1. Switch 'Impute Start Concentration' to FALSE/OFF
2. Click Run NCA
3. Get error message that something went wrong. Console will have error message that 'PKNCA_impute_method_FALSE' not found 

#### Potential Choices
1. NCA should run without imputation, but would cause calculation of AUCs to differ without clarity on why (reason: time from C0 t0 C1 would not be included in calculation, C1 would be first time point of calculation)
2. Prevent analysis if FALSE (either provide warning or remove switch if it'll prevent analysis 100% of time, latter option requiring changing the settings file)

### LINKED ISSUE #1266: Bug: error in `PKNCA_impute_fun_list()

### LINKED ISSUE #1445: Bug: generated R-script template omits blq_imputation_rule in PKNCA_update_data_object
### Description

The generated R-script template (`inst/www/templates/script_template.R`, produced by `get_code()` / `get_session_code()` / `get_settings_code()`) does not reproduce the app's BLQ handling. A script exported from the app therefore yields different NCA results than the app itself whenever a BLQ imputation rule is configured.

**Root cause.** The app applies the BLQ rule in **two** places:

1. `PKNCA_update_data_object(..., blq_imputation_rule = settings()$data_imputation$blq_imputation_rule)` — sets the per-interval `impute` column so the BLQ method is part of interval setup (`inst/shiny/modules/tab_nca/nca_setup.R:137`, forwarded to `update_main_intervals()` → `R/intervals.R`).
2. `PKNCA_calculate_nca(blq_rule = settings()$data_imputation$blq_imputation_rule)` — sets the PKNCA `conc.blq` option at calculation time (`inst/shiny/modules/tab_nca.R:226`).

The template only replicates **(2)**: it extracts `blq_rule <- settings_list$settings$data_imputation$blq_imputation_rule` and passes it to `PKNCA_calculate_nca()`, but its `PKNCA_update_data_object(...)` call **omits** `blq_imputation_rule`. So the interval-level BLQ imputation the app performs is missing from the exported script.

This is a **pre-existing** bug (independent of the imputation-argument work in #1440).

### To reproduce

1. In the app, configure a BLQ imputation rule under Data Imputation.
2. Run NCA, then export the R script (session code).
3. Inspect the generated script: the `PKNCA_update_data_object(...)` block has no `blq_imputation_rule` argument.
4. Running the script reproduces different results than the app for BLQ-affected records.

### Expected behaviour

The template's `PKNCA_update_data_object(...)` call should pass `blq_imputation_rule` (from `settings_list$settings$data_imputation$blq_imputation_rule`), matching the app, so the exported script reproduces the app's results.

Gero1999:
## Issue

Closes #1121
Closes #1266
Closes #1445

## Description

Two problems surfaced when **Impute Start Concentration** was turned off. This PR fixes both.

### Bug A — `PKNCA_impute_method_FALSE` error

Running NCA with start imputation off failed with:

```
PKNCA_impute_fun_list(): The following imputation functions were not found: PKNCA_impute_method_FALSE
```

**Root cause — a dplyr data-masking name collision.** When start imputation is off, `create_start_impute()` is skipped, so the intervals data frame has **no `impute` column**. The BLQ step then referenced a bare `impute` symbol inside `mutate()`:

```r
impute = ifelse(is.na(impute) | impute == "", "blq", paste0("blq, ", impute))
```

With no column present, dplyr resolved the bare `impute` up into the calling environment, finding the **function argument** `impute = FALSE`. That produced the method string `"blq, FALSE"`; PKNCA split on the comma and looked for the non-existent `PKNCA_impute_method_FALSE`.

Because the app's default BLQ strategy ("No BLQ handling") still returns a non-NULL `keep/keep/keep` rule, the BLQ block runs by default — so simply toggling *Impute Start Concentration* off triggers it (#1121). It also triggers via a settings YAML with `impute_c0: no` (#1266).

### Bug B — interval start collapsed to the predose time

With start imputation off, the first interval started at the **predose time instead of C1** (the first sample at or after the dose), contrary to the documented behavior. This skews AUCs because the predose-to-C1 span is pulled into the interval.

**Root cause.** In `format_pkncadata_intervals()`, the sample feeding the start time was chosen by an **unordered `slice(1)`** per conc-group + dose. In real data the predose sample (negative `ARRLT`) can be that row, so with `start_from_last_dose = FALSE`:

```r
start = dose_time + ARRLT   # = dose_time + (negative predose ARRLT) = predose time
```

The `-0.084` interval start in the original #1121 traceback is this leak. (A code comment claimed non-predose samples were prioritized, but that was never implemented.)

### Bug C — generated R-script omitted `blq_imputation_rule` (pre-existing, #1445)

The exported session script did not reproduce the app's BLQ handling. The app applies the BLQ rule in two places — `PKNCA_update_data_object(blq_imputation_rule = ...)` at interval setup and `PKNCA_calculate_nca(blq_rule = ...)` at calculation — but the template only replicated the calculation step, so interval-level BLQ imputation was missing from generated scripts. This is independent of Bugs A/B and pre-existing on `main`; included here since it sits in the same BLQ/imputation area.

## Changes

**Bug A — `R/intervals.R`, `update_main_intervals()`:**

1. **Guarantee the column exists** before the BLQ block, mirroring the guard already used in `rm_impute_obs_params()`:
   ```r
   if (!"impute" %in% names(data$intervals)) {
     data$intervals$impute <- NA_character_
   }
   ```
   This also covers the case where `create_start_impute()` early-returns (empty intervals / no matched records) without adding the column.
2. **Pin the reference to the column** with the `.data` pronoun (`.data$impute`) so the symbol can never again fall through to a same-named variable — if the column were ever missing, dplyr now raises a clear error instead of silently building `"blq, FALSE"`.
3. **Rename the argument `impute` -> `start_impute`** so it can no longer collide with the per-interval `impute` column at all. The sole caller (`PKNCA.R`) already had a `start_impute` variable, so the call site reads `start_impute = start_impute`.

**Bug B — `R/intervals.R`, `format_pkncadata_intervals()`:**

Order the samples before `slice(1)` so the retained row is C1 (the first sample at or after the dose):

```r
arrange(ARRLT < 0, ARRLT, .by_group = TRUE) %>%
slice(1)
```

`ARRLT < 0` sorts post-dose rows (FALSE) ahead of predose rows (TRUE); ascending `ARRLT` then puts the smallest non-negative value (C1) first. The default path (`start_from_last_dose = TRUE`, where `start` is the dose time) is unaffected, and `max_end` is computed before the slice so end times are unchanged.

**Bug C — `inst/www/templates/script_template.R`:**

Add `blq_imputation_rule = settings_list$settings$data_imputation$blq_imputation_rule` to the template's `PKNCA_update_data_object(...)` call so the exported script matches the app.

## Definition of Done

- [x] NCA runs successfully with start imputation off (Bug A)
- [x] BLQ imputation still applies correctly in that case (`impute == "blq"`, never `"blq, FALSE"`)
- [x] Works whether the rule comes from the UI toggle or an uploaded settings YAML
- [x] Argument name no longer collides with the column
- [x] First interval starts at C1, not the predose time, when start imputation is off (Bug B)
- [x] Generated R-script applies the BLQ rule at interval setup, matching the app (Bug C)

## How to test

1. Load data that includes a predose sample, open the NCA setup.
2. Turn **Impute Start Concentration** off.
3. Run NCA. It completes without the `PKNCA_impute_method_FALSE` error, BLQ handling still applies, and the first profile's interval starts at C1 (not the predose time).

All three are covered by unit tests: `start_impute = FALSE` with a non-null `blq_imputation_rule` (Bug A), a predose-sample fixture asserting the interval start anchors to C1 (Bug B), and a `get_settings_code` assertion that the generated script passes `blq_imputation_rule` to `PKNCA_update_data_object` (Bug C).

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

`update_main_intervals()` is exported, so the `impute` -> `start_impute` rename is a signature change; given this is a pre-release dev version with a single internal caller, it is renamed directly rather than deprecated.

The single "Impute Start Concentration" switch drives both `start_impute` (c0 imputation) and `start_from_last_dose` (interval start anchoring). Bug B fixes the interval-start behavior; the broader question in #1121 of whether NCA should *warn or refuse* when it cannot compute AUCs (per @js3110's comment) is a separate behavioral decision not addressed here.

R is not available in this environment, so lintr and tests were verified statically (line length within 100, brace/paren balance) rather than executed. Please run `devtools::test()` and `lintr::lint_package()` locally. `man/update_main_intervals.Rd` was updated by hand since `devtools::document()` could not be run here.


Gero1999:
Now it works, I just realized that when start imputation is off, then start=predose... which might not be the best solution, we will try to fix and opt for C1/C0 when the start concentration imputation is off

Gero1999:
I can confirm it works

PR #1450, opened 2026-08-26
Fix M/P ratio TLG outputs

### LINKED ISSUE #1429: Bug: metabolite/parent TLG outputs summarize metabolite rows instead of computing M/P ratios
### Description
Three TLG catalog entries are labelled as metabolite-to-parent (M/P) ratio outputs, but none of them compute a ratio. They filter the data down to metabolite rows and then summarize those rows as-is, so the output shows raw metabolite values under a heading that promises a ratio.

Raised by @Shaakon35 in [#1356](https://github.com/pharmaverse/aNCA/pull/1356#issuecomment-5116050169)

  | Entry | Label in `tlg.yaml` | Actual behaviour |
  |---|---|---|
  | `t_pkpt03_MP_col` | pkpt03 - Summary M/P Ratios (stats in cols) | summarizes metabolite rows |
  | `l_pkpl01_mp` | pkpl01 - M/P Ratio Listing | lists metabolite rows |
  | `p_pkpg06_mp` | pkpg06 - Boxplot Metabolite/Parent Ratios | boxplots of metabolite rows |

  ## Root cause

  All three delegate to `filter_metabolite_rows()` (`R/utils-tlg.R:86`) and then hand the filtered data
  to the ordinary non-ratio function:

  - `R/t_pkpt.R:215` — `t_pkpt03_col(filter_metabolite_rows(data, "t_pkpt03_MP_col"), ...)`
  - `R/l_pkpl01.R:138` — `l_pkpl01(filter_metabolite_rows(data, "l_pkpl01_mp"), ...)`
  - `R/g_pkpg.R:177` — `p_pkpg03_boxp(filter_metabolite_rows(data, "p_pkpg06_mp"), ...)`

  `filter_metabolite_rows()` also has to guess which rows are metabolites, via a three-tier fallback (`METABFL` → `PPCAT` grep `"metab"` → `PARAM` grep `"metab"`). `METABFL` and `DOSETRT` are not part of a default ADPP, so in practice this is a substring match on an analyte name. Even when it picks the right rows, nothing identifies the denominator, and a PD parameter that happens to be named with "metab" would be swept in.

  ## Expected behaviour

  M/P ratios are already computed upstream. When a user sets them up in **Parameter Selection > Ratios**, `calculate_ratios.data.frame()` (`R/ratio_calculations.R:207`) writes ratio rows carrying:

  ```
  PPANMETH = "<PPTESTCD> TO <PPTESTCD_ref> [<refcol>: <value>, ...]"
  ```

  The TLG entries should consume those existing ratio rows rather than re-deriving anything:

  - **Select ratio rows on `grepl(" TO ", PPANMETH)`.**
  - **Numerator / metabolite context:** `PPCAT` (`export_cdisc` sets `PPCAT = PARAM`, then overwrites
    `PARAM = PPTEST`, so in ADPP the analyte lives in `PPCAT` and `PARAM` is the parameter name).
  - **Parent / reference context:** parse out of `PPANMETH`.

  Some previous attempts that were incorrect during the #1356 review:

  - **Do not** select on `grepl("^MR", PARAMCD)`. It matches eight mean-residence-time parameters, and it *misses* real ratio rows — the default `PPTESTCD` for a ratio is `RA<param>` (e.g. `RAAUCLST`). `MR*` only appears when the user picks the metabolite-ratio preset
  - `PPANMETH` is free text with the reference group keys bracketed, not a clean reference column, so parsing needs to be defensive

  ## To reproduce

  1. Upload a dataset with a parent and a metabolite analyte
  2. **Parameter Selection > Ratios** — set up an M/P ratio and run NCA
  3. **TLG** tab, enable `pkpt03 - Summary M/P Ratios`, submit the order
  4. Observe: the table shows summary statistics of the metabolite's raw values, not a ratio

  ## Definition of Done

  - [ ] `t_pkpt03_MP_col`, `l_pkpl01_mp` and `p_pkpg06_mp` display actual M/P ratios sourced from the ratio rows written by `Parameter Selection > Ratios`
  - [ ] Ratio rows are selected on `PPANMETH`, not on a `PPTESTCD`/`PARAMCD` name prefix
  - [ ] Parent/reference analyte is shown alongside the metabolite so the ratio is self-explanatory
  - [ ] A clear message is shown when the user has not set up any ratios, instead of an empty or misleading output
  - [ ] `filter_metabolite_rows()` is removed or narrowed to whatever still genuinely needs it
  - [ ] Unit tests cover ratio-row selection, the no-ratios-configured path, and the mean-residence-timefalse-positive case
  - [ ] `l_pkpl04_mp` (treatment ratios) sanity-checked against the same mechanism

  ## Related

  - #1356 — Expansion of TLG Catalog (where this was raised)

h5hoang:
## Issue

Closes #1429

## Description

`t_pkpt03_MP_col`, `l_pkpl01_mp` and `p_pkpg06_mp` never divided anything. They filtered to metabolite rows with `filter_metabolite_rows()` and summarized those as-is, so the output showed raw metabolite values under a ratio heading. On `adnca_example` AUClast showed 11.399 (the metabolite's value) instead of the ratio of about 0.585.

They now use the ratio rows that **Parameter Selection > Ratios** computes during the NCA run:

- Rows are selected on `PPANMETH`, not on a `PPTESTCD`/`PARAMCD` prefix. A row is a ratio when `"<test> TO <ref>"` is the whole field, or when a reference bracket comes with it.
- A reference on the analyte column means metabolite/parent, and goes to the three entries above. Everything else (treatment, dose profile, route, specimen) goes to `l_pkpl04_mp`.
- Two derived columns: `RATIOREF` (denominator) and `RATIO` (`"Metab-DrugA / DrugA"`). `RATIO` is the split key, so the denominator is named on every output.
- With no ratios set up, the output errors and says how to add them.

`filter_metabolite_rows()` is deleted.

Two fixes found while checking `l_pkpl04_mp`:

- `"DOSE 2 / DOSE 1"` is the same label for parent and metabolite, so splitting on `RATIO` alone put both in one column and `pivot_wider` kept one value per subject. Half the ratios were dropped. `PPCAT` now splits the listings too.
- The listing and table took the leading row even when its value was missing, while the plots dropped missing values first, so the three disagreed about which row to show. All three now prefer a row that has a value.

## Definition of Done

- [x] `t_pkpt03_MP_col`, `l_pkpl01_mp` and `p_pkpg06_mp` display actual M/P ratios sourced from the ratio rows written by `Parameter Selection > Ratios`
- [x] Ratio rows are selected on `PPANMETH`, not on a `PPTESTCD`/`PARAMCD` name prefix
- [x] Parent/reference analyte is shown alongside the metabolite so the ratio is self-explanatory
- [x] A clear message is shown when the user has not set up any ratios, instead of an empty or misleading output
- [x] `filter_metabolite_rows()` is removed or narrowed to whatever still genuinely needs it
- [x] Unit tests cover ratio-row selection, the no-ratios-configured path, and the mean-residence-time false-positive case
- [x] `l_pkpl04_mp` (treatment ratios) sanity-checked against the same mechanism

## How to test

1. Upload `adnca_example`.
2. **Parameter Selection > Ratios**, add `AUClast` as test with `PARAM: DrugA` as reference. Add a second on `Cmax`. Run the NCA.
3. **TLG** tab, enable `pkpt03 - Summary M/P Ratios`, `pkpl01 - M/P Ratio Listing`, `pkpg06 - Boxplot Metabolite/Parent Ratios`, submit.

Split should read `RATIO: Metab-DrugA / DrugA / PPSPEC: SERUM`, values around 0.4 to 0.5.

Also worth checking:

- Enable the same three without configuring any ratio. Each should say how to add one, not go blank.
- Add a ratio against `ATPTREF: DOSE 1` and enable `pkpl04 - Treatment Ratio Listing`. One listing per analyte, `PPCAT: DrugA` and `PPCAT: Metab-DrugA` in the split names.
- A subject with a value in the listing should be in the table's `n` and in the box plot.

## Contributor checklist

- [x] New logic is documented

## Notes to reviewer

Ratios have to be set up before the NCA runs. 

`PPANMETH` parsing handles three cases, all tested: a same-group ratio has no bracket so the label comes from the parameter pair, `", "` can appear inside an analyte name so unprefixed fragments are rejoined, and `export_cdisc()` can prepend an analysis method with `"; "`.

Not fixed here: the summary table and box plot still show one dose profile per subject in a multi-dose study, since `.build_pkpp_table()` and `p_pkpg03_boxp()` dedupe to one row per subject x parameter and `ATPTREF` isn't in the default stratification. On `adnca_example` that's `n = 2` and `n = 3` per arm against 16 and 24 ratio rows. Pre-existing on `main` and it hits every `pkpt03` table, not just the ratio ones.


Shaakon35:
Hello @h5hoang 

I think it's not implemented the right way.
Metabolite / Parents TLGs should be easier to build:
- Boxplots for Metabolite/Parent ratios by TRT
- Individual listings of metabolite to parent ratios vs. TRT, by TRT/dose (Metabolite/Parent Ratio as header names)
- Summary of metabolite to parent ratios by TRT [stats in columns]

For the three of them it should be easy:
in ADPC (which is the same as ADNCA), the metabolite is defined by METABFL = "Y".
If you do adpc %>% select(PARAM, METABFL, DOSETRT) you should have:
maximum two rows with the same DOSETRT. in the case of the dummy dataset:

PARAM = DrugA, METABFL = "", DOSETRT = Drug A
PARAM = Metab-DrugA, METABFL = "Y", DOSETRT = Drug A
-> so we need to associate these two parameters. one being the parent, the other is the metabolite.

Then re-ruse the logic of the code row 10, 18, 19 and 37 to 45 (which is jsut simply for every pk parameter do the value of the parent divided by the value of the metabolite (if both of them are available).
[g_pkpar_boxp_MP.txt](https://github.com/user-attachments/files/31912578/g_pkpar_boxp_MP.txt)]
What we need to make sure when doing a ratio is avoid Infinity values, na values etc.

Then, apply the same logic for the 3 TLGs mentionned previously.[

h5hoang:
thanks @Shaakon35 for clearing things up, that was really helpful!! it should be fixed now where the three M/P TLGs use PARAM, METABFL and DOSETRT to associate the parent and metabolite, then calculate metabolite / parent from the matched NCA values so no manual ratio setup is needed.

missing/non-finite values and zero denominators are excluded. i've checked the tables, listings and plots against adnca_example.csv and the numbers match the calculated ratios

one thing i wanted to confirm ask is your example uses AVISIT, but adnca_example has multiple ATPTREF dose profiles and no AVISIT. i've kept each dose profile separate, both when matching the parent/metabolite values and when displaying the outputs. would you expect separate outputs per dose, or the doses shown together within each output?

also could you take another look when you get a chance? thanks again!

Gero1999:
>  it should be fixed now where the three M/P TLGs use PARAM, METABFL and DOSETRT to associate the parent and metabolite, then calculate metabolite / parent from the matched NCA values so no manual ratio setup is needed

I don't agree with thiss @Shaakon35 @h5hoang . In my opinion:

- There should be only 1 point/code to calculate ratios, otherwise we risk misalligning
- The user can always use a standard settings file in the session to create their ratios "automatically" 
- ADPP must reflect ALL values represented in TLGs

Gero1999:
> I don't agree with thiss @Shaakon35 @h5hoang . In my opinion:
> 
> * There should be only 1 point/code to calculate ratios, otherwise we risk misalligning
> * The user can always use a standard settings file in the session to create their ratios "automatically"
> * ADPP must reflect ALL values represented in TLGs

I think mee and @Shaakon35 aalready agreed on this

PR #1452, opened 2026-08-27
Block exports when selected outputs fail validation

### LINKED ISSUE #1451: Enhancement: metadata-driven CDISC data-type validation + QC report on export
### Motivation

When results are exported/saved, there is currently no automated check that the
CDISC datasets (ADNCA, ADPP, PP) conform to the data types and lengths declared
in the package metadata. Adding a lightweight validation step lets us:

1. Detect values of the wrong data type and prevent non-conforming data from
   being saved (a data-integrity control; cf. 21 CFR 11.10(a)).
2. Emit a human-readable QC report as an export artifact.

This is intentionally **not** a full CDISC conformance engine (no controlled
terminology / structural rules). It covers exactly what the existing
`metadata_nca_variables` table can assert: expected **class** and **max length**
per variable. It adds **no new package dependencies** — it uses `dplyr` and
`htmltools`, which are already in `Imports`.

### Proposed API

**`R/validate_cdisc_types.R`**

- `validate_cdisc_types(cdisc_data, metadata = metadata_nca_variables, check_length = TRUE)`
  Checks each column of the supplied CDISC datasets (named list keyed
  `pp`/`adpp`/`adnca`, as returned by `export_cdisc()`) against the metadata.
  Returns a tidy findings data frame:
  `Dataset, Variable, Check, Severity, Expected, Observed, N_Affected, Message`.
  Zero rows means the data conforms.
  - Type → class mapping from `metadata_nca_variables$Type`:
    - `text` / `Char` / `Character` → `character`
    - `float` / `integer` / `Num` → `numeric`
    - `dateTime` / `duration` → `character` (ISO 8601 strings)
    - unknown/empty types are skipped (never fails on metadata it can't interpret)
  - Checks:
    - **class** (severity `error`): column storage class must match the expected
      class. Numeric accepts integer/double; character accepts character/factor.
      All-`NA` columns pass.
    - **length** (severity `error`): character values must not exceed the
      declared `Length`.
    - **unknown_variable** (severity `warning`): columns present in the data but
      absent from the metadata are surfaced but do **not** block saving.

- `cdisc_validation_blocks_save(findings)`
  Predicate returning `TRUE` when any `error`-severity finding exists. This is
  the hook the Save/export flow uses to refuse a save.

**`R/cdisc_validation_report.R`**

- `cdisc_validation_report(findings, datasets_checked = NULL, project = NULL, generated_at = Sys.time())`
  Renders the findings into a self-contained, styled HTML string (inline CSS,
  no external assets): PASS/FAIL banner, per-finding table, metadata-driven note.
- `write_cdisc_validation_report(cdisc_data, target_dir, filename = "cdisc_validation_report.html", project = NULL, metadata = metadata_nca_variables)`
  Convenience wrapper for the export flow: runs the validation, writes the HTML
  report into `target_dir`, and returns `list(findings, report_path, blocks_save)`.

### Wiring into the export/Save flow

```r
result <- write_cdisc_validation_report(cdisc_data, target_dir = export_dir)

if (result$blocks_save) {
  showNotification(
    sprintf("Save blocked: %d data-type error(s). See validation report.",
            sum(result$findings$Severity == "error")),
    type = "error"
  )
  return()  # abort BEFORE writing the datasets
}
# ... proceed with the existing export ...
```

### Design decisions (open to feedback)

- All-`NA` columns are treated as conforming (an empty column has no type to violate).
- Unknown columns are warnings, not errors (do not block save). Can be made strict.
- Scope is class + length only; report footer states it does not replace full
  CDISC conformance validation.

### Tests

`tests/testthat/test-validate_cdisc_types.R` covers: the type→class mapping,
conforming data, wrong-type in both directions, length violations,
`check_length = FALSE`, unknown columns as non-blocking warnings, all-`NA`
columns, NULL/empty input, a smoke test against the packaged
`metadata_nca_variables`, and the HTML report (PASS/FAIL/write).

### Definition of Done

- [ ] `validate_cdisc_types()` and `cdisc_validation_blocks_save()` implemented and exported
- [ ] `cdisc_validation_report()` and `write_cdisc_validation_report()` implemented and exported
- [ ] Export/Save flow blocks saving when error-severity findings exist and writes the HTML report as an artifact
- [ ] `tests/testthat/test-validate_cdisc_types.R` added and passing
- [ ] `devtools::document()` to export the new functions in `NAMESPACE`
- [ ] `devtools::test()`
- [ ] `lintr::lint_package()`

Gero1999:
## Issue

Closes #1451

## Description

This PR adds pre-export validation to the Save flow so invalid selected outputs are detected before any export files are written.

The validation covers three levels:

- Export object validation for all selected outputs:
  - plot outputs must be `ggplot` or `plotly`
  - table outputs must be data frames with unique, non-empty names and atomic columns
  - code outputs must be single character strings
  - unsupported objects block the export
- Selected generated artifacts are preflighted:
  - settings must be YAML-serialisable
  - R-script exports require serialisable input data and an available template
  - slides require NCA results, an output format, and their required packages
- CDISC data validation for selected `PP`, `ADPP`, and `ADNCA` outputs:
  - column storage classes are checked against `metadata_nca_variables`
  - numeric-backed datetime variables ending in `DTM` accept either numeric SAS-datetime storage or R `POSIXt` values
  - character values are checked against declared metadata lengths
  - indexed CDISC variable families are matched to their metadata templates
- Export writer failure handling:
  - XPT writer errors now propagate instead of being logged only to the R console
  - non-validation export failures show an app notification before the export stops

When error-severity validation findings are detected, the export is blocked before files are written and the user receives an error notification identifying the affected output(s). Warning-only findings, such as columns not declared in metadata, are reported but do not block export.

This is intended to support the CFR 21 Part 11 expectation that non-conforming generated outputs are not silently exported.

## Definition of Done

- Selected outputs are validated before export files are written.
- Invalid output object types block export.
- CDISC type and length errors block export.
- Valid non-missing ADNCA `PCRFTDTM` values are accepted whether represented numerically or as `POSIXct`.
- XPT writer failures block export instead of being swallowed.
- Selected settings, R-script/raw-data, and slide artifacts fail validation before export when their required structure is unavailable.
- The user is notified when validation or export writing blocks the export.
- Warning-only findings do not block export.
- Validation behavior is covered by unit tests and integration-style export tests.
- Documentation and NEWS are updated.

### Automated checks

Run:

```r
devtools::document()
devtools::test()
lintr::lint_package()
```

The relevant test coverage is in:

- `tests/testthat/test-validate_cdisc_types.R`
- `tests/testthat/test-validate_export_outputs.R`
- `tests/testthat/test-export_cdisc.R`
- `tests/testthat/test-zip-utils.R`

### Manual app validation

First confirm the successful path:

1. Start the Shiny app.
2. Upload or load data and run the NCA workflow until exportable results are available.
3. Open the Save/export panel.
4. Select valid outputs, including at least one CDISC output such as `ADNCA`, `ADPP`, or `PP`.
5. Click Save and confirm the export completes successfully.

Then confirm the blocking path with a temporary developer-only mutation:

1. In `inst/shiny/functions/zip-utils.R`, add the following temporary line inside `prepare_export_files()`, immediately after `export_list` is built and before `.validate_outputs_pre_export()` is called:

```r
export_list$CDISC$adnca$AVAL <- as.character(export_list$CDISC$adnca$AVAL)
```

`ADNCA.AVAL` is declared as `float` in `metadata_nca_variables`, so this forces a selected CDISC output to fail metadata-driven type validation.

2. Restart the app.
3. Run the NCA workflow again.
4. In the Save/export panel, select `ADNCA` for export.
5. Click Save.
6. Confirm that:
   - the export is blocked,
   - no output files are written,
   - the app shows an error notification,
   - the notification identifies the affected output.
7. Remove the temporary debug line before committing or merging.

A similar length-check test can be done by temporarily assigning a value longer than the metadata length, for example setting `export_list$CDISC$adnca$STUDYID[1]` to a string longer than 8 characters before validation.

Also confirm that warning-only CDISC findings, such as an extra column not declared in metadata, do not block export.

## Contributor checklist

- [x] New logic is documented

## Notes to reviewer

The validation applies a basic structural contract to every selected artifact. PP, ADPP, and ADNCA additionally use their metadata-driven CDISC schema; other tables are checked for export-safe structure rather than a dataset-specific CDISC schema.

For SAS interoperability, the metadata continues to define `PCRFTDTM` as integer-backed. The validator now also accepts R native `POSIXct`/`POSIXt` representation for variables whose names end in `DTM`; the exception does not apply to ordinary integer variables.

Unexpected export failures use a clean user notification and remove any partially created ZIP. Warning-only findings are shown to the user and do not block export.

The export-failure path is covered in `tests/testthat/test-zip-utils.R`: selected invalid outputs fail before files are written, validation notifications are emitted, and XPT writer failures propagate instead of being swallowed.


Gero1999:
thiss one will be needed for validation @Shaakon35

Shaakon35:
The PR description says validation writes `cdisc_validation_report.html` as an artifact, but I cannot find any report writer, generated file, or test for it. At present `.validate_outputs_before_export()` only shows a brief notification and aborts, so a user whose export is blocked cannot see the affected variables, expected/observed values, or row counts needed to correct the data.

Could we add a report-generation step that renders the full `findings` data frame to HTML (for example via a small `.write_cdisc_validation_report(findings, path)` helper)? Because `stop()` prevents the normal ZIP from being delivered, the report also needs a user-visible delivery path: either show the findings in a modal with a dedicated report download, or produce and download a report-only ZIP when validation fails. Writing it solely to the temporary export directory before aborting would not make it available to the user.

Please add an integration test that triggers a type/length error and verifies both the report contents and the chosen user-facing delivery path.

Shaakon35:
### High severity: valid ADNCA exports with `PCRFTDTM` are blocked

`export_cdisc()` creates a non-missing `ADNCA$PCRFTDTM` as `POSIXct`, while the metadata declares this variable as `integer`. The new pre-export check maps `integer` to `numeric` and then requires `is.numeric(col)`, which is false for `POSIXct`.

Because this validation runs before any output is written, exports that include a normal reference datetime are rejected even though the app generated the dataset. Please handle datetime-backed CDISC variables in the validator (or align the export representation and metadata), and add an integration test covering a non-missing `PCRFTDTM`.

Shaakon35 (approved):

KOBANAK (commented):
The PR states that warning-only findings, such as CDISC columns absent from `metadata_nca_variables`, are reported without blocking export but when I try I don't see any warning. 

I also agree with Valentin comment regarding non mising values for `PCRFTDTM` variable blocking the export because the type is not matching the metadata definition.

Issue #1102, opened 2026-03-23
Bug: PKNCA release blocker


billdenney:
### Description
I'm trying to make a new release of PKNCA, and 8 tests failed on CRAN with this attempt. Do the errors at the bottom make sense to you (I can try to troubleshoot if they're not obvious).

### To reproduce
Steps to reproduce the issue are run aNCA with the proposed release version of PKNCA (https://github.com/humanpred/pknca/pull/523).

### Expected behaviour
All tests pass

### Errors

Originally from https://win-builder.r-project.org/incoming_pretest/PKNCA_0.12.2_20260323_150409/reverseDependencies/summary.txt (but this link will disappear in about a week)

```
Package check result: OK

Changes to worse in reverse depends:

Package: aNCA
Check: tests
New result: ERROR
    Running ‘testthat.R’ [44s/44s]
  Running the tests in ‘tests/testthat.R’ failed.
  Complete output:
    > # This file is part of the standard setup for testthat.
    > # It is recommended that you do not modify it.
    > #
    > # Where should you do additional test configuration?
    > # Learn more about the roles of various files in:
    > # * https://r-pkgs.org/testing-design.html#sec-tests-files-overview
    > # * https://testthat.r-lib.org/articles/special-files.html
    > 
    > library(testthat)
    > library(shinytest2)
    > library(aNCA)
    > 
    > test_check("aNCA")
    Joining with `by = join_by(ATPTREF)`
    * USUBJID -> id
    * PARAM -> analyte
    * DOSETRT -> drug
    * AVAL -> CONC
    * AVALU -> CONC_unit
    * DOSE -> dose
    * ADOSEDUR -> dosedur
    * AFRLT -> TIME
    * ATPTREF -> DOSNO
    Row 1: Converted volume from 1 L to 1000 mL based on concentration unit ug/mL
    Row 1: Converted volume from 2 mL to 2 g based on concentration unit mg/g
    Row 1: Converted volume from 1 L to 1000 mL based on concentration unit ug/mL
    Row 2: Converted volume from 0.5 mL to 0.5 g based on concentration unit mg/g
    Loading required namespace: ggh4x
    Saving _problems/test-pivot_wider_pknca_results-71.R
    Saving _problems/test-pivot_wider_pknca_results-85.R
    Saving _problems/test-pivot_wider_pknca_results-92.R
    Saving _problems/test-pivot_wider_pknca_results-119.R
[...]

PR #1455, opened 2026-08-28
fix: shorten PKNCA parameter descriptions to the SDTM 40-character limit

billdenney:
Part of #1102.

The `desc` strings registered with `PKNCA::add.interval.col()` for `ertlst`, `ertmax`, and `volpk` are 96, 88, and 41 characters. SDTM allows at most 40.

PKNCA 0.12.1.9000 warns on longer descriptions, naming the parameter and the offending text. An earlier development build enforced the limit with an assertion instead, which made aNCA **fail to install outright** — `add.interval.col()` runs at load time, so the error surfaced during lazy loading before any aNCA code could run.

Only the registered `desc` values change. The roxygen `@title` and `@return` prose is not SDTM-constrained and is left alone, so no documentation needs regenerating.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 ## Maintenance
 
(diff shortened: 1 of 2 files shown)


billdenney:
Fixed by @Gero1999 in https://github.com/pharmaverse/aNCA/commit/571f624c9aa77ca7ca34827fd04859e48e256879

PR #1456, opened 2026-08-28
fix: read grouping columns from the PKNCAconc object, not the results

billdenney:
Part of #1102. Related to the "group_vars consider `start` & `end`" item in #481.

`group_vars()` on a `PKNCAresults` object now returns the interval bounds alongside the grouping columns:

```r
PKNCA 0.12.1       group_vars(results): id
PKNCA 0.12.1.9000  group_vars(results): start,end,id
```

`group_vars()` on `PKNCAconc` and `PKNCAdose` is unchanged.

Both call sites in `inst/shiny/functions/zip-utils.R` want the *data's* grouping columns and apply the answer to concentration data, which has no `start` or `end`:

* **line 211** subsets `d_conc_i` by them → `undefined columns selected` (4 failures)
* **line 509** builds `group_by_vars`, which flows through `color_by` into `g_pkconc_summary()` and reaches `group_by()` → ``Column `start` is not found. Column `end` is not found.`` (2 failures)

Reads the grouping from the PKNCAconc object instead. That is what both sites mean, and it is stable across both PKNCA versions.


Gero1999:
changes LGTM, thanks!

PR #1459, opened 2026-08-28
test: cover the NULL volume_column guard in detect_study_types

billdenney:
Follow-up to 571f624 on `1102-bug/pknca-dev-compat`.

That commit's `detect_study_types()` guard handles an explicit `NULL` as well as a missing argument, but nothing exercises that path, so it can regress silently. The original bug was exactly this shape — `missing()` looked sufficient until PKNCA started passing `NULL`.

Asserts that an explicit `NULL` gives the same result as omitting the argument and does not classify the data as excretion.

Verified the test gates: with the guard reverted to its previous `missing()`-only form it errors with `Can't convert NULL to a symbol`; with the guard in place the file passes 27 of 27.


Gero1999:
changes LGTM, thanks!

Issue #1461, opened 2026-08-31
Enhancement: Offer ROUTE as a "Group by" option in NCA descriptive statistics (and fix ROUTE join crash)


Gero1999:
### Description

Users want to group the NCA descriptive statistics table (NCA tab) by `ROUTE`. Today `ROUTE` is not offered in the descriptive-statistics "Group by" picker, so users try to add it as a supplemental grouping variable in the Mapping section instead — which crashes the app:

```
Error in purrr::pmap: ℹ In index: 1.
Caused by error in current_interval[rep(1, nrow(calculated_interval)), c("start", "end", options$keep_interval_cols)]:
! Can't subset columns that don't exist.
✖ Column `ROUTE` doesn't exist.
```

**Key finding:** `ROUTE` is already always carried into the results data (`res_nca()$data$conc$data`) by `PKNCA_create_data_object()` — it is included in `group_columns <- intersect(colnames(adnca_data), c("STUDYID", "ROUTE", "DOSETRT"))`, even though it is not part of the PKNCA grouping formula. There is existing precedent in `inst/shiny/modules/tab_nca/nca_results.R` (`extra_vars_to_keep <- c(grouping_vars(), "DOSEA", "ATPTREF", "ROUTE")`) that already treats `ROUTE` as an always-available extra column.

So users do **not** need to add `ROUTE` in the Mapping section at all — the descriptive statistics module just wasn't offering it in its "Group by" picker.

**Two parts:**

1. **Feature** — Add `"ROUTE"` to the descriptive-statistics "Group by" picker (mirroring how `DOSEA`/`ATPTREF` are handled). The existing `classification_cols[... %in% names(res_nca()$data$conc$data)]` filter keeps it safe if `ROUTE` is ever absent.

2. **Bug fix (defensive)** — Fix the crash regardless. Root cause: in `PKNCA_update_data_object()` (`R/PKNCA.R`), after `format_pkncadata_intervals()` returns intervals that already contain `ROUTE` (passed via `keep_interval_cols`), the code unconditionally re-joins `ROUTE` from the dose data with a `by=` that doesn't include `ROUTE`. `dplyr` then produces `ROUTE.x`/`ROUTE.y` and the plain `ROUTE` column disappears — which `pk.nca()` later requires via `keep_interval_cols`, causing the crash. Fix: only join route columns that aren't already present in the intervals.

[...]

PR #1464, opened 2026-08-31
Custom naming and per-study-type selection for Partial Interval Calculations

### LINKED ISSUE #1463: Enhancement: Custom naming and per-study-type selection for Partial Interval Calculations
### Description

The **Partial Interval Calculations** section (Parameter Selection tab) lets users define partial-AUC and related interval parameters (AUCINT, CAVGINT, RCAMINT, FREXINT, ...). Today each row has only three fields: **Parameter**, **Start**, **End**, and the interval is:

1. always named with the standard auto-generated pattern (`PPTESTCD_<start>-<end>`, e.g. `AUCINT_0-24`), with no way to give it a meaningful label; and
2. always calculated for **every** study type in the dataset, with no way to restrict it (e.g. compute a urine `FREXINT` interval only for `Excretion Data`).

This makes results harder to read and forces irrelevant calculations.

### Proposed change

Extend each partial-interval row with two optional fields:

1. **Custom name** — optional free-text label. When provided, it replaces the standard `PPTESTCD_<start>-<end>` name in the user-facing outputs (results table, descriptive statistics, plots, and ratio-parameter options). When empty, the current standard naming is kept. The underlying PKNCA parameter and CDISC `PPTESTCD` used for the actual calculation are unchanged.
2. **Study type** — restrict which study type the interval is calculated for. Choices are `All` (default, current behaviour) plus each study type detected in the loaded data (via `detect_study_types()`). Selecting a specific type calculates that partial interval only for groups of that type. To target several specific types, add one row per type.

Both fields must round-trip through the **settings YAML** and the **exported R script**, so saved/restored settings and generated scripts reproduce the same intervals.

### Affected areas

- `inst/shiny/modules/tab_nca/setup/settings.R` — `int_parameters` schema (new `custom_name`, `study_type` columns), table columns, edit observers, help text, settings serialize/restore
- `R/intervals.R` — `update_main_intervals()`: carry `custom_name` onto interval rows; filter each manual interval by `study_type`
- Naming sites: `R/utils.R` (`rename_interval_params`), `R/label_operators.R` (`add_label_attribute`), `R/pivot_wider_pknca_results.R`, and ratio-parameter options
- `inst/www/templates/settings_template.yaml` and `inst/www/templates/script_template.R`
- `R/zzz.R` globalVariables if needed
- Tests under `tests/testthat/`

### Definition of Done

- [ ] Each partial-interval row has optional **Custom name** and **Study type** fields in the UI
- [ ] Custom name (when set) is used in results, descriptive stats, plots, and ratio options; empty falls back to the standard name
- [ ] Study type restricts calculation to matching groups; `All` keeps current behaviour
- [ ] Both fields are saved to and restored from the settings YAML
- [ ] Both fields appear in and drive the exported R script
- [ ] Backward compatible with existing settings files that lack the new columns
- [ ] New logic covered by unit tests; NEWS + version bump

Gero1999:
## Issue

Closes #1463

## Description

Extends the **Partial Interval Calculations** section (Parameter Selection tab) with two optional per-row fields:

1. **Custom Name** — optional free-text label. When set, it replaces the standard `PPTESTCD_<start>-<end>` name (e.g. `AUCINT_0-24`) in the results table, plots, and ratio-parameter options. When blank, the standard name is kept. The underlying PKNCA parameter and CDISC `PPTESTCD` used for the actual calculation are unchanged.
2. **Study Type** — restricts the interval to a single study type detected in the data (via `detect_study_types()`). `All` (default) keeps the previous behaviour of calculating for every study type. To target several specific types, add one row per type.

Both fields round-trip through the settings YAML and the exported R script, and are backward compatible with settings files saved before this change.

### Design notes
- The custom name is carried through `pk.nca()` via a new `interval_name` column added to `keep_interval_cols` in `PKNCA_update_data_object()`. `update_main_intervals()` always adds this column (NA for main intervals) so it can be kept unconditionally — avoiding the missing-column crash pattern seen with ROUTE (#1461).
- `int_parameters` stays a flat table: `study_type` is a single scalar per row. This keeps the YAML and generated script simple (parallel vectors / plain columns) rather than nesting a per-row list. The `reactable.extras` cell editors are single-select only, which also fits this model.
- The script and settings serialization read the whole `int_parameters` table generically, so no template code changes were needed beyond documenting the new columns in `settings_template.yaml`.

### Key files
- `inst/shiny/modules/tab_nca/setup/settings.R` — schema, table columns (Custom Name text + Study Type dropdown), edit observers (text vs numeric coercion), detected study-type choices, help text, backward-compatible restore
- `R/intervals.R` — `update_main_intervals()`: normalise optional columns, filter each partial interval by `study_type`, carry `interval_name`
- `R/PKNCA.R` — add `interval_name` to `keep_interval_cols`
- `R/utils.R`, `R/label_operators.R`, `R/pivot_wider_pknca_results.R` — honour custom name in the user-facing naming; drop the internal `interval_name` from the reshaped output
- `inst/shiny/modules/tab_nca/setup/ratio_calculations_table.R` — `.build_interval_param_options()` uses the custom name
- `inst/www/templates/settings_template.yaml` — document new keys
- `R/zzz.R` — globalVariables (`custom_name`, `study_type`, `interval_base`)

## Definition of Done

- [x] Each partial-interval row has optional **Custom Name** and **Study Type** fields in the UI
- [x] Custom name (when set) is used in results, descriptive stats, plots, and ratio options; empty falls back to the standard name
- [x] Study type restricts calculation to matching groups; `All` keeps current behaviour
- [x] Both fields are saved to and restored from the settings YAML
- [x] Both fields appear in and drive the exported R script
- [x] Backward compatible with existing settings files that lack the new columns
- [x] New logic covered by unit tests; NEWS + version bump

## How to test

1. Load data with more than one study type, run NCA setup, open **Partial Interval Calculations**.
2. Add a row; set a **Custom Name** (e.g. `Early exposure`) and confirm the results column / plot label / ratio option use it instead of `AUCINT_0-24`.
3. Set **Study Type** to a specific type; confirm the interval is only calculated for groups of that type, and `All` calculates it for every type.
4. Save settings, reload them, and confirm both fields are restored. Export the R script and confirm it reproduces the same intervals.
5. Load an older settings file lacking the new columns and confirm it restores without error (defaults: no custom name, `All`).

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

- **R is not available in this environment**, so lintr and tests were **not executed here** — they were verified statically (added `.R` lines are under the 100-char limit). Please run `devtools::document()`, `lintr::lint_package()`, and `devtools::test()` locally.
- Because `document()` could not be run, I hand-edited `man/update_main_intervals.Rd` to match the updated roxygen for the new `int_parameters` columns. Please re-run `devtools::document()` to confirm it matches.
- Version is bumped to `0.1.0.9195` to avoid colliding with the ROUTE PR (#1462, which uses `.9194`) if both merge; adjust if merge order differs.
- The descriptive-statistics table already keys manual intervals off the same renamed `PPTESTCD`, so custom names flow there too.
- Study-type detection reuses the exact logic from the parameter-selection matrix (`detect_study_types()` with the same group/route/volume args).


Gero1999:
The feature wroks perfectly: 

image

PR #1468, opened 2026-08-31
1441 bug/axis break label width

### LINKED ISSUE #1441: Bug: axis break filtering ignores label width, so long tick labels overlap
### Description
`filter_breaks()` decides which axis breaks to keep by comparing the **distance between break positions** against a minimum (`min_cm_distance`, default `0.5` cm). It never considers how wide the rendered **labels** are, so any label longer than about three characters can overlap its neighbour while still satisfying the rule.

This is most visible in the concentration plots (`pkcg01`/`pkcg02`/`pkcg03`), whose x axis is actual relative time and often carries values like `-0.083`, `23.917`, `119.917`.

The rule is the loop at `R/filter_breaks.R:68-77` — `bdist >= min_cm_distance` is the only test applied.

### To reproduce
1. Run NCA on a dataset with a **dense PK profile** like many timepoints, especially non-integer ones like `-0.083`, `23.917`
2. Go to the **TLG** tab and submit an order including the concentration plots (`pkcg01`, `pkcg02` or `pkcg03`)
3. Look at the x axis and see if the tick labels run into each other

but on a dense PK profile with ~26 irregular timepoints (`adnca_example.csv`):

- `filter_breaks()` keeps **23** labels
- rendered at the image-export size (10 in wide), each label gets **0.43 in**
- the widest label (`119.917`, 7 characters at 11pt) needs roughly **0.63 in**

The labels therefore collide like this:

Image

### Where it shows

- **Exported images** (PNG/PDF, added in #1344) — most obvious, because 11pt text in a 10-inch panel occupies proportionally more width than the browser's ~12px text in a ~1000px panel
- **On-screen plots at narrow viewport widths** — same breaks, so the same collision once the panel is small enough
- Anywhere else `filter_breaks()` is used with long numeric labels

The breaks are identical on the interactive and export paths — this is not an export-specific calculation, only an export-specific *manifestation*.

### Expected behavior

Breaks should be thinned until the rendered labels actually fit. `filter_breaks()` should measure the width of each candidate label (e.g. via `graphics::strwidth()` or the gtable text grob) and drop those that would collide, instead of assuming a fixed 0.5 cm gap is sufficient regardless of label length. On the profile above that would mean keeping roughly 15 labels rather than 23.

note: changing the rule will shift the `vdiffr` snapshots for `g_pkcg.R`, so those need regenerating as part of the fix

h5hoang:
## Issue

Closes #1441

## Description

`filter_breaks()` only compared tick positions against `min_cm_distance` (0.5 cm), never how wide the label rendered, so anything longer than about three characters overlapped its neighbour. On the first subject of `adnca_example` it kept 24 breaks where about 16 fit.

Worst on the side-by-side views, which is the case in the issue screenshot. They filtered before `facet_wrap()` split the panel, so each half-width panel was handed the same 24 breaks as the full-width plot.

Changes:

- A gap must now clear the larger of `min_cm_distance` and the room the two labels either side of it need, measured from the plot's own `axis.text` styling plus a space either side. Sized per pair, so a wide label next to a narrow one costs less than two wide ones.
- The width being filled came from the panel border grob, whose width is `unit(1, "npc")`, the whole device. It now comes from the gtable layout: on a 10 inch device that is 24.20 cm for a single panel instead of 25.40, 11.61 cm for side-by-side, and the legend is subtracted.
- The x scale is applied after the side-by-side branch, so the breaks are filtered against the panel that actually gets drawn.

Dense profile at the export size goes from 24 breaks to 16 on LIN and 24 to 8 on SBS. A sparse profile stays at 8 on LIN, so short-label plots are not over-thinned.

Housekeeping: the label formatter was copy-pasted at all three call sites and is now `format_axis_labels()`, shared with `filter_breaks()`'s default so the filter measures the labels that get drawn; the three duplicated `scale_x_continuous()` blocks collapse into `add_filtered_x_scale()`; `grid` added to DESCRIPTION Imports; `filter_breaks()` errors on a bad `labels` argument instead of falling through to `base::labels()`.

## Definition of Done

- [x] Breaks are thinned until the rendered labels actually fit
- [x] `filter_breaks()` measures each candidate label instead of assuming a fixed 0.5 cm gap
- [x] The dense profile keeps roughly 15 labels rather than 23 (16 rather than 24 on the subject tested)
- [x] `vdiffr` snapshots for `g_pkcg.R` regenerated

## How to test

1. `run_app(datapath = "data-raw/adnca_example.csv")`, map, run NCA, go to TLG
2. Submit `pkcg01 - Individual Side-by-Side` and `pkcg02 - Combined Side-by-Side`. About 8 ticks per half-panel, clearly fewer than the Linear sibling. On main these carried the same 24 ticks as the full-width plot
3. Submit `pkcg01 - Individual Linear`, `pkcg02 - Combined Linear`, `pkcg03 - Mean Linear by Treatment`. No labels touching, roughly 16 ticks against 24 on main
4. Set `xmax` to `12` on a `pkcg01 - Individual Linear` order. Tick count unchanged from main, short labels are still on the 0.5 cm floor
5. Change **X ticks** from `NFRLT` to `AFRLT` and resubmit. Values change to actual times, still no overlap.
6. Spot-check `pkcg01 - Individual Log` and `pkcg03 - Mean Linear by Dose`

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

There is no TLG export on main yet, so the exported PNGs the issue calls the most obvious symptom are not reachable in the app today. I checked that path by cherry-picking this onto #1428 locally.


Gero1999 (approved):
LGTM

Shaakon35:
Potential edge case with manual x-axis limits: `filter_breaks()` currently thins the full candidate-break vector, including values outside the plotted `coord_cartesian(xlim = ...)` range. A retained off-screen break can therefore suppress the first visible break, even though its label is never drawn.

For example, with candidate breaks `0:100`, `xlim = c(51, 57)`, and a spacing rule that retains every second break, the current greedy pass can retain `0, 2, ..., 50, 52, 54, 56`, leaving `51` absent from the displayed axis. Filtering candidates to the built panel's `x.range` before the greedy loop would make the visible range start at `51` instead.

Could we trim `breaks` to `scale_range` immediately after extracting it, and add a regression test using a non-default `xmin`/`xmax`?

Shaakon35:
To fix later:
sometimes the page x out of y does not display:
image

Shaakon35 (approved):
I can still approve :)

h5hoang:
@Shaakon35 thanks for catching this! you're right, the greedy pass started from the first candidate even when it was off-screen, so the visible ticks inherited their alignment from a break that never gets drawn

`breaks` is now trimmed to the panel's `x.range` right after it's extracted, before the greedy pass so in a scenario with `0:100` and `xlim = c(51, 100)` the axis used to start at 52 with `51` missing, but now it should start at 51

added two regression tests with a non-default `xlim`: every returned break lies inside the panel range, and pre-trimming the candidates gives the same result

LMK if this works!

and on the page selector bug u saw, I somehow can't recreate it. it shows `1 out of 21` for me every time i go through the TLG graphs tab and the page selector numbers render properly. do you remember what you were doing right before it went blank? I tried resubmitting an order, changing entries per page, tabbed back and forth between TLG tabs, and more but wasn't able to recreate the `BLANK out of 21`

PR #1480, opened 2026-09-07
Fix profile-level Missing derivation in NCA results

### LINKED ISSUE #1479: Bug: NCA results Missing column can duplicate profile rows
### Description
The NCA Results `Missing` column is currently derived from the long PKNCA result table and joined back to the already-pivoted results using all overlapping columns. Because the intermediate missing table can retain parameter-level metadata, the join can split what should be one subject/profile row into multiple rows when those retained columns differ across parameters.

This makes `Missing` behave as a parameter-row-derived value instead of a subject/profile-level summary.

### To reproduce
1. Run NCA with flag rules enabled.
2. Use data where active flag parameters differ in missingness and/or retained metadata within the same subject/profile.
3. Inspect the NCA Results table.

The same subject/profile can appear more than once because one active flag parameter is missing while another is not.

### Expected behaviour
`Missing` should represent the entire NCA profile row. For each pivoted subject/profile row, it should summarize which requested/active flag parameters are missing, without introducing duplicate rows.

A robust implementation should explicitly aggregate missing active flag parameters at the intended profile level, then join back by those profile keys only, or compute missingness directly from the pivoted profile-level result.

Gero1999:
## Issue

Closes #1479

## Description

Derive the NCA Results `Missing` value once per pivoted subject/profile row instead of allowing parameter-level metadata from the long PKNCA result to participate in the join back to the pivoted table.

This keeps the existing source of truth for missing requested flag parameters (`pknca_res$result`) but explicitly collapses those rows by the profile keys shared with the pivoted output before joining. The change prevents duplicate pivoted NCA Results rows when different active flag parameters carry different parameter-level metadata.

Also removes the now-unused `is_missing` global variable and updates the development version and NEWS entry.

## Definition of Done

- `Missing` is summarised at subject/profile level.
- Active flag parameters with missing `PPSTRES` still produce `Missing` messages.
- Parameter-level metadata can no longer duplicate pivoted result rows.
- Existing accepted/flagged/missing behavior is preserved for active flag rules.

## How to test

Run:

```r
testthat::test_file("tests/testthat/test-pivot_wider_pknca_results.R")
```

Manual check:
1. Run NCA with active flag rules.
2. Confirm the NCA Results table has one row per subject/profile.
3. Confirm the `Missing` column lists missing active flag parameters for that profile.

## Contributor checklist
- [x] New logic is documented

## Notes to reviewer

`git diff --check` passes locally. R-based tests could not be run in this environment because `Rscript` is not installed.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: aNCA
 Title: (Pre-)Clinical NCA in a Dynamic Shiny App
-Version: 0.1.0.9193
+Version: 0.1.0.9194
 Authors@R: c(
     person("Ercan", "Suekuer", email = "ercan.suekuer@roche.com", role = "aut",
            comment = c(ORCID = "0009-0001-1626-1526")),
--- a/NEWS.md
+++ b/NEWS.md
@@ -6,6 +6,7 @@
 
 ## Bug Fixes
 
+* NCA Results now derive the `Missing` flag at the subject/profile level, so parameter-level metadata from active flag parameters can no longer duplicate rows in the pivoted results table (#1479)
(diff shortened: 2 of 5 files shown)


Gero1999:
Was tested successfully with the internal data reported with the problem

